PR #36, opened 2024-04-24
Integrate de-async branch of aptos-core

mzabaluev:
Removing async from `aptos-consensus-types` and `aptos-vm` as per https://github.com/movementlabsxyz/aptos-core/pull/7.


mzabaluev:
This is not being worked on for the time being.

PR #38, opened 2024-04-25
Aptos ZKFP Example

l-monninger:
# Summary
Proposing merge for collaboration viz #36 .


l-monninger:
We've discussed offline, but just commenting for completeness. I agree with the above. This is not intended to be merged in. It's just open for collaboration.

mzabaluev:
To take away from the same discussion, we should run this as a feature branch and sync with the mainline from time to time.

PR #39, opened 2024-04-26
Broken Monza Executor

l-monninger:
# Summary
While investigating the transaction confirmation bug, I eventually discovered that our executor is actually not committing state successfully. Sorry, I should have reviewed that more closely in #30 .

It turns out the issue I was experiencing is not due to our current `monza-opt-executor` failing to move transactions from pending. But, instead, it is a result of state not being reliably committed at all. 

## Steps to Reproduce
Check out the tests in the [`monza-opt-executor` ](https://github.com/movementlabsxyz/movement/tree/partial-monza/broken-executor/protocol-units/execution/monza/opt-executor) crate. 

In particular, I've expanded **`test_execute_block_state_db`** to actually run checks of state values with the state view, and it has not passed. 

Strangely, you can successfully get transactions from the `reader.get_transaction_by_hash`. But, anything involving the `state_view` API does not work. This breaks upstream compatibility with the rest API implementation obtained from Aptos Core. 

## Recommended Approach
- Ostensibly, if you can get  **`test_execute_block_state_db`** to pass, this should fix all downstream issues.
- The `commit_block` method is responsible for actually triggering the ledger info to be committed to state. We can see from just using `db.reader` methods that this is functional. But, something appears to be off about the ledger versioning.
- This should help and is certainly more relevant for the current Aptos API: https://github.com/movementlabsxyz/aptos-core/blob/ea91067b81f9673547417bff9c70d5a2fe1b0e7b/execution/executor-test-helpers/src/integration_test_impl.rs#L53

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
 [[package]]
 name = "abstract-domain-derive"
 version = "0.1.0"
-source = "git+https://github.com/movementlabsxyz/aptos-core?branch=monza#2027a153efa4fda4a2d0abe5e4ded8f95ce49ee7"
+source = "git+https://github.com/movementlabsxyz/aptos-core?branch=monza#06443b81f6b8b8742c4aa47eba9e315b5e6502ff"
 dependencies = [
  "proc-macro2 1.0.81",
  "quote 1.0.36",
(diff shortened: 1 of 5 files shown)


l-monninger:
I believe this was already been merged into Monza. I will confirm and close to clean this up by EOD.

0xmovses (approved):
Let's get these tests merged into main. :ship:

PR #42, opened 2024-05-01
ENG-244 Create tests for `RStarM.sol`

0xmovses:
Adds solidity unit tests to test the gorth16 verification capabilities of `RStarM.sol`

Tests are: 
`testRegisterValidator`
`testVerifyKnownGoodReceipt`
`testVerifyKnownGoodImageIdAndJournal`
`testVerifyMangledReceipts`
`testHonestValidators`
`testDishonestValidators`

### Run the tests 
`cd protocol-units/dispute && forge test`

https://movement.atlassian.net/browse/ENG-189?atlOrigin=eyJpIjoiNWU3ODE4OTQ4NGY2NDZlYWIxZDc2MGU4MTRlNDQzYzgiLCJwIjoiaiJ9


protocol-units/dispute/src/RStarM.sol
```diff
@@ -0,0 +1,259 @@
...
+        console2.log("claim0: ", claim0);
+        console2.log("claim1: ", claim1);
+        Seal memory seal = abi.decode(receipt.seal, (Seal));
+        console2.log("sealed");
+        bool is_verified = this.verifyProof(seal.a, seal.b, seal.c, [CONTROL_ID_0, CONTROL_ID_1, claim0, claim1, BN254_CONTROL_ID]);
+        console2.log("verified", is_verified);
+        return is_verified;
+    }
+
+    function submitOptimisticCommitment(bytes32 blockHash, bytes calldata stateCommitment) external {
```
l-monninger, line 240: This is good. We may change this to the staked rollup concept sooner than later.

protocol-units/dispute/src/RStarM.sol
```diff
@@ -0,0 +1,259 @@
...
+    function resolveDispute(bytes32 disputeHash, DisputeState state) external {
+        require(state > DisputeState.SUBMITTED && state <= DisputeState.UNVERIFIABLE, "Invalid state transition");
+        Dispute storage dispute = disputes[disputeHash];
+        require(dispute.timestamp != 0, "Dispute not found");
+        require(dispute.state == DisputeState.SUBMITTED, "Dispute already resolved");
+        dispute.state = state;
+        emit DisputeResolved(disputeHash, state);
+    }
+
+    function submitProof(bytes32 blockHash, Receipt calldata receipt) external {
```
l-monninger, line 196: So, still for RStarM we would want those optimistic commitment rounds. But, hold off on patching that in for a bit.

protocol-units/dispute/src/RStarM.sol
```diff
@@ -0,0 +1,259 @@
...
+    /// @notice splits a digest into two 128-bit words to use as public signal inputs.
+    /// @dev RISC Zero's Circom verifier circuit takes each of two hash digests in two 128-bit
+    /// chunks. These values can be derived from the digest by splitting the digest in half and
+    /// then reversing the bytes of each.
+    function splitDigest(bytes32 digest) internal pure returns (uint256, uint256) {
+        uint256 reversed = reverseByteOrderUint256(uint256(digest));
+        return (uint256(uint128(uint256(reversed))), uint256(reversed >> 128));
+    }
+
+    function registerValidator() external payable {
```
l-monninger, line 156: It makes even more sense if we going with the staked rollup approach to call these `stake` and `unstake`.

protocol-units/dispute/test/RStarM.t.sol
```diff
@@ -0,0 +1,94 @@
...
+
+        console2.log("verified!");
+        mangled.claim.postStateDigest ^= bytes32(uint256(1));
+        require(!rStarM.verify_integrity(mangled), "verification passed on mangled postStateDigest value");
+        mangled = TEST_RECEIPT;
+
+        mangled.claim.output ^= bytes32(uint256(1));
+        require(!rStarM.verify_integrity(mangled), "verification passed on mangled input value");
+        mangled = TEST_RECEIPT;
+    }
```
l-monninger, line 93: I'd like to see some tests on the optimistic commitment. Include cases for fraudulent actors, honest actors, a mix, etc.

protocol-units/dispute/src/RStarM.sol
```diff
@@ -0,0 +1,259 @@
...
+
+    function submitOptimisticCommitment(bytes32 blockHash, bytes calldata stateCommitment) external {
+        require(validators[msg.sender].isRegistered, "Validator not registered");
+        OptimisticCommitment storage commitment = optimisticCommitments[blockHash];
+        if (commitment.validatorCount == 0) {
+            commitment.blockHash = blockHash;
+            commitment.stateCommitment = stateCommitment;
+            commitment.validatorCount = 1;
+        } else {
+            require(keccak256(commitment.stateCommitment) == keccak256(stateCommitment), "State commitment mismatch");
```
l-monninger, line 248: This is actually a race right now instead of m of n. You would have that the subsequent committers have to  match the commitment of the first committer.

Instead, you'll want to track all commitments and check each time whether you have m of n agreeing on a state.

Even after the state is agreed upon, you'll want to commit to track other commitments for slashing.

l-monninger:
1. Tests are passing.
2. Rewards are omitted, but that's okay for now.
3. The last logical issue above needs to be resolved and tested. 
4. I'm going to push an update to the nix flake to include the foundry packages. 

Once 3 and 4 are done, we can merge this. We can address the rest in subsequent PRs.

l-monninger (changes requested):

l-monninger:
In the future, it would also be good to roll these up into some e2e tests. For now, it could just be something similar to the e2e tests with Sovereign Labs SDK from before.

0xmovses:
Two new tests have been added `honestValidators`, `dishonestValidators`. The race condition has been fixed also.

l-monninger:
@0xmovses 

There's still confusion here: https://github.com/movementlabsxyz/movement/blob/6c3f0095febb72e382aea6615ad5dd421add6162/protocol-units/dispute/src/RStarM.sol#L250

1. Rounds are not global. There are rounds for a given block height. 
2. The race condition still exists. This is roughly what this should look like: https://github.com/movementlabsxyz/rfcs/blob/rfc/mcr/0026-r*m/rfc-r*m.md

You want to keep track of all of the different block commitments submitted. I'll update alongside MCR to show you what I'm talking about.

protocol-units/dispute/src/RStarM.sol
```diff
@@ -0,0 +1,277 @@
...
+        return MIN_STAKE * (100 + incrementFactor) / 100;
+    }
+
+
+    function isCommitmentAccepted(uint256 round) public view returns (bool) {
+        return rounds[round].isAccepted;
+    }
+
+    // Submit an optimistic commitment
+    function submitOptimisticCommitment(bytes32 blockHash, bytes calldata stateCommitment) external payable {
```
l-monninger, line 250: Still not quite right. See comment with permalink.

l-monninger (changes requested):

l-monninger:
@0xmovses these changes are mine to make.

0xmovses:
Do we still want RStarM on main?

0xmovses:
RStarM is not being shipped for now

PR #48, opened 2024-05-10
ENG-317 implement mcr contracts

0xmovses:
### Changelog
- Creates MCR contract
- Adds basic essential test cases for the MCR contract

### Testing 
- Adds the tests: `testUpdateEpoch`,`testHonestValidatorsCommit`, `testDishonestValidatorsMinorityCommit`

To run the tests: `cd protocol-units/dispute && foundry test`

### Notes
- This branch was cut from https://github.com/movementlabsxyz/movement/pull/42 Once this is merged, the diff will make more sense. 
- Does not implement over our staking contracts as these have not been created yet. 
- I've assumed that we are not supporting chain state reversion and Groth16 verification for this iteration of MCR, so I've omitted that logic. We did discuss using this system as a fail-safe strategy in case network participants decide it is worthy to revert chain state for whatever reason, but as mentioned, focusing on core MCR logic for now.


protocol-units/dispute/src/MCR.sol
```diff
@@ -0,0 +1,165 @@
...
+    uint256 public constant SECONDS_IN_MINUTE = 60;
+
+    uint256 public currentEpoch;
+    uint256 public epochDuration;
+    uint256 public constant MIN_STAKE = 1 ether;
+    uint256 public delta = 1 * SECONDS_IN_DAY;
+    uint256 public p = 1 * SECONDS_IN_MINUTE;
+    uint256 public supermajorityStake;
+    uint256 public epochStartTimestamp;
+    uint256 public totalStake;
```
mzabaluev, line 58: These look like internal members of the contract state, why do they need to be public?

protocol-units/dispute/src/MCR.sol
```diff
@@ -0,0 +1,165 @@
...
+        INCORRECT,
+        UNVERIFIABLE
+    }
+
+    uint256 public constant SECONDS_IN_DAY = 86400;
+    uint256 public constant SECONDS_IN_MINUTE = 60;
+
+    uint256 public currentEpoch;
+    uint256 public epochDuration;
+    uint256 public constant MIN_STAKE = 1 ether;
```
mzabaluev, line 53: This should better be grouped with the other constants above.

protocol-units/dispute/src/MCR.sol
```diff
@@ -0,0 +1,165 @@
...
+        Validator storage validator = validators[msg.sender];
+        require(validator.isRegistered, "Validator not registered");
+        require(validator.stake >= amount, "Insufficient stake");
+        validator.stake -= amount;
+        totalStake -= amount;
+        if (validator.stake == 0) {
+            validator.isRegistered = false;
+        }
+        payable(msg.sender).transfer(amount);
+        emit ValidatorDeregistered(msg.sender);
```
mzabaluev, line 111: This event should emitted when any amount was unstaked and regardless of whether the validator has any remaining stake, correct?
So in effect, it's a confirmation that `unstake` was called by the sender with a valid argument, and it carries no other information.

protocol-units/dispute/src/MCR.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    uint256 public constant SECONDS_IN_DAY = 86400;
+    uint256 public constant SECONDS_IN_MINUTE = 60;
+
+    uint256 public currentEpoch;
+    uint256 public epochDuration;
+    uint256 public constant MIN_STAKE = 1 ether;
+    uint256 public delta = 1 * SECONDS_IN_DAY;
+    uint256 public p = 1 * SECONDS_IN_MINUTE;
+    uint256 public supermajorityStake;
```
mzabaluev, line 56: This is not used by any methods, is it a stub for future revisions?

protocol-units/dispute/src/MCR.sol
```diff
@@ -0,0 +1,165 @@
...
+        return epochs[epoch].isAccepted;
+    }
+
+    function submitOptimisticCommitment(bytes32 blockHash, bytes calldata stateCommitment) external {
+        require(validators[msg.sender].isRegistered, "Validator not registered");
+
+        updateEpoch();
+
+        OptimisticCommitment storage commitment = epochs[currentEpoch];
+        commitment.blockHash = blockHash;
```
mzabaluev, line 144: This field can be overwritten by any registered validator and once accepted, the last value is what will be emitted in `BlockAccepted`. Is this by design?

protocol-units/dispute/src/MCR.sol
l-monninger: You can request public members via `methods.data()` (in most ETH client implementations). These are parameters which other users of the protocol might benefit from accessing as well.

protocol-units/dispute/src/MCR.sol
l-monninger: Agreed.

protocol-units/dispute/src/MCR.sol
l-monninger: This is going to be modded progressively. For now, the above is fine, but I will be changing it fairly shortly in a manner that makes emitting an `unstake` amount more reasonable.

protocol-units/dispute/src/MCR.sol
0xmovses: Yes they were left over constants from previous thoughts. I thought maybe its useful to leave them in for future iterations. I'll leave them here for now as I know Liam may add changes.

protocol-units/dispute/src/MCR.sol
l-monninger: I believe this renders the same race condition as in the original RStarM implementation. 

It should look more like this (pseudo-code from RFC)...

```
 function submitOptimisticCommitment(bytes32 blockHash, bytes calldata stateCommitment) external {

        require(validators[msg.sender].isRegistered, "Validator not registered");

        OptimisticCommitment storage commitment = optimisticCommitments[blockHash];
        
        // Increment the count for the submitted stateCommitment
        uint256 currentCount = commitment.stateCommitments[stateCommitment]++;
        // Update the state with the highest commit count if this one surpasses it
        if (currentCount + 1 > commitment.highestCommitCount) {
            commitment.highestCommitCount = currentCount + 1;
            commitment.highestCommitState = stateCommitment;
        }

        emit OptimisticCommitmentSubmitted(blockHash, stateCommitment, currentCount + 1);

        // Check if the count of the highest commitment has reached the threshold
        if (commitment.highestCommitCount >= m) {
            // Block is accepted optimistically after receiving minimum number of commitments
            emit BlockAccepted(blockHash, commitment.highestCommitState);
        }

    }
```

I will be deving on this shortly. Likely will be something I address.

l-monninger (changes requested):

l-monninger:
@0xmovses These changes are mine to make.

l-monninger:
@0xmovses can you review when you get a chance?

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -205,7 +316,7 @@ contract MCR {
             BlockCommitment memory blockCommitment = commitments[blockHeight][validatorAddress];
```
0xmovses, line 316: This might be an unnecessary optimisation at this point, but consider instead of looping. Another constant-time op, as a nested mapping. Again for gas. Where the `uint256` is the blockheight
`mapping(uint256 => mapping(address => BlockCommitment)) public blockCommitments;`

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -165,16 +261,19 @@ contract MCR {
...
 
-        require(commitments[blockCommitment.height][validatorAddress].height < lastAcceptedBlockHeight + leadingBlockTolerance, "Validator has committed to a block too far ahead of the last accepted block");
+        require(blockCommitment.height < lastAcceptedBlockHeight + leadingBlockTolerance, "Validator has committed to a block too far ahead of the last accepted block");
 
         // assign the block height to the current epoch if it hasn't been assigned yet
-        if (!blockHeightEpochAssignments[blockCommitment.height] == 0) {
+        if (blockHeightEpochAssignments[blockCommitment.height] == 0) {
             // note: this is an intended race condition, but it is benign because of the tolerance
-            blockHeightEpochAssignments[blockCommitment.height] = getCurrentEpoch();
+            blockHeightEpochAssignments[blockCommitment.height] = getEpochByBlockTime();
```
0xmovses, line 276: I did read the comment, but I just wonder if there are any unintended side effects from the race condition :thinking:

0xmovses (commented):
Good to see this fleshing out! Mainly comments around gas concerns.

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -57,9 +60,50 @@ contract MCR {
...
+        bytes32 commitment,
+        bytes32 blockId
+    ) public pure returns (BlockCommitment memory) {
+        return BlockCommitment(height, commitment, blockId);
+    }
+
+    // gets the the genesis stake required
+    function getGenesisStakeRequired() public view returns (uint256) {
+        return genesisStakeRequired;
+    }
```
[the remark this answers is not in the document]
l-monninger: Yeah, it's vestigial from a point at which this looked slightly different.

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -57,9 +60,50 @@ contract MCR {
...
+
+    // gets the epoch duration
+    function getEpochDuration() public view returns (uint256) {
+        return epochDuration;
+    }
+
+    // gets the leading block tolerance
+    function getLeadingBlockTolerance() public view returns (uint256) {
+        return leadingBlockTolerance;
+    }
```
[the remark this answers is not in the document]
l-monninger: Vestigial

protocol-units/settlement/mcr/contracts/test/MCR.t.sol
```diff
@@ -0,0 +1,340 @@
...
+        mcr.stakeGenesis{value : 33 ether}();
+
+        // honest signers
+        honestSigners.push(signer1);
+        honestSigners.push(signer2);
+
+        // dishonest signers
+        dishonestSigners.push(signer3);
+
+        for (uint i = 0; i < 50; i++) {
```
0xmovses, line 260: Can we `uint256 epochs = 50`, then
`for (uint i = 0; i < epochs; i++) {`

Without reading the readme, its not obvious to the reader  what the 50 represents.

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -107,7 +151,7 @@ contract MCR {
         
         uint256 totalStake = 0;
         for (uint256 i = 0; i < validators.length(); i++){
-            totalStake += getCurrentEpochStake(validators[i]);
+            totalStake += getStakeAtEpoch(validators.at(i), epoch);
```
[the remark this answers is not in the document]
l-monninger: Yeah, I can add a step to the rollover to precompute this.

protocol-units/settlement/mcr/contracts/src/MCR.sol
```diff
@@ -205,7 +316,7 @@ contract MCR {
             BlockCommitment memory blockCommitment = commitments[blockHeight][validatorAddress];
 
             // check the total stake on the commitment
-            uint256 totalStakeOnCommitment = commitmentStake[blockCommitment.height][blockCommitment.commitment];
+            uint256 totalStakeOnCommitment = commitmentStakes[blockCommitment.height][blockCommitment.commitment];
 
             if (totalStakeOnCommitment > (2 * getTotalStakeForEpoch(blockEpoch))/3 ) {
```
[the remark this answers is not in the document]
l-monninger: Which part? With the update to getTotalStakeForEpoch this would be $O(1)$.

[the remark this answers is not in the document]
l-monninger: We otherwise do need to loop over and greedily check what each validator committed to.

protocol-units/settlement/mcr/contracts/src/MCR.sol
l-monninger: I'm not sure about this one; we'll want to benchmark. 

The reason for the loop here is that we need to find each commitment at each block height. We could store a separate set of `mapping(uint256 => EnumerableBytes32)` and `mapping(bytes32 => BlockCommitment` then iterate over just the unique commitments. But, this ends up bloating storage which is also costly. 

The current operations within the loop become constant time following the change to `getEpochStakeAt`. With the above, we end up moving these constant time operations up into the commitment insertion logic. I think we basically end up paying the same either way.

protocol-units/settlement/mcr/contracts/src/MCR.sol
l-monninger: This race lends to a simpler system than trying to have stake vote on this mapping. With the block height tolerance, you can ensure that it does not allow validators to run ahead and lock the validator set on arbitrarily many block heights; you can only be so far ahead of the work of the honest network. Without that tolerance check, that would be the primary attack vector.

protocol-units/settlement/mcr/contracts/test/MCR.t.sol
l-monninger: It's not exactly epochs. We have have ticks of +1 vm time in the inner loop. I can call it reorgs, because there is one validator set reorg requested per iteration.

protocol-units/settlement/mcr/contracts/src/MCR.sol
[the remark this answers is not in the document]
l-monninger: I'm going to implement this in a way that makes it invalid outside of the current epoch. If we ever do want to peek ahead, what we can do is...
1. Add `stake` values to total stake mapping.
2. Zero out before rolling over. 
3. Do the same as I just added.

l-monninger:
@0xmovses available for review again.

l-monninger:
@musitdev this is the current state of the contract against which you'll want to implement the client.

protocol-units/settlement/mcr/contracts/src/MCR.sol
0xmovses: Hmm yea, it will be expensive either way, was more of a curiosity comment.

protocol-units/settlement/mcr/contracts/src/MCR.sol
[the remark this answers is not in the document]
0xmovses: Ah yes, we need to loop either way.

0xmovses:
```
cd ./protocol-units/settlement/mcr/contracts && forge test
sh: line 1: forge: command not found
error: Recipe `mcr-contract-tests` failed on line 7 with exit code 127
```
CI failing, after that's fixed this looks good.

0xmovses:
I'd be inclined to not run the long running test for CI something like `forge test --no-match-test <test-we-don't want>`

l-monninger:
@0xmovses the long-running test is now like 5 seconds haha. 

In general, I like breaking short tests, i.e., "Checks" and longer tests "Tests," into separate workflows. But, I do want think we want this pipeline running e2e tests. There's a lot that can go wrong which we really don't want to miss from testnet on. We can afford to wait 30 minutes to an hour.

l-monninger (approved):

PR #49, opened 2024-05-10
Remove the zkfp crates

mzabaluev:
The ZKFP implementation is not needed in the current roadmap,
and we never got it to build successfully. Cut it out for the main branch.

The revision prior to removing the crates has been preserved in the `zkfp` branch.


l-monninger (approved):
rollup done lgtm

PR #52, opened 2024-05-10
Git Hygiene

l-monninger:
# Summary
- **Status**: `draft`
- **RFCs**: $\emptyset$
- **Categories**: `misc`

Proposes `CODEOWNERS` and `PR.md` for git hygiene. 

# Core Software Components

N/A

# Ancillary Software Components

N/A

# Testing

N/A

# Outstanding Issues
1. Any other templates we want to include? Issue template?


PR.md
```diff
@@ -0,0 +1,50 @@
...
+-->
+
+# Testing
+
+## e2e and Integration Tests
+
+### Test Plan
+<!---
+Describe how your PR is tested. 
+-->
```
0xmovses, line 28: I'd opt for just a ### Test Plan section. 
If any new unit or e2e tests were added, they can be described in the same section. 

Reasoning: 

I'm not a fan of ### Test Plan and ### Test components. In many cases tests will not need to be added but simply passing CI will suffice for a small change. So in those cases, these sections end up being ignored / redundant. 

We want the PR.md to provide meaningful and useful guidelines, without being overly restrictive / brittle. 

We also want the sections to be filled out for every merge, even if with just small amounts of detail, so they should be flexible enough

PR.md
```diff
@@ -0,0 +1,50 @@
...
+# Core Software Components
+
+<!--
+Describe core software components added or modified by the PR. 
+-->
+
+# Ancillary Software Components
+<!--
+Describe components added or modified while working on core software components.
+-->
```
0xmovses, line 19: I've seen many Repos where `PR.md` guidelines get ignored because their not relevant for the change. For small PRs there may be no changes to any Core or Ancillary software components, so these sections would be ignored. 

How about a more general section instead, just a # Changelog section where updates can be listed. 

If the developer wants to expand that and break it into sub-sections they can do so ad-hoc and we encourage their literary inclinations!

0xmovses (commented):
Needed, just two thoughts.

PR.md
0xmovses: I also like # Changelog because it can encapsulate all changes, for eg. updating docs, or a utility script, this wouldn't be a software component change.

PR.md
l-monninger: What about just?:

# Testing
...

0xmovses: 👍

l-monninger:
Updated. See if that's more to your liking.

0xmovses (approved):
LGTM

l-monninger:
I think you mean RD.

l-monninger:
Pretty tough `libssl` linking bug. Starting to keep track of attempted solutions. 

- Generally, nix `openssl.dev` doesn't seem to work for Ubuntu x86_64. This messes up the client tests presumably because of the dynamic linking in Rust `openssl-sys` from the Aptos client.
- Ultimately, I might need to do this to the nix flake: https://nixos.wiki/wiki/Rust
- Current, I'm working off of this thread: https://github.com/sfackler/rust-openssl/issues/1663

l-monninger:
@mzabaluev @musitdev @andyjsbell Going to merge to get a template in since we have a lot of activity at the moment.

PR #58, opened 2024-05-13
ENG-326: Full Node Commitments

l-monninger:
# Summary
- **Status**: `review`
- **RFCs**: [RFC 29: MCR (draft)](https://github.com/movementlabsxyz/rfcs/pull/29).
- **Categories**: `protocol-units`, `networks`, `scripts`.

Add MCR commitment processing to Suzuka full node.

# Changelog

* Suzuka full node settless block commitments with the MCR contract.

# Testing

* Unit tests of mcr-settlement-manager using the mock of mcr-settlement-client.

# Design
full-node-commitments
1. The top portion of the graphic above shows a high-level overview of dataflow in the system. Full-nodes execute blocks, generate commitments for the block, and then wait for events from the settlement contract.
2. The bottom part shows an asynchronous pattern for this scheme wherein a commitment loop is responsible for handling the actual submission via the client to the settlement contract. This may be desirable for a few reasons:
   1. Execution can still optimistically proceed, i.e., does not have to wait on a commitment to be accepted.
   2. Honest nodes can still avoid making faulty commitments. The job of withholding these commitments is just delegated to the commitment loop.
   3. Separation of concern makes it easier to deal with batching and adjustments faulty commitment avoidance scheme (multisig).
   4. This can be generalized to a very generic MCR commitment service if the demand presents itself in the SDK.

# Entrypoints for understanding
**MCR Settlement Client (Stub)**: https://github.com/movementlabsxyz/movement/tree/eng-326/full-node-commitments/protocol-units/settlement/mcr/client


l-monninger:
We'll use the working CI.

protocol-units/settlement/mcr/client/src/stub.rs
```diff
@@ -0,0 +1,100 @@
...
+use std::sync::Arc;
+use tokio::sync::{Mutex, mpsc};
+use crate::CommitmentStream;
+
+#[tonic::async_trait]
+pub trait McrSettlementClientOperations {
+    async fn post_block_commitment(&self, block_commitment: BlockCommitment) -> Result<(), anyhow::Error>;
+    async fn stream_block_commitments(&self) -> Result<CommitmentStream, anyhow::Error>;
+    async fn get_commitment_at_height(&self, height: u64) -> Result<Option<BlockCommitment>, anyhow::Error>;
+}
```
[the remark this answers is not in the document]
l-monninger: My b. Should be deduped.

util/movement-types/src/lib.rs
```diff
@@ -138,4 +138,32 @@ impl Block {
...
+    fn from(commitment : Commitment) -> Vec<u8> {
+        commitment.0
+    }
+}
+
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct BlockCommitment {
+    pub height : u64,
+    pub block_id : Id,
+    pub commitment : Commitment,
```
0xmovses, line 168: I think we need e2e tests here to call the MCR contract https://github.com/movementlabsxyz/movement/pull/48

We need these tests before we can be confident.

0xmovses (commented):
One comment about e2e tests, up to yo whether we want to add them in this PR or a separate one.

l-monninger:
@0xmovses for now, Mikhail will be working against a stub. e2e tests can come elsewhere.

l-monninger:
@mzabaluev Updated the stub wit batching.

l-monninger:
Oh, one more thing you can now expect that will be helpful...

l-monninger:
In the contract and now stub, I added a method `get_max_tolerable_height` which returns the height to which the contract allows any validator to commit. The corresponding restriction is needed to prevent validators from committing to far ahead and effectively staging an attack that locks down the validator set by assigning every block-height to the current epoch. 

Here's the type signature. 

```rust
async fn get_max_tolerable_block_height(&self) -> Result<u64, anyhow::Error>;
```

protocol-units/execution/maptos/opt-executor/src/executor.rs
```diff
@@ -282,14 +286,19 @@ impl Executor {
...
-		}
+			)?
+		};
 
-		Ok(())
+		let commitment = Commitment::digest_state_proof(&proof);
+		Ok(BlockCommitment {
+			block_id: Id(block_id.to_vec()),
+			commitment,
+			height: latest_version,
```
mzabaluev, line 300: This is probably bonkers. Should the current height be available in the executor state? I assume it is stateful and can only execute blocks in a single chain.

l-monninger: Yes, there's only a single chain that should be considered inside `execute_block`, no forks that you have to consider. All of that will be handled outside by reverting blocks and the like.

I think `latest_version` should be valid. One test to write is to check that it is increasing with no gaps, i.e.,

$$
f(k) = k : k \in N
$$

$$
f(n) = f (n - 1) + 1 
$$

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -138,6 +138,14 @@ where
             ).await?
         }.into_inner();
 
+        let settlement_client = self.settlement_client.clone();
+        // TODO: consume the commitment stream to finalize blocks
+        let commitment_stream = settlement_client.stream_block_commitments().await?;
+
+        tokio::spawn(async move {
+            process_commitments(receiver, settlement_client).await;
+        });
```
mzabaluev, line 147: This is less well organized than tasks-as-methods, but it enables parallel execution. `try_join!` only gives you a compound future that's polled by a single thread.

Unfortunately, the top-down design dictated by traits like `SuzukaFullNode` where the "god object" is borrowed immutably as `&self` by async methods does not work well with background tasks that ideally need to work on moved, rather than shared, bits of state. You end up needing `Arc<Mutex<_>>` for anything mutable even if conceptually it's only used in one task. Another smell is needing `async_channel`: there is hardly ever a genuine need to have multiple consumers on the channel, you are only forced to clone it by the constraints on higher-level API, but it's a slower, more synchronization-intensive channel implementation than `tokio::sync::mpsc`.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -225,6 +225,14 @@ where
 {
     while let Some(commitment) = receiver.recv().await {
         println!("Got commitment: {:?}", commitment);
+        let max_height = settlement_client.get_max_tolerable_block_height().await?;
```
mzabaluev, line 228: This is probably naive and wasteful to do once per block.

If the `post_block_commitment` call below can be expected to provide back-pressure, we can implement simple batching by draining the receiver to form the next batch.

protocol-units/settlement/mcr/client/src/stub.rs
```diff
@@ -0,0 +1,149 @@
...
+        for commitment in block_commitment {
+            self.post_block_commitment(commitment).await?;
+        }
+        Ok(())
+    }
+
+	async fn stream_block_commitments(&self) -> Result<CommitmentStream, anyhow::Error> {
+		let receiver = self.stream_receiver.clone();
+		let stream = async_stream::try_stream! {
+			let mut receiver = receiver.write().await;
```
musitdev, line 65: The Mutex is write-locked for the whole duration of the stream. So if you call `stream_block_commitments` twice, the second stream will be locked and send nothing.
I think this pattern to return a stream in a trait that can be called several times is problematic because you'll need to implement sort of broadcast channel under and not all impl can do this easily.
To avoid that, I separate the post from the received part. The post call is implemented in the trait and the stub constructor return the stream so this way we are sure that the stream is created once, and it removes all Mutex need. In the end, we need only one stream.

mzabaluev: @musitdev I think we should have the stream type somehow available from the client trait (as an associated type?) to be able to mock the client in tests and have the same generic code for production.

musitdev: You have it from the constructor. One trait to post and one trait to stream. You get them differently.
For example:
```
impl McrSettlementClient {
	pub fn new() -> Result<(Self, CommitmentStream), anyhow::Error>  {
		...
	}
}
let (client, stream) = McrSettlementClient::new();
```

musitdev: A more idiomatic way can be 
```
impl McrSettlementClient {
	pub fn build() -> Result<(impl McrSettlementClientOperations, CommitmentStream), anyhow::Error>
	{
		todo!()
	}
}
```

mzabaluev: Then it will have to be passed in as [another parameter](https://github.com/movementlabsxyz/movement/blob/843822aa2cd6d9d9c88df9abc6c067bd64d07a6d/networks/suzuka/suzuka-full-node/src/partial.rs#L41) when constructing the full node. Also, what should happen if the stream handle is dropped (because, perhaps the consumer is only interested in submitting)?

mzabaluev: > I think this pattern to return a stream in a trait that can be called several times is problematic because you'll need to implement sort of broadcast channel under and not all impl can do this easily.

This is a problem with the stub (and can be solved by "don't do that, then" for now), but in a protocol like gRPC, it wouldn't be a problem to establish multiple subscriptions except duplicate traffic on the wire. Maybe we should redesign the API to not allow this accidentally.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: Which tasks specifically do you want to have work on moved instead of shared state?

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: 1. The intent I had is to have the client crate simply call and stream events from the contract--keeping it fairly raw.
2. The layer or an intermediate layer to the one you're working on is where I would envision backpressure to be handled. 
3. You would not use multiple `post_block_commitment`calls, you would use `post_batch_block_commitment` (which sends in one transaction). 
4. I would do this when you detect a condition for draining such as all sent block heights have been either accepted or rejected. I would then  call `get_max_tolerable_block_height().await?` out to that height.

l-monninger: Also, when we start addressing reorgs we might want to expose the contract `getCommittedBlockAtHeight` through this client so that we avoid trying to commit over where we had already made mistakes.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: Operations on receivers I take it?

l-monninger: I think the easiest workaround is just to have all methods using the receiver be `self` which is fairly straightforward refactor from here. Maybe toss into a new issue?

l-monninger:
I had a chance to look through. I would like extract this out into a crate with a strict `McrCommitmentManager` or similar. That struct will own an MCR client instance (no other logic needs). Per comment above, the idea is that this will handle back pressure and grow in complexity. We can then work on backpressure implementation therein.

I can sketch this out a bit if you want. I’ll also hit the `mpsc` refactor therein. It’s actually rather concerning to me because logically there is no reason to desire racing to read transactions off the or mpmc channel.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Yes, node object's public API should perhaps have only two stages:

- Synchronous construction, either with a simple associated constructor fn, or, if needed, a builder API;
- Something like `async fn run(self) -> Result<(), NodeError>` to consume and run the instance. Maybe even `Result<!, NodeError>` to signal that the function never returns in the normal case.

protocol-units/settlement/mcr/client/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+pub mod stub;
+
+#[cfg(feature = "stub")]
+pub use stub::*;
+
+type CommitmentStream =
+	std::pin::Pin<Box<dyn Stream<Item = Result<BlockCommitment, anyhow::Error>> + Send>>;
+
+#[async_trait::async_trait]
+pub trait McrSettlementClientOperations {
```
musitdev, line 14: With the stream discussion, I took a closer look at the code and I have this remark.
The stub McrSettlementClient use send+sync structure to be thread safe.
It's not imposed by the trait. So if the client is means to be used between thread, we should add send+sync to the trait or remove the Arc/Mutex part of the implementation. 
ie :
```
pub trait McrSettlementClientOperations: Send + Sync {
...
}
```

protocol-units/settlement/mcr/client/src/stub.rs
musitdev: In fact, I was wrong. I focus on the implementation and not of the interface. What we should do is return the trait from the constructor. The stream creation can be differed. What the implementation should do it to take care if the creation is not possible (for example only one creation like in the stub) is to return an error.
So the constructor should be 
```
impl McrSettlementClient {
	pub fn build() -> impl McrSettlementClientOperation
	{
		//current code
	}
}
```
If McrSettlementClientOperation is send+sync it should be the same.

musitdev: In fact as stub impl is for test purpose you can do has you want. I was more thinking as the way I'm doing.

protocol-units/settlement/mcr/client/src/lib.rs
mzabaluev: I think it's stylistically better to not make auto-traits pre-requisites; you might never need to pass the client itself between threads.

protocol-units/execution/maptos/opt-executor/src/executor.rs
```diff
@@ -609,6 +607,12 @@ mod tests {
 			assert!(queried_account_address.is_some());
 			let account_resource = account_state_view.get_account_resource()?;
 			assert!(account_resource.is_some());
+
+			// Check the commitment against state proof
+			let state_proof = db_reader.get_state_proof(latest_version)?;
+			let expected_commitment = Commitment::digest_state_proof(&state_proof);
+			// assert_eq!(block_commitment.height, i);
```
mzabaluev, line 614: Looks like using latest DB version for block height was not a good idea.

protocol-units/execution/maptos/opt-executor/src/executor.rs
```diff
@@ -529,27 +556,30 @@ mod tests {
...
-				vec![0],
+				vec![],
 				vec![],
 				current_time_micros,
-			));*/
+			));
 
 			// Create a state checkpoint transaction using the block ID.
-			// let state_checkpoint_tx = Transaction::StateCheckpoint(block_id.clone());
+			let state_checkpoint_tx = Transaction::StateCheckpoint(block_id.clone());
```
mzabaluev, line 582: This transaction is not used anywhere, was it meant to be added to any of the blocks?

protocol-units/execution/maptos/opt-executor/src/executor.rs
```diff
@@ -615,7 +638,7 @@ mod tests {
 			// Check the commitment against state proof
 			let state_proof = db_reader.get_state_proof(latest_version)?;
 			let expected_commitment = Commitment::digest_state_proof(&state_proof);
-			// assert_eq!(block_commitment.height, i);
+			assert_eq!(block_commitment.height, i);
```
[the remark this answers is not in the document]
mzabaluev: No, that's passing after https://github.com/movementlabsxyz/movement/pull/58/commits/c8edb4a8b85691475c3d537fe53707e0c5c47215

0xmovses:
It seems to be just a parralism issue with the tests.
With `cargo test --package maptos-opt-executor --lib -- executor::tests --show-output --test-threads=1` all tests pass with the latest commit. Without the `test-threads=1` `test_execute_block_state_get_api` will fail.

I'm fine with disabling parallel testing until we want to stablise it later.

l-monninger:
I figured it out fellas.

If you actually just trace the error, you'll see that it comes from Aptos speculation. A little ways down in the trace you'll see an error in `aptos-move/aptos-vm/src/block_executor/mod.rs:439:21` that is wonderfully descriptive.

```
if state_view.id() != StateViewId::Miscellaneous {
    // Speculation is disabled in Miscellaneous context, which is used by testing and
    // can even lead to concurrent execute_block invocations, leading to errors on flush.
    flush_speculative_logs(pos);
}
``` 

So, this is a concurrency issue that Aptos knows about. My guess is that it shows up when running all test merely because there's more switching between tasks.

Simply changing the a line where the `StateView` is constructed to use `StateViewId::Miscellaneous` works. This is what I've done [here](https://github.com/movementlabsxyz/aptos-core/tree/monza-issues/state-view) and used to solve this issue be set the rev to this fix in `aptos-core`. 

Unfortunately, this doesn't also happen to solve the `BlockMetadata` in the same transaction issue; the workaround is still needed. 

Ultimately, if we're going to keep this `StateViewId::Miscellaneous` bit, we should probably feature flag it in our `aptos-core` fork.

mzabaluev:
CI attempts to run monza-full-node in somewhat arcane script setup. I can try to adapt this for suzuka, but I'm afraid most of it will be guesswork as I'm not familiar with the scripts or their expectations.

mzabaluev:
I have adapted scripts to run the suzuka node and client, they fail. I can make out a "blob not found" error, but need to see if this is the cause of failure. We need to cut down on the amount of output in these scripts, the actions view is nearly unusable.

l-monninger:
I'll take care of it. Let me merge in some logging stuff as well.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -177,10 +177,23 @@ where
...
+			let block_metadata = self.executor.build_block_metadata(
+				HashValue::sha3_256_of(block_id.as_bytes()),
+				block_timestamp
+			).await?;
+			let block_metadata_transaction = SignatureVerifiedTransaction::Valid(
+				Transaction::BlockMetadata(
+					block_metadata
+				)
+			);
+			block_transactions.push(block_metadata_transaction);
```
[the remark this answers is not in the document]
mzabaluev: Maybe even lower it into maptos-opt-executor, then?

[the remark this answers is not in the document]
l-monninger: Yeah, we may have to reevaluate at some point. But, for now, we are going to assume that the executor can correctly decide the block height based on, well, the number of blocks it has seen. 

This is, at least on the surface, a viable assumption because the if it is an honest node, it should fetch all the blocks up from the DA (and when we have state syncing from other nodes that have done so).

However, if we do need have it come up from the DA, the complexity we will have to deal with is that Celestia blob heights are continually pushed regardless of whether additional blobs have been added to a given namespace.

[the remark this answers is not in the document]
l-monninger: In theory, the sequencer could address by keeping its own counter. But, this could be problematic.

protocol-units/execution/maptos/opt-executor/src/executor.rs
```diff
@@ -329,7 +329,7 @@ impl Executor {
 		// Context has a reach-around to the db so the block height should
```
l-monninger, line 329: LGTM hopefully we'll have a productive conversation about where implementations are separate on Friday and Monday.

l-monninger: Scratch that. Build error. Please try to fix in the morn.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: I have reworked this to be more along the lines suggested in https://github.com/movementlabsxyz/movement/issues/63#issuecomment-2120184646

The commitment processing loop has been moved out to the mcr-settlement-manager crate.

util/movement-types/src/lib.rs
mzabaluev: https://github.com/movementlabsxyz/movement/pull/58#issuecomment-2107390306

flake.nix
```diff
@@ -12,12 +13,21 @@
...
       system: let
 
+        # nix does not handle .cargo/config.toml
+        RUSTFLAGS = if pkgs.stdenv.hostPlatform.isLinux then
+          "--cfg tokio_unstable -C force-frame-pointers=yes -C force-unwind-tables=yes -C link-arg=-fuse-ld=lld -C target-feature=+sse4.2"
+        else if pkgs.stdenv.hostPlatform.isWindows then
+          "--cfg tokio_unstable -C force-frame-pointers=yes -C force-unwind-tables=yes -C link-arg=/STACK:8000000"
+        else
+          "--cfg tokio_unstable -C force-frame-pointers=yes -C force-unwind-tables=yes";
+
```
0xmovses, line 30: 👀 Be great if nix did handle config.toml. Seems subpar that it doesn't

networks/monza/monza-full-node/src/partial.rs
```diff
@@ -62,12 +62,16 @@ impl <T : MonzaExecutor + Send + Sync + Clone>MonzaPartialNode<T> {
...
 
             match transaction_result {
                 Ok(transaction) => {
-                    println!("Got transaction: {:?}", transaction);
+
+                    #[cfg(feature = "logging")]
+                    {
+                        tracing::debug!("Got transaction: {:?}", transaction)
+                    }
+
```
0xmovses, line 74: 👌

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -1,57 +1,77 @@
...
+use std::sync::Arc;
+use std::time::Duration;
 
-#[derive(Clone)]
-pub struct SuzukaPartialNode<T : SuzukaExecutor + Send + Sync + Clone> {
-    executor: T,
-    transaction_sender : Sender<SignedTransaction>,
-    pub transaction_receiver : Receiver<SignedTransaction>,
-    light_node_client: Arc<RwLock<LightNodeServiceClient<tonic::transport::Channel>>>,
+pub struct SuzukaPartialNode<T> {
```
0xmovses, line 26: Let's add in a doc comment for the struct and all its fields.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -1,57 +1,77 @@
...
+			read_commitment_events(commitment_events),
+		)
+	}
 
 	fn bind_transaction_channel(&mut self) {
 		self.executor.set_tx_channel(self.transaction_sender.clone());
 	}
 
-	pub fn bound(
+	pub fn bound<C>(
```
0xmovses, line 64: any `pub` methods should have a doc comment.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -110,9 +140,9 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
     // receive transactions from the transaction channel and send them to be executed
     // ! This assumes the m1 da light node is running sequencer mode
     pub async fn read_blocks_from_da(&self) -> Result<(), anyhow::Error> {
-        
+
```
0xmovses, line 143: Let's turn those comments into doc comments

protocol-units/execution/suzuka/executor/src/v1.rs
```diff
@@ -225,7 +248,9 @@ mod opt_tests {
         ));
         let txs = ExecutableTransactions::Unsharded(vec![tx]);
         let block = ExecutableBlock::new(block_id.clone(), txs);
-        executor.execute_block(FinalityMode::Opt, block).await?;
+        let commitment = executor.execute_block(FinalityMode::Opt, block).await?;
+
+        // TODO: test the commitment
```
0xmovses, line 253: Is this TODO outside the scope of this PR? Maybe lets make a tracking ticket . issue.

protocol-units/settlement/mcr/contracts/broadcast/DeployMCR.s.sol/1569/run-1716348078.json
0xmovses, line 1: Let's `.gitignore` the genned file.

protocol-units/settlement/mcr/contracts/broadcast/DeployMCR.s.sol/1569/run-latest.json
0xmovses, line 1: All of these files

protocol-units/settlement/mcr/manager/src/manager.rs
```diff
@@ -0,0 +1,182 @@
...
+	async fn post_block_commitment(
+		&self,
+		block_commitment: BlockCommitment,
+	) -> Result<(), anyhow::Error> {
+		self.sender.send(block_commitment).await?;
+		Ok(())
+	}
+}
+
+fn process_commitments<C: McrSettlementClientOperations + Send + 'static>(
```
0xmovses, line 45: I'd really love to see a unit test or two, just testing `process_comitments`

protocol-units/settlement/mcr/manager/src/manager.rs
```diff
@@ -0,0 +1,182 @@
...
+							max_height = new_max_height;
+							ahead_of_settlement = false;
+						}
+					}
+				}
+				else => break
+			}
+		}
+	})
+}
```
0xmovses, line 119: If this works, for now it can get merged in as I don't want to slow this PR from being merged too much. But I think it could be refactors to not always `match` on results, instead update the Return types of the outer scope so that we can just bubble up the error with ?`. Then also, generaly reduce the nested ifs where reasonable

0xmovses (commented):
Thanks for this! Just a few thoughts

protocol-units/settlement/mcr/manager/src/manager.rs
mzabaluev: I'd like to use `?`, but it was not possible due to https://github.com/tokio-rs/async-stream/issues/63
There is a comment about this on top of the macro invocation.

protocol-units/settlement/mcr/manager/src/manager.rs
mzabaluev: I believe this is exercised by the two unit tests I did add, that exercise the module's API. I'm not a fan of implementation detail tests, but if there's a problem with coverage (and it there is a good reason why the code can't be reached from public API), we can test the implementation fn as well.

protocol-units/settlement/mcr/contracts/broadcast/DeployMCR.s.sol/1569/run-1716348078.json
mzabaluev: Thanks for spotting, cleaned up in https://github.com/movementlabsxyz/movement/pull/58/commits/06ce28dcc9c98c3f89730886eb156b8bef6e03eb

protocol-units/settlement/mcr/contracts/broadcast/DeployMCR.s.sol/1569/run-latest.json
mzabaluev: https://github.com/movementlabsxyz/movement/pull/58/commits/06ce28dcc9c98c3f89730886eb156b8bef6e03eb

protocol-units/settlement/mcr/manager/src/manager.rs
mzabaluev: I can add tests in the same pattern to see if back-pressure works as expected, that will check more dynamic conditions.

0xmovses: Yes more back pressure / load.  The unit tests here are good, but perhaps one more test a little more dynamic would be appreciated.

protocol-units/settlement/mcr/contracts/broadcast/DeployMCR.s.sol/1569/run-1716348078.json
0xmovses: For resolving comments, no worries about replying with commit hash, you can just unresolve the comment yourself after committing the change.

protocol-units/settlement/mcr/manager/src/manager.rs
mzabaluev: Done in https://github.com/movementlabsxyz/movement/pull/58/commits/df9cb3849a06beb674e9d93160089fc84431260c

protocol-units/execution/suzuka/executor/src/v1.rs
mzabaluev: I've added some checks in https://github.com/movementlabsxyz/movement/pull/58/commits/fae30bec917049a87ce0a577ff679f550f005f07
Not testing the commitment hash itself, it would require replicating much of the internal code.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: There should be a wider ticket for documenting our public APIs (when they are stabilized enough to avoid unnecessary churn). I have filed #81 to track this.

andyjsbell (approved):
lgtm

PR #60, opened 2024-05-13
ENG-337 Implement Move-side bridge contract

0xmovses:
### Changelog 
- Adds `METHBridge.move`
- Adds basic tests

For full description of implementation see the RFC https://github.com/movementlabsxyz/rfcs/blob/main/0021-m2-native-bridge-contract/rfc-21-m2-native-bridge-contract.md

### Tests
`cd protocol-unites/bridge && aptos move test`

### Notes
I had to deviate slightly away from the pseudo code for the fields of this struct 
```rust
struct PendingWithdrawalEvent has drop, store {
        nonce: u256,
        owner: address,
        token_id: u128,
        amount: u64,
    }
 ```
The original RFC
```rust
struct PendingWithdrawal {
    request: PendingWithdrawalRequest,
    nonce: u256
}
```
In my implementation `PendingWithdrawalRequest` would have to have the ability `copy` which we wouldn't want, so instead we just use its fields.


protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
+module 0x1::METHBridge {
```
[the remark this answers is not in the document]
andyjsbell: METH or mETH would be my preference

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+    use std::signer;
+    use std::vector;
+    use aptos_framework::coin;
+    use aptos_framework::event;
+    use aptos_framework::aptos_coin::AptosCoin;
+    use aptos_framework::event::{EventHandle};
+    use aptos_framework::account;
+
+    const BRIDGE_ACCOUNT: address = @0x1; //Should change this later to actual bridge address
```
andyjsbell, line 10: Changing this to anything but `0x1` breaks the tests.

We can also think about having this pulled out into `Move.toml` as an address as `bridge_account` and remove this constant and use `@bridge_account` potentially

protocol-units/bridge/sources/METHBridge.move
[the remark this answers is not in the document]
l-monninger: MOVETH

andyjsbell, line 1: Could be great to get some documentation here for the module and maybe name it, something like `native_bridge::moveth`

protocol-units/bridge/Move.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "METHBridge"
+version = "1.0.0"
+authors = []
+
+[addresses]
+
```
andyjsbell, line 7: As mentioned below but we can use this to define `token_address`

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+    const EPENDING_WITHDRAWAL_NOT_FOUND: u64 = 3;
```
andyjsbell, line 11: Noticed we asserts for `1` and then we have this `const` pointing to `3` .. maybe it would be good to get the two defined here

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    struct PendingWithdrawalRequest has drop, store {
+        owner: address,
+        token_id: u128,
+        amount: u64,
+    }
+
+    struct PendingWithdrawal has drop, store {
+        request: PendingWithdrawalRequest,
```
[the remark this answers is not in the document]
0xmovses: I tried to stay closely alligned with the design in the RFC https://github.com/movementlabsxyz/rfcs/blob/main/0021-m2-native-bridge-contract/rfc-21-m2-native-bridge-contract.md#design 

Each `PendingWithdrawal` needs a request and its inner fields associated with it, the inner fields could just be cloned into fields of `PendingWithdrawal` but this seemed fine and cleaner.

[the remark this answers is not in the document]
andyjsbell: Yeah I saw that but I would say it is open for interpretation and is pseudocode.  Probably a rethink in how we are storing these requests and structure could be worthwhile.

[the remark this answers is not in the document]
0xmovses: Hashmap might be nicer. I'll refactor this.

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+        deposit_owner: address,
+        token_id: u128,
+        nonce: u256,
+        amount: u64
+    ) acquires BridgeAccount {
+        let trusted_address = signer::address_of(trusted);
+        assert!(trusted_address == BRIDGE_ACCOUNT, 1); // Verify trusted signer
+
+        let bridge_account = borrow_global_mut<BridgeAccount>(trusted_address);
+        let coin = coin::withdraw<AptosCoin>(trusted, amount);
```
andyjsbell, line 60: In the RFC its mentioned minting the asset and not depositing, unless I have got the wrong end of the stick.  What are the advantages over depositing?  One thing I am also curious about is the security of the bridge account, we are using this to secure the bridge as `trusted_address` and the funds that are bridged.  Could this maybe use another pattern?

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+
+        let pending_withdrawal_event = PendingWithdrawalEvent {
+            nonce,
+            owner: owner_address,
+            token_id,
+            amount,
+        };
+
+        // Add the pending withdrawal to the pending_withdrawals vector
+        vector::push_back(&mut bridge_account.pending_withdrawals, pending_withdrawal);
```
andyjsbell, line 105: I think this might have been mentioned but couldn't we use a different type of container for this for the lookups?

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+        assert!(trusted_address == BRIDGE_ACCOUNT, 1); // Verify trusted signer
+
+        let bridge_account = borrow_global_mut<BridgeAccount>(trusted_address);
+
+        // Find the pending withdrawal with the matching owner, token_id, and nonce
+        let pending_withdrawal_index = find_pending_withdrawal_index(bridge_account, owner, token_id, nonce);
+
+        // ensure the pending withdrawal was found 
+        assert!(pending_withdrawal_index != vector::length(&bridge_account.pending_withdrawals), EPENDING_WITHDRAWAL_NOT_FOUND); 
+
```
andyjsbell, line 129: I guess `simple_map` or the like could be better for this

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+            if (pending_withdrawal.request.owner == owner &&
+                pending_withdrawal.request.token_id == token_id &&
+                pending_withdrawal.nonce == nonce) {
+                return i
+            };
+            i = i + 1;
+        };
+        len
+    }
+
```
andyjsbell, line 209: Looks expensive to me if I'm at the bottom

protocol-units/bridge/sources/METHBridge.move
0xmovses: Yes this seemed a not-ideal to me also when implementing .. I'll have to research other bridging patterns, look at hyperlane implementations and come back with some ideas. @l-monninger let me know if you already have some other ideas.

protocol-units/bridge/sources/METHBridge.move
```diff
@@ -0,0 +1,455 @@
...
+    fun init_module(bridge: &signer) {
+        let bridge_address = signer::address_of(bridge);
```
andyjsbell, line 211: We might need to think how we are managing this bridge address.  It seems here we could just store the bridge account and then potentially have some god function to rotate it if needed/compromised

protocol-units/bridge/sources/METHBridge.move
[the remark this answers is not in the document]
andyjsbell: let it moveth 👍

PR #61, opened 2024-05-14
Executor API cleanup and simplification

mzabaluev:
Make `FinalityMode` passed by value like a simple mode selector parameter should be.
Change signatures of methods that don't need to be async
or fallible.

### DIFF
--- a/networks/monza/monza-full-node/src/partial.rs
+++ b/networks/monza/monza-full-node/src/partial.rs
@@ -41,16 +41,18 @@ impl <T : MonzaExecutor + Send + Sync + Clone>MonzaPartialNode<T> {
         }
     }
 
-    pub async fn bind_transaction_channel(&mut self) -> Result<(), anyhow::Error> {
-        self.executor.set_tx_channel(self.transaction_sender.clone()).await?;
-        Ok(())
-    }
+	fn bind_transaction_channel(&mut self) {
+		self.executor.set_tx_channel(self.transaction_sender.clone());
+	}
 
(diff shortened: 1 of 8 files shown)


mzabaluev:
I have further YAGNI feeling about parts of the abstract API:

- Do we really need the dynamic selection with `FinalityMode`? It's currently unimplemented except for the optimistic mode.
- What do we need the `*Executor` traits for? Future mocking?

l-monninger:
@mzabaluev `FinalityMode` is entirely unnecessary (it doesn't actually make sense w.r.t. to the intended design). I thought I had removed it already. 

Executor traits are for organization and reuse in a would be SDK.

mzabaluev:
> Executor traits are for organization and reuse in a would be SDK.

In general, I would prefer not introducing traits until there are at least two implementations genuinely needing the abstraction. Especially those transformed by `async_trait` as they have a runtime cost.

mzabaluev:
> @mzabaluev `FinalityMode` is entirely unnecessary (it doesn't actually make sense w.r.t. to the intended design). I thought I had removed it already.

I can cut it out in this PR while we are at it, the only problem would be a merge conflict with #58.

l-monninger:
> In general, I would prefer not introducing traits until there are at least two implementations genuinely needing the abstraction. Especially those transformed by async_trait as they have a runtime cost.

I tend to do a fair bit of type-level programming first and like be able to refer to traits for quick understanding on the first pass. So, I'll need to work on that. 

I'm trying to use `async_trait` only where there would be or I anticipate `Box<dyn ...>` usage.  The `async_trait` vtables also increase binary size, no?

l-monninger (approved):
RD

PR #66, opened 2024-05-17
ENG 394 additional unit test for memseq

cde8:
- **rustfmt**
- **test(memseq): add initialization unit test**
- **test(memseq): Add tests for with_block_size and with_building_time_ms methods**
- **test(memseq): Add new unit tests for try_move_rocks function**
- **test(memseq): Add test for waiting on next block with no transactions**
- **test(memseq): add support for concurrent access testing**
- **test(memseq): add tests for error propagation in `Memseq` operations**
- **test(memseq): add test for block building timeout**

# Summary

- **Status**: `review`

# Changelog

1. Updated the test suite to cover scenarios involving testing memseq concurrency control mechanisms.
2. Added `futures` to dev-dependencies in `Cargo.toml` and `Cargo.lock` to test asynchronous operations using FuturesUnordered.
3. Made syntax and format adjustments for better code clarity and maintainability. `rustfmt +nightly`

# Testing

The changes have been covered with tests, particularly focusing on:
- **Concurrency control**: Test the new setup under concurrent access scenarios
  using `FuturesUnordered` and ensure correct transaction handling.
- **Error Propagation**: Ensure errors are propagated appropriately when
  operations fail, using mock implementations.
- **Initialization and setup**: Verify that `Memseq` can be initialized
correctly and handle paths and environments as expected.
- **Transaction Processing**: Transactions should be correctly published and
  awaited within blocks under various configurations (e.g., block size,
  timeouts).

# Outstanding issues

- **Integration Testing**: Further end-to-end integration testing with other
  system components is essential to ensure full system compatibility and
  performance adherence.
- **Documentation Updates**: Additional documentation detailing the changes in
  system behavior and setup due to these changes should be prepared.
- **Code**: There are some techniques used, such as sleeping for 1ms to give
  control back to the runtime, which could be optimized. Furthermore, the choice
  has been made to use `anyhow::Error` in a library, which is not idiomatic. We
  could use a custom Error type using the `thiserror` crate.

### DIFF
--- a/protocol-units/sequencing/memseq/src/lib.rs
+++ b/protocol-units/sequencing/memseq/src/lib.rs
@@ -112,6 +112,29 @@ pub mod test {
 	use tempfile::tempdir;
 
 	#[tokio::test]
+	async fn test_wait_for_next_block_building_time_expires() -> Result<(), anyhow::Error> {
+		let dir = tempdir()?;
+		let path = dir.path().to_path_buf();
+		let memseq = Memseq::try_move_rocks(path)?.with_block_size(10).with_building_time_ms(500);
+
+		// Add some transactions
+		for i in 0..5 {
+			let transaction = Transaction::new(vec![i as u8]);
+			memseq.publish(transaction).await?;
+		}
+
+		// Wait for the block to be built, not enough transactions as such
+		// the building time should expire
+		let block = memseq.wait_for_next_block().await?;
+		assert!(block.is_some());
+
+		let block = block.ok_or(anyhow::anyhow!("Block not found"))?;
+		assert_eq!(block.transactions.len(), 5);
+
+		Ok(())
+	}
+
+	#[tokio::test]
 	async fn test_publish_error_propagation() -> Result<(), anyhow::Error> {
 		let mempool = Arc::new(RwLock::new(MockMempool));
 		let parent_block = Arc::new(RwLock::new(Id::default()));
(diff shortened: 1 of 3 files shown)


protocol-units/sequencing/memseq/src/lib.rs
```diff
@@ -1,241 +1,469 @@
...
 #[derive(Clone)]
-pub struct Memseq<T : MempoolBlockOperations + MempoolTransactionOperations> {
-    pub mempool : Arc<RwLock<T>>,
-    // this value should not be changed after initialization
-    block_size : u32,
-    pub parent_block : Arc<RwLock<Id>>,
-    // this value should not be changed after initialization
-    building_time_ms : u64,
+pub struct Memseq<T: MempoolBlockOperations + MempoolTransactionOperations> {
+	pub mempool: Arc<RwLock<T>>,
```
mzabaluev, line 10: Can we settle on indentation with spaces (I think the default with rustfmt)? This does not look like it does in the editor.

0xmovses: I believe it's because of the rule `hard_tabs = true` set here `https://github.com/movementlabsxyz/movement/blob/main/rustfmt.toml` and probably your editor is using the stable channel for `fmt`.

I would personally vote for using `cargo +nightly` for fmt. And then we all set our IDE's to use that. But happy to let the rest of the team weigh in on this.

l-monninger (approved):
Added tests look good to me!

protocol-units/sequencing/memseq/src/lib.rs
cde8: Yes, the setting `hard_tabs = true` enforces using tabs instead of spaces. Personally, I also prefer the use of spaces. For now, I will just save using the provided `rustfmt` config. As such, I will not change anything until we have a consensus on which `rustfmt` configuration we are going to use, and then we can do one big pull request to fix the formatting.

0xmovses: Yep we can use spaces and get rid of that rule.

PR #67, opened 2024-05-17
ENG 318  MCR settlement smart contract call

musitdev:
# Summary
- **Status**: `draft` 
- **RFCs**: [RFC MCR](https://github.com/movementlabsxyz/rfcs/blob/d5183c32a5044e63bd4e8801d9682833b67b8101/draft-mcr/draft-mcr.md),
- **Categories**: any of `protocol-units`.
- **Task**: [ENG-318](https://movement.atlassian.net/browse/ENG-318)

# Changelog

Implements the `McrSettlementClientOperations` to send block commitment to the MRC smart contract.
Stream `BlockAccepted` event.

# Testing

Do some Test with Sepolia test network.
Send commitment works.
`BlockAccepted` event is not notified.
I've to do a local test with Anvil to see what is missing to get the notification.

# Outstanding issues

`BlockAccepted` event notification doesn't works.

# Remarks
This PR will merge in eng-326/full-node-commitments because I've started from it. To avoid PR #58 file modification in it.
Pick the PR #65 commit. 
To merge in main the #58 should be merged before and this PR rebased.


0xmovses:
Thanks for this! For testing, could you leave some commands and steps for me to also do the manual tests? Don't hold back from going into details about testing steps if they are manual for now.

musitdev:
For testing, I use an Alchimy account to have an RPC and WS access to Setopia network.
I've deployed a SC on Setopia where you can submit and try to get some notification. I change the MCP.sol to notify AcceptedBlock for each submit.
In eth_stub.rs, I define a test, you've to fill your Setopia private key and Alchimy key where there's the XXX.
The const MRC_CONTRACT_ADDRESS is the address of the smart contract that I deploy. You can use your own and change this address. I've to change this part to allow setting test SC address.
To deploy my SC I've created a default foundry project and put the MCP.sol source to be compiled.
To deploy the SC I use this command:
```
forge script script/Deploy.s.sol --rpc-url https://eth-sepolia.g.alchemy.com/v2/XXXX --private-key XXXXX --broadcast -vvvv --chain sepolia
```

0xmovses:
Thanks @musitdev, can you give me a numbered list of a step-by-step manual testing flow? Just to make sure we are testing in the same manner. 

For eg.
1. set up Alchemy account for Sepolia Network
2. Deploy `MCR.sol` onto Sepolia 
3. Make a call to contract <some-curl-command-here>
4. Update line <line-number> with your private key 

etc.. 

Would appreciate this so I can test thank you!

musitdev:
1) set up Alchemy account for Sepolia Network and create an App for Sepolia
   2) Get Alchemy Sepolia API key. 
   3) Create a Sepolia account with metamask for example and get some Sepolia Eth using a Faucet.
   4) Deploy MCR.sol in Sepolia or Use the already deployed one: 0xBf7c7AE15E23B2E19C7a1e3c36e245A71500e181
   5) Update MRC_CONTRACT_ADDRESS in eth_stub.rs with the SC address you want to connect.
   6) Update the test: `test_send_commitment` in eth_stub.rs, set the signer and api_key with your private key and Api Key (remove the XXX).
   7) Remove the ignore tag of the test and Run the Test with `cargo test test_send_commitment -- --nocapture`.
   8) The test never end because the event notification never arrive, but the commitment submit is done.

musitdev:
The notification works. It was coming from the SC that wasn't initialized correctly so the submit commitment call was aborting. I test with a simple SC and it works.

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,247 @@
+use crate::{AcceptedBlockCommitment, CommitmentStream, McrSettlementClientOperations};
```
l-monninger, line 1: Let's not call this `eth_stub` unless we're using a stub contract (which is a good idea for separating testing concerns).

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,247 @@
...
+		&self,
+		block_commitment: Vec<BlockCommitment>,
+	) -> Result<(), anyhow::Error> {
+		todo!()
+	}
+
+	async fn stream_block_commitments(&self) -> Result<CommitmentStream, anyhow::Error> {
+		//register to contract BlockCommitmentSubmitted event
+
+		let contract = MCR::new(MRC_CONTRACT_ADDRESS.parse().unwrap(), &self.ws_provider);
```
l-monninger, line 167: Drop these unwraps please.

l-monninger:
So, for these tests, I'd actually like for us to move to a similar structure to what is used for `m1-da-light-node`, `monza-full-node`, etc.

That is, we orchestrate the tests using the pattern provided in [scripts](https://github.com/movementlabsxyz/movement/tree/main/scripts). We use `process-compose` overwrites to define doing this for different networks. See an example with `arabica`, `mocha`, and `local` Celestia networks here: https://github.com/movementlabsxyz/movement/tree/main/process-compose/m1-da-light-node

`anvil` should generally be the first and most commonly used test. 

`kurtosis` should be used within our dynamic environments for long running tests, collaboration, and contention.

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,247 @@
...
+			let call_builder = base_call_builder.clone().gas(estimate_gas);
+			//send the Tx and wait for 2 confirmation.
+			let pending_tx = match call_builder.send().await {
+				Err(alloy_contract::Error::TransportError(TransportError::ErrorResp(payload))) => {
+					match payload.code {
+						//transaction underpriced
+						-32000 => {
+							if payload.message.contains("transaction underpriced") {
+								//increase gas of 10% and retry
+								estimate_gas += (estimate_gas * 10) / 100;
```
l-monninger, line 120: It would be nice to extract this out into a strategy.

protocol-units/settlement/mcr/client/src/stub.rs
```diff
@@ -104,35 +103,35 @@ pub mod test {
...
-            commitment2.clone(),
-        ]).await.unwrap();
-        let guard = client.commitments.write().await;
-        assert_eq!(guard.get(&1), Some(&commitment));
-        assert_eq!(guard.get(&2), Some(&commitment2));
-        Ok(())
-    }
-	
+	#[tokio::test]
+	async fn test_post_block_commitment_batch() -> Result<(), anyhow::Error> {
```
l-monninger, line 113: It will be nice to have antagonistic tests at some point, though this may only be reasonable with the a stub contract implemented.

l-monninger (changes requested):

l-monninger:
https://github.com/movementlabsxyz/movement/pull/67/commits/4af0b0ef18eaea3102ed76269dad6eca0bea22e4

@musitdev I added an example of how to orchestrate the tests. You'll want to trace downwards from the `mcr` task in the `justfile`.

To actually have this work with your tests, you'll need to pick up some private keys from the $MCR_ANVIL_APP_PATH. You can see how I do this when deploying in the `deploy-mcr-to-anvil` script.

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: Yes, but I've no idea that seems to me good. That why I use `eth_strb`. If you can propose something.

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,307 @@
...
+			>,
+			RootProvider<BoxTransport>,
+			BoxTransport,
+			Ethereum,
+		>,
+		BoxTransport,
+		Ethereum,
+	>
+{
+	pub async fn build_with_urls<S2>(
```
musitdev, line 85: Builder with url and private key.

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,257 @@
...
+
+impl<P: Provider<T, Ethereum>, T: Transport + Clone> McrEthSettlementClient<P, T> {
+	pub async fn build_with_provider<S>(
+		rpc_provider: P,
+		ws_url: S,
+		gas_limit: u128,
+	) -> Result<Self, anyhow::Error>
+	where
+		S: Into<String>,
+	{
```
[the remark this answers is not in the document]
musitdev: Yes, it was done, but for an unknown reason, the commit wasn't pushed. I push them and now they are present.

protocol-units/settlement/mcr/client/src/eth_stub.rs
```diff
@@ -0,0 +1,257 @@
...
+	async fn post_block_commitment(
+		&self,
+		block_commitment: BlockCommitment,
+	) -> Result<(), anyhow::Error> {
+		let contract = MCR::new(MRC_CONTRACT_ADDRESS.parse().unwrap(), &self.rpc_provider);
+		let call_builder = contract.createBlockCommitment(
+			U256::from(block_commitment.height),
+			alloy_primitives::FixedBytes(block_commitment.commitment.0[..32].try_into()?),
+			alloy_primitives::FixedBytes(block_commitment.block_id.0[..32].try_into()?),
+		);
```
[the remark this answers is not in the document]
musitdev: Yes, I wasn't sure which to use.

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: OK I try to do some refactoring and define some rules and associated processing.

musitdev (commented):
I see your test update and how it's plug inside the project test process. I finish the unit test so that it works locally and will see how it is run by this process.

0xmovses:
Let me know @musitdev once this is ready to review 🚀

musitdev:
I'm still with the tx gas issue. I have the same be behavior with Sepolia.
You can start your review to see if you don't have some main remarks about the code. Most of the code won't change.
I need to update the rule part of the Tx gas error detection, but it's linked with my current test. All the other part won't change.
After I'll finish the 2 other trait methods that will be similar to the current one.

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: I call it `eth_client.rs` and move all eth Tx send code in `send_eth_tx.rs `file.

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: I've created a simple rule based system to process send tx errors (see in send_eth_tx.rs).

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: Done

protocol-units/settlement/mcr/client/src/stub.rs
musitdev: I test the commitment batch with the other calls in one unit test because all the smart contract call must be done in the right order.

musitdev:
I've done all my changes, the PR can be reviewed.

l-monninger:
Did the toolchain get changed in the nix flake?: https://github.com/movementlabsxyz/movement/actions/runs/9257011400/job/25464244473

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.75"
+channel = "1.78"
```
l-monninger, line 2: We need to modify the `flake.nix` to handle this. Is there anything that actually depends on this? I assume this forced by `alloy`?

scripts/mcr/deploy-mcr-to-anvil
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
+
```
l-monninger, line 2: Just a note, that this is getting refactored to use a file standard.

protocol-units/settlement/mcr/client/src/send_eth_tx.rs
```diff
@@ -0,0 +1,137 @@
...
+	_kind: PhantomData<Kind>,
+}
+
+impl<Kind> SendTxErrorRule<Kind> {
+	pub fn new() -> Self {
+		SendTxErrorRule { _kind: PhantomData::default() }
+	}
+}
+
+// Define the current 2 errors managed.
```
l-monninger, line 27: Let's use https://docs.rs/thiserror/latest/thiserror as with #69 that will be predominant elsewhere in the workspace.

protocol-units/settlement/mcr/contracts/script/DeployMCR.s.sol
```diff
@@ -0,0 +1,38 @@
...
+        MCR mcr = new MCR(
+            5, 
+            128,
+            100 ether, // should accumulate 100 ether
+            100 ether, // each genesis validator can stake up to 100 ether
+            0
+        );
+        vm.stopBroadcast();
+
+        // Comment because the Genesis ceremony works (Assert ok)
```
l-monninger, line 19: Do we want to keep this in?

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -0,0 +1,536 @@
...
+		// use tracing_subscriber::EnvFilter;
+
+		// tracing_subscriber::fmt()
+		// 	.with_env_filter(
+		// 		EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+		// 	)
+		// 	.init();
+
+		// Inititalize Test variables
+		let rpc_port = env::var("MCR_ANVIL_PORT").unwrap();
```
l-monninger, line 294: We should pull this out into a config pattern similar to that used elsewhere in the workspace. Save for a future ticket.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -0,0 +1,536 @@
...
+#[cfg(test)]
+pub mod test {
+	use super::*;
+	use alloy_provider::ProviderBuilder;
+	use alloy_signer_wallet::LocalWallet;
+	use movement_types::Commitment;
+	use std::env;
+
+	// Define 2 validators (signer1 and signer2) with each a little more than 50% of stake.
+	// After genesis ceremonial, 2 validator send the commitment for height 1.
```
l-monninger, line 278: Should be "genesis ceremony."

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -0,0 +1,536 @@
...
+			.signer(EthereumSigner::from(signer1))
+			.on_http(rpc_url.parse().unwrap());
+
+		let config = McrEthSettlementConfig {
+			mrc_contract_address: mcr_address.to_string(),
+			gas_limit: DEFAULT_TX_GAS_LIMIT,
+			tx_send_nb_retry: MAX_TX_SEND_RETRY,
+		};
+
+		let client1 = McrEthSettlementClient::build_with_provider(
```
l-monninger, line 320: Yeah, so eventually, what I'll want to have is a `try_from_env` constructor defined that would wrap this all up. But, for now, this is good.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -0,0 +1,536 @@
...
+		let event =
+			tokio::time::timeout(tokio::time::Duration::from_secs(5), client1_stream.next())
+				.await
+				.unwrap()
+				.unwrap()
+				.unwrap();
+		assert_eq!(event.commitment.0[0], 5);
+		assert_eq!(event.block_id.0[0], 4);
+		let event =
+			tokio::time::timeout(tokio::time::Duration::from_secs(5), client2_stream.next())
```
l-monninger, line 396: Even in tests, I would prefer not use unwrap as it just makes everything more verbose.

l-monninger (changes requested):

l-monninger:
@musitdev I'll take a look at the issue with the `flake.nix`. Can you resolve the above and handle the merge?

l-monninger:
It looks like the `flake.nix` got reverted.

l-monninger:
`anvil` is missing.

l-monninger:
It's using an older foundry installation that doesn't work.

l-monninger:
And, it doesn't show in the commit history, I'm guessing because of the force push.

protocol-units/settlement/mcr/client/src/eth_stub.rs
musitdev: done

protocol-units/settlement/mcr/contracts/script/DeployMCR.s.sol
musitdev: I don't understand. From Slack discussion we decide to use forge script to manage smart contract init and not in Rust. It's your question?

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: It uses the var define in mcr-env, it's not the right pattern?

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: I'm mostly agreeing with you, but I found that unwrap failure is more informative that using a ?.
For example, when getting an environment variable with unwrap:
```
---- eth_client::test::test_send_commitment stdout ----
thread 'eth_client::test::test_send_commitment' panicked at protocol-units/settlement/mcr/client/src/eth_client.rs:294:52:
called `Result::unwrap()` on an `Err` value: NotPresent
```
And with ?
```
failures:

---- eth_client::test::test_send_commitment stdout ----
Error: environment variable not found
```
To find the issue, it's easier when you have the line in the file that why I keep them when you can have, or I add issues.
Perhaps we can format the `?` output in a more informative way. Using RUST_BACKTRACE=1 you have all the stack trace, but with Tokio it's not always easy to find the right root place.
 It's like you prefer, I can replace with `???`

musitdev:
About 

> And, it doesn't show in the commit history, I'm guessing because of the force push.
I use the force push because after a rebase the push is always rejected, I have to re-apply all the rebase conflict update.
I've done:
 - git pull 
 -  git rebase
 - correct rebase conflict
 - git push that fail.
 That why I force the push. Perhaps you've a better way to do and avoid reapplying all conflict change or use a force?

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: I've implemented one version.

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: Done

rust-toolchain.toml
musitdev: I set to 1.76 because of Alloy. I didn't know about nix overlay.

0xmovses: Is it possible to stay on `1.75`? I ask for hyperlane fork comptability. We need `movement` to run in the same env as `hyperlane` binaries, which is on `1.75`.

musitdev: Alloy need 1.76 that why I change the version. To allow using different compilers, we can remove settlement/mcr/client from the workplace to have its own build and own compiler version.

l-monninger:
@musitdev The CI needs to run the e2e mcr tests.

l-monninger:
@musitdev Nix takes care of `anvil` dependencies. But, your CI update is still not running the e2e tests.

You should add:

```
CELESTIA_LOG_LEVEL=FATAL MOVEMENT_PREBUILT=false nix develop --command bash  -c "just mcr test.local -t=false"
```

And, if this isn't passing locally, that means there is still work to do.

l-monninger (approved):
LGTM

PR #69, opened 2024-05-21
ENG-400 introducing error types Mempool{Block,Transaction}OperationsError

cde8:
- **refactor(movement-types): Refactor error handling with `thiserror`**
- **refactor: introducing error types Mempool{Block,Transaction}OperationsError**

# Summary
- **Status**: `draft`

This PR involves changes across multiple components within the `protocol-units`
category, focusing on updating the error handling from `anyhow` to `thiserror`
and integrating transaction and block operations in the mempool and sequencer
protocols. Refactoring has introduced new error types for more granular error
management and modified transaction and block operation interfaces to be
asynchronous and more robust in error handling.

# Changelog
- Refactored error handling from `anyhow` to `thiserror` for clearer and structured errors.
- Updated Mempool and Sequencer interfaces to use the new error types introduced.
- Added new interfaces and updates for error handling in mempool-related operations, enhancing support for async operations.
- Added a convenience trait DbExt to refactor the RocksDb Mempool{Block,Transaction}Operations impls.
- Changed dependencies in Cargo.toml files from `anyhow` to `thiserror`.
- Improved method signatures and documentation for the involved modules.

# Testing
- Adjusted existing unit tests to accommodate the changes in error handling and the use of `thiserror`.
- Ensured all affected unit tests are passing with the updated error types and handling mechanisms.

# Outstanding issues
- Additional real-world integration testing is recommended to ensure the
  changes perform well under different operational cases.

### DIFF
--- a/protocol-units/mempool/move-rocks/src/lib.rs
+++ b/protocol-units/mempool/move-rocks/src/lib.rs
@@ -1,277 +1,366 @@
 use mempool_util::{
-    MempoolBlockOperations,
-    MempoolTransactionOperations,
-    MempoolTransaction
+	MempoolBlockOperations, MempoolBlockOperationsError, MempoolBlockOperationsResult,
+	MempoolTransaction, MempoolTransactionOperations, MempoolTransactionOperationsError,
+	MempoolTransactionOperationsResult,
 };
-use rocksdb::{DB, Options, ColumnFamilyDescriptor};
-use std::sync::Arc;
-use serde_json;
-use tokio::sync::RwLock;
 use movement_types::{Block, Id};
-use anyhow::Error;
+use rocksdb::{BoundColumnFamily, ColumnFamilyDescriptor, Options, DB};
+use serde_json;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+#[derive(thiserror::Error, Debug)]
+pub enum RocksdbMempoolError {
+	#[error("Underlying db error")]
+	RocksDbError(#[from] rocksdb::Error),
+	#[error("Family handle missing")]
+	FamilyHandleMissing,
+}
+
+pub type RocksdbMempoolResult<T> = Result<T, RocksdbMempoolError>;
+
+trait DBExt {
+	fn get_from_handle(
+		&self,
+		handle: &str,
+		key: impl AsRef<[u8]>,
+	) -> RocksdbMempoolResult<Option<Vec<u8>>>;
+
+	fn put_to_handle(
+		&self,
+		handle: &str,
+		key: impl AsRef<[u8]>,
+		value: impl AsRef<[u8]>,
+	) -> RocksdbMempoolResult<()>;
+
+	fn delete_from_handle(&self, handle: &str, key: impl AsRef<[u8]>) -> RocksdbMempoolResult<()>;
+
+	fn iter_from_handle(
+		&self,
+		handle: &str,
+	) -> RocksdbMempoolResult<(Arc<BoundColumnFamily>, rocksdb::DBIterator)>;
+}
+
+impl DBExt for rocksdb::DB {
+	fn get_from_handle(
+		&self,
+		handle: &str,
+		key: impl AsRef<[u8]>,
+	) -> RocksdbMempoolResult<Option<Vec<u8>>> {
+		let cf_handle = self.cf_handle(handle).ok_or(RocksdbMempoolError::FamilyHandleMissing)?;
+		self.get_cf(&cf_handle, key).map_err(RocksdbMempoolError::from)
+	}
+
+	fn put_to_handle(
+		&self,
+		handle: &str,
+		key: impl AsRef<[u8]>,
+		value: impl AsRef<[u8]>,
+	) -> RocksdbMempoolResult<()> {
+		let cf_handle = self.cf_handle(handle).ok_or(RocksdbMempoolError::FamilyHandleMissing)?;
+		self.put_cf(&cf_handle, key, value).map_err(RocksdbMempoolError::from)
+	}
+
+	fn delete_from_handle(&self, handle: &str, key: impl AsRef<[u8]>) -> RocksdbMempoolResult<()> {
+		let cf_handle = self.cf_handle(handle).ok_or(RocksdbMempoolError::FamilyHandleMissing)?;
+		self.delete_cf(&cf_handle, key).map_err(RocksdbMempoolError::from)
+	}
+
+	fn iter_from_handle(
+		&self,
+		handle: &str,
+	) -> RocksdbMempoolResult<(Arc<BoundColumnFamily>, rocksdb::DBIterator)> {
+		let cf_handle = self.cf_handle(handle).ok_or(RocksdbMempoolError::FamilyHandleMissing)?;
+		let iterator = self.iterator_cf(&cf_handle, rocksdb::IteratorMode::Start);
+		Ok((cf_handle, iterator))
+	}
+}
 
 #[derive(Debug, Clone)]
 pub struct RocksdbMempool {
-    db: Arc<RwLock<DB>>,
+	db: Arc<RwLock<DB>>,
 }
 impl RocksdbMempool {
-    pub fn try_new(path: &str) -> Result<Self, Error> {
-        let mut options = Options::default();
-        options.create_if_missing(true);
-        options.create_missing_column_families(true);
+	pub fn try_new(path: &str) -> RocksdbMempoolResult<Self> {
+		let mut options = Options::default();
+		options.create_if_missing(true);
+		options.create_missing_column_families(true);
 
-        let mempool_transactions_cf = ColumnFamilyDescriptor::new("mempool_transactions", Options::default());
-        let transaction_truths_cf = ColumnFamilyDescriptor::new("transaction_truths", Options::default());
-        let blocks_cf = ColumnFamilyDescriptor::new("blocks", Options::default());
-        let transaction_lookups_cf = ColumnFamilyDescriptor::new("transaction_lookups", Options::default());
+		let mempool_transactions_cf =
+			ColumnFamilyDescriptor::new("mempool_transactions", Options::default());
+		let transaction_truths_cf =
+			ColumnFamilyDescriptor::new("transaction_truths", Options::default());
+		let blocks_cf = ColumnFamilyDescriptor::new("blocks", Options::default());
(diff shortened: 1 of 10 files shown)


cde8:
Refer to issue #71, as I believe using generics instead of `Box<dyn Error>` would be preferable in this case.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -1,277 +1,366 @@
...
-                // Optionally, remove from the lookup table as well
-                let lookups_cf_handle = db.cf_handle("transaction_lookups").ok_or_else(|| Error::msg("CF handle not found"))?;
-                db.delete_cf(&lookups_cf_handle, tx.transaction.id().to_vec())?;
-                
-                Ok(Some(tx))
-            }
-        }
-
-    }
+	fn iter_from_handle(
```
mzabaluev, line 38: Hmm, this is what I tried to avoid, even though eventually I want do be able to do format on save in any of our source. At least rustfmt is in a separate commit so I can filter it out and review the substantial changes. But this will conflict with many other changes that will land on `main`. We do want to rip the bandaid on the "fmt day" and commit a massive formatting change for the entire code base some time later.

protocol-units/mempool/util/src/lib.rs
```diff
@@ -1,172 +1,224 @@
...
+#[derive(Error, Debug)]
+pub enum MempoolTransactionOperationsError {
+	#[error("Serialization error: {0}")]
+	SerializationError(String),
+	#[error("Deserialization error: {0}")]
+	DeserializationError(String),
+	#[error("Underlying store error")]
+	StoreError(#[from] BoxedStoreError),
+	#[error("Mock error: {0}")]
+	MockError(String),
```
mzabaluev, line 18: Why expose mock errors specially as part of the API?
The mock is supposed to be a stand-in for the real component and as such should not generate any special errors or code paths, should it?

protocol-units/mempool/util/src/lib.rs
```diff
@@ -1,172 +1,224 @@
...
 
-use movement_types::{Block, Transaction, Id};
+use movement_types::{Block, Id, Transaction};
 use std::cmp::Ordering;
+use thiserror::Error;
+
+pub type BoxedStoreError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+#[derive(Error, Debug)]
+pub enum MempoolTransactionOperationsError {
```
mzabaluev, line 10: Let's make naming less verbose and erase `Operations` everywhere except the trait names.
```suggestion
pub enum MempoolTransactionError {
```

protocol-units/sequencing/memseq/src/lib.rs
```diff
@@ -404,65 +413,78 @@ pub mod test {
...
+
 	struct MockMempool;
 	impl MempoolTransactionOperations for MockMempool {
 		async fn has_mempool_transaction(
 			&self,
 			_transaction_id: Id,
-		) -> Result<bool, anyhow::Error> {
-			Err(anyhow::anyhow!("Mock has_mempool_transaction"))
+		) -> MempoolTransactionOperationsResult<bool> {
+			mock_error!("has_mempool_transaction")
```
mzabaluev, line 432: This should just be left at `todo!` until there is a meaningful mocked implementation that is capable of simulating real success and error cases.

mzabaluev (changes requested):

protocol-units/mempool/util/src/lib.rs
l-monninger: I agree with this. `-Operations` should only be a suffix for traits and only where there's not another obvious name.

protocol-units/mempool/move-rocks/src/lib.rs
cde8: I am using the standard rustfmt config and tooling from the Nix environment. I agree that we need to fix this formatting issue once and for all. Until then I am just using the rustfmt configuration as provided.

protocol-units/mempool/util/src/lib.rs
cde8: Correct, placed it in to `#[cfg(test)]` it out in a later stadium.

protocol-units/sequencing/memseq/src/lib.rs
cde8: @mzabaluev why should this just be left at `todo!`? We prefer to `panic!` instead of returning an Err

protocol-units/mempool/util/src/lib.rs
cde8: We do not want to have error variants in the public API. One solution is to use the "Other" variant as an error variant, which can be utilized in cases where one wishes to return a string error. Pushed it and consider it as resolved.

protocol-units/mempool/move-rocks/src/lib.rs
mzabaluev: OK, I just did not enable auto-formatting in my IDE to avoid the massive changes.
After the big fmt commit, we all can enable "format on save" and similar settings and not have to deal with these problems.

protocol-units/mempool/util/src/lib.rs
mzabaluev: The trait names are indeed a bit verbose too, so maybe we can come up with terser names all around.
I don't see why errors need to be prefixed by the trait name, though. If anything, I'd like to see package names used instead of verbose prefixes, as this is more in Rust style. So you could have `mempool::TransactionOperations` (re-exported from `mempool::transaction`) and `mempool::transaction::Error`.

protocol-units/sequencing/memseq/src/lib.rs
mzabaluev: If the mock cannot provide a useful implementation to support the tests at this point, it's better to panic up front, which is what `todo!` does. When we have code under test that actually calls the method, we'll naturally need to supply something life-like, rather than a "mock error", from the mock side.

protocol-units/mempool/util/src/lib.rs
```diff
@@ -1,172 +1,224 @@
 use serde::{Deserialize, Serialize};
 
-use movement_types::{Block, Transaction, Id};
+use movement_types::{Block, Id, Transaction};
 use std::cmp::Ordering;
+use thiserror::Error;
+
+pub type BoxedStoreError = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+#[derive(Error, Debug)]
```
mzabaluev, line 9: It's probably also good to annotate the error enums with `#[non_exhaustive]` so that adding new variants would not break consumer code.

protocol-units/mempool/move-rocks/src/lib.rs
cde8: That is also an option indeed. Will do the same on my following changes.

protocol-units/mempool/util/src/lib.rs
cde8: I will leave it for now, and after we merge #71, we can settle on a naming convention for these public APIs.

protocol-units/mempool/util/src/lib.rs
cde8: Good point, added

0xmovses:
It's really hard to review with the fmt thing. Can we see what happens if you don't dev in the nix shell and just use defauly IDE settings. And don't run any fmt commands. Auto-format should be set to true in your IDE and using `rust-analyzer` to format. See if that brings the diff in line.

cde8:
> It's really hard to review with the fmt thing. Can we see what happens if you don't dev in the nix shell and just use defauly IDE settings. And don't run any fmt commands. Auto-format should be set to true in your IDE and using `rust-analyzer` to format. See if that brings the diff in line.

Yes, I disabled the formatting for the future, and after the big `rust fmt` pull request, we should not run into this issue again as we all should be aligned. @0xmovses: Is it an idea to have a peer session so we can run through all the changes together?

protocol-units/sequencing/memseq/src/lib.rs
```diff
@@ -148,14 +148,14 @@ pub mod test {
 		assert!(result.is_err());
 		assert_eq!(
 			result.unwrap_err().to_string(),
-			"MempoolTransactionOperationsError error: Mock error: add_transaction"
+			"MempoolTransactionOperationsError error: Other error: add_transaction"
```
mzabaluev, line 151: Should we eventually have a success path test as well?
I still don't see the added value of the `Other` error variant if it can only be produced by the mock.

cde8: "Other" can be used by a consumer of the trait if it falls outside of the variants of the error enum.

networks/monza/monza-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : MonzaExecutor + Send + Sync + Clone>MonzaPartialNode<T> {
...
+		.into_inner();
+
+		while let Some(blob) = stream.next().await {
+			println!("Stream hot!");
+			// get the block
+			let block_bytes = match blob?
+				.blob
+				.ok_or(anyhow::anyhow!("No blob in response"))?
+				.blob_type
+				.ok_or(anyhow::anyhow!("No blob type in response"))?
```
l-monninger, line 115: Just to confirm, are we planning to address these internal error types in a later PR?

networks/suzuka/suzuka-config/src/lib.rs
```diff
@@ -1,36 +1,44 @@
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum ConfigError {
+	#[error("Could not read the Environment variables")]
+	EnvReadError,
+	#[error("Could not write the Environment variables")]
+	EnvWriteError,
+	#[error("Could not write the Bash export string")]
```
l-monninger, line 9: I feel like we probably want to report which variables causes the error.

l-monninger (changes requested):
Looks good! Couple comments.

networks/suzuka/suzuka-config/src/lib.rs
cde8: I agree, added them.

networks/monza/monza-full-node/src/partial.rs
cde8: Yes these will be added in a later PR. Added to my todo list

PR #71, opened 2024-05-21
ENG-400: Generalized `protocol-unit-types`, error handling upgrade and `async-trait` refactoring

cde8:
# Summary
- **Status**: `open`

A series of refactoring changes aimed at enhancing code maintainability,
adaptability, and fixing warnings throughout several components in our
software repository. Changes include dependency updates, trait adjustments,
import optimizations, and error handling transformations. 

# Changelog
- The `async-trait` dependency has been added across multiple crates to preserve
  compatibility with older versions.
- Generalized `Executor` and other traits with `protocol-unit-types` for more
  generalized abstraction.
- Suppressed and fixed warnings across different modules, less noise when compiling.
- Modified error handling with generics, allowing compile time matching on error types.
- Various dependencies and imports in `Cargo.toml` and `Cargo.lock` updated or
  removed to reflect current needs.
- Introduced the non_exhaustive attribute to error enums to protect against
  future breaking changes.
- Enhanced trait definitions with `async_trait` for harmonization and to meet
  Rust's requirements for async interfaces.
- Cleanup of code base by removing redundant types and imports, generalizing
  error handling.

# Testing
- Runing compilation and unit tests to ensure no disruptions with trait and
  async modifications.

# Outstanding issues
- More interfaces can now be placed into `protocol-unit-types`
- Future considerations for extending `non_exhaustive` to more enums as the codebase evolves.


mzabaluev:
I'm not sure the error type needs to be a generic parameter of any trait. It's usually the output type decided by the implementation, and as such it should be an associated type? See for example `TryFrom`, `FromStr`.

cde8:
Hey, thanks for the feedback! Do we need the flexibility to use generics?: Are we expecting to have multiple implementations of traits like `MempoolTransactionOperations`, `MempoolBlockOperations`, `Sequencer`, or `SharedSequencer` for the same struct with different error types? If not, then we should go for an associated type.

musitdev:
Error with Generic is painful because it propagates everywhere, so If we can avoid it's better. If it's different error, we should use different variant or type.

cde8:
My first implementation uses a `StoreError` with a boxed error. This complicates the code when implementing a trait, as it requires a `map_err` to convert to the desired error type. This adds complexity at runtime and involves some allocations, which is not ideal. See #69, and look at the implementation, after refactoring: https://github.com/movementlabsxyz/movement/blob/c26bfe7fe284f29b37d38840955b09f9fbce5445/protocol-units/mempool/move-rocks/src/lib.rs#L149-L163

When using a generic, there is propagation, which is somewhat limited due to the use of associated types. However, what we gain is compile-time verification, no allocations, and an explicit definition of the error type. We get a cleaner implementation at #71, see: https://github.com/movementlabsxyz/movement/blob/8e019382544aab9de0df9d96a2780a935e20dbf8/protocol-units/mempool/move-rocks/src/lib.rs#L148-L162

Personally, I like the latest version using associated types, due to reasons above. Once we settle on our preference, I will implement that methodology in the rest of the public APIs.

networks/suzuka/suzuka-config/src/lib.rs
```diff
@@ -1,36 +1,44 @@
...
 }
 
 impl Config {
+	pub fn new(execution_config: maptos_execution_util::config::Config) -> Self {
+		Self { execution_config }
+	}
+
+	pub fn try_from_env() -> Result<Self, ConfigError> {
+		let execution_config = maptos_execution_util::config::Config::try_from_env()
+			.map_err(|_| ConfigError::EnvReadError)?;
```
0xmovses, line 25: If you had `maptos_execution_util::config::Config::try_from_env()` return `Result<Self, ConfigError>`, as opposed to `Result<Self, anyhow::Error>`, I don't think you would have to call `map.err` and could just `?` directly on `try_from_env`

networks/suzuka/suzuka-config/src/lib.rs
```diff
@@ -1,36 +1,44 @@
...
+
+		Ok(Self { execution_config })
+	}
+
+	pub fn write_to_env(&self) -> Result<(), ConfigError> {
+		self.execution_config.write_to_env().map_err(|_| ConfigError::EnvWriteError)?;
+		Ok(())
```
0xmovses, line 32: Same comment with all the `map_err`'s here.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -1,45 +1,40 @@
...
-
+use tokio_stream::StreamExt;
 
 #[derive(Clone)]
-pub struct SuzukaPartialNode<T : SuzukaExecutor + Send + Sync + Clone> {
-    executor: T,
-    transaction_sender : Sender<SignedTransaction>,
-    pub transaction_receiver : Receiver<SignedTransaction>,
-    light_node_client: Arc<RwLock<LightNodeServiceClient<tonic::transport::Channel>>>,
+pub struct SuzukaPartialNode<T: ExecutorOps + Send + Sync + Clone> {
```
0xmovses, line 18: Let's add struct level doc comment and a doc comment for all its fields.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
...
+
+		let mut transactions = Vec::new();
+
+		while let Ok(transaction_result) =
+			tokio::time::timeout(Duration::from_millis(100), self.transaction_receiver.recv()).await
+		{
+			match transaction_result {
+				Ok(transaction) => {
+					println!("Got transaction: {:?}", transaction);
+					let serialized_transaction = serde_json::to_vec(&transaction)?;
```
0xmovses, line 65: swap `println` for `log:info!`

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
...
+			if std::time::Instant::now() > end_time {
+				break;
+			}
+		}
+
+		if transactions.len() > 0 {
+			let client_ptr = self.light_node_client.clone();
+			let mut light_node_client = client_ptr.write().await;
+			light_node_client.batch_write(BatchWriteRequest { blobs: transactions }).await?;
+			println!("Wrote transactions to DA");
```
0xmovses, line 82: Swap println from `log::info!`

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
...
+			let client_ptr = self.light_node_client.clone();
+			let mut light_node_client = client_ptr.write().await;
+			light_node_client
+				.stream_read_from_height(StreamReadFromHeightRequest { height: block_head_height })
+				.await?
+		}
+		.into_inner();
+
+		while let Some(blob) = stream.next().await {
+			println!("Stream hot!");
```
0xmovses, line 109: Swap again for log.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
...
+			{
+				blob_response::BlobType::SequencedBlobBlock(blob) => blob.data,
+				_ => {
+					anyhow::bail!("Invalid blob type in response")
+				},
+			};
+
+			// get the block
+			let block: Block = serde_json::from_slice(&block_bytes)?;
+			println!("Received block: {:?}", block);
```
0xmovses, line 125: Again

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -54,180 +49,145 @@ impl <T : SuzukaExecutor + Send + Sync + Clone>SuzukaPartialNode<T> {
...
+			hasher.update(&block_bytes);
+			let slice = hasher.finalize();
+			let block_hash = HashValue::from_slice(slice.as_slice())?;
+
+			// form the executable block and execute it
+			let executable_block = ExecutableBlock::new(block_hash, block);
+			let block_id = executable_block.block_id;
+			self.executor.execute_block(FinalityMode::Opt, executable_block).await?;
+
+			println!("Executed block: {:?}", block_id);
```
0xmovses, line 151: Again

0xmovses:
We're going to use `tracing` instead of log https://docs.rs/tracing/latest/tracing/

networks/monza/monza-full-node/Cargo.toml
```diff
@@ -14,14 +14,18 @@ rust-version = { workspace = true }
 [dependencies]
 monza-executor = { workspace = true }
 m1-da-light-node-client = { workspace = true }
+protocol-unit-types = { workspace = true }
+
+movement-types = { workspace = true }
```
mzabaluev, line 19: Could we consolidate these two crates? `movement-types` has some common definitions as well, but it's under `util`.

protocol-units/types/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "protocol-unit-types"
```
mzabaluev, line 2: As far as I understand @l-monninger's organization, `protocol-units` is a directory name used to group crates in the workspace, so it should not figure in the global-ish crate names.

This directory is also meant for protocol crates, but I'm not sure about common data definitions. There is also already `movement-types` fulfilling a similar role. Maybe we should coalesce the two as suggested in the other comment.

mzabaluev: To reduce the confusion, a description here, or maybe the top-level crate doc should describe the purpose of each crate.

protocol-units/types/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+	block_executor::partitioner::ExecutableTransactions,
+	transaction::signature_verified_transaction::SignatureVerifiedTransaction,
+	transaction::{SignedTransaction, Transaction},
+};
+
+pub mod executor;
+pub mod finality_mode;
+
+pub use executor::ExecutorOps;
+pub use finality_mode::FinalityMode;
```
mzabaluev, line 14: For now this looks like a grab bag of re-exports, and an abstract crate definition, for everything related to executors. Is it the sole purpose of the crate?

mzabaluev, line 14: We've talked with @l-monninger about `FinalityMode`; it's a leftover from an older design and should be removed. Maybe not in this PR, though.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: This and similar printouts were addressed in #58, let's just merge that and sort out any other `println` macro uses that remain.

networks/monza/monza-full-node/src/partial.rs
```diff
@@ -1,45 +1,40 @@
 use std::{sync::Arc, time::Duration};
 
+use crate::*;
```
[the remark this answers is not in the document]
mzabaluev: Agreed, let's not introduce more glob imports than there are already, except maybe `use super::*` in unit test modules.

Cargo.toml
```diff
@@ -34,12 +36,15 @@ license = "MIT OR Apache-2.0"
 authors = ["Movement Labs"]
 repository = "www.github.com/movementlabs/movement"
 homepage = "www.homepage.com"
-resolver = "2"
 publish = false
```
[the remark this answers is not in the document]
cde8: It should not be in that section. It should be in another part of the `Cargo.toml` file, I believe it was defined there. I already disabled the `rustfmt on save` for now.

networks/monza/monza-full-node/src/partial.rs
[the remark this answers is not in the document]
cde8: This is not my addition, I already removed a few of them. Let me fix that one as well.

networks/monza/monza-full-node/src/partial.rs
```diff
@@ -1,45 +1,40 @@
...
-
+use tokio_stream::StreamExt;
 
 #[derive(Clone)]
-pub struct MonzaPartialNode<T : MonzaExecutor + Send + Sync + Clone> {
-    executor: T,
-    transaction_sender : Sender<SignedTransaction>,
-    pub transaction_receiver : Receiver<SignedTransaction>,
-    light_node_client: Arc<RwLock<LightNodeServiceClient<tonic::transport::Channel>>>,
+pub struct MonzaPartialNode<T: ExecutorOps + Send + Sync + Clone> {
```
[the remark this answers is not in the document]
cde8: I tried to remove it, it seems we need it.

networks/monza/monza-full-node/src/partial.rs
```diff
@@ -1,45 +1,40 @@
...
-    pub fn new(executor : T, light_node_client: LightNodeServiceClient<tonic::transport::Channel>) -> Self {
-        let (transaction_sender, transaction_receiver) = async_channel::unbounded();
-        Self {
-            executor : executor,
-            transaction_sender,
-            transaction_receiver,
-            light_node_client : Arc::new(RwLock::new(light_node_client)),
-        }
-    }
+impl<T: ExecutorOps + Send + Sync + Clone> MonzaPartialNode<T> {
```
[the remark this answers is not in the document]
cde8: It seems we need it, I get an error when I remove it.

networks/suzuka/suzuka-full-node/src/partial.rs
cde8: Not my addition, but I am assuming you want a tracing. I am going to remove it and put a tracing::trace! in its place. We need to have a discussion on how to determine the different logging levels (info, debug, trace). There is an art to it.

networks/suzuka/suzuka-full-node/src/partial.rs
cde8: Will use  `tracing::trace!` calls, as we do not want this level of granularity at the info level.

networks/suzuka/suzuka-full-node/src/partial.rs
cde8: Ok, then I will merge.

networks/suzuka/suzuka-full-node/src/partial.rs
cde8: These are resolved in a previous PR, will merge instead of change.

networks/suzuka/suzuka-full-node/src/partial.rs
cde8: These are resolved in a previous PR, will merge instead of change.

networks/monza/monza-full-node/Cargo.toml
cde8: I was thinking the same when creating the new crate. I went for the protocol-units granularity because some enums are only used in the protocol-units. I will merge for now; we can consolidate them later if needed.

protocol-units/types/Cargo.toml
cde8: When reading the code, I would like to see a prefix on all crates indicating they are part of the movement ecosystem. I noted this to discuss in a later meeting. We should come up with a naming convention. I agree.

protocol-units/types/src/lib.rs
cde8: The purpose of this crate is to allow consumers of the protocol units to include this types crate to get access to all the types that the interfaces or traits use. This way, we can add new protocol units and have the types shared.

cde8: I was not aware. Will leave it in for now.

PR #78, opened 2024-05-24
chore: remove Status field from PR template

mzabaluev:
# Summary
- **Categories**: `misc`.

As currently intended, the Status text field will duplicate the actual status as managed by GitHub, so it's either double work or prone to be out of date.


l-monninger:
Intent here was to reflect PRs that are suspended vs. abandoned as discussed in person. But, we can use labels instead.

l-monninger:
We’re actually probably better off using labels for categories as well once we have our act together.

l-monninger (approved):

PR #79, opened 2024-05-25
init optimization comments

0xPrimata:
# Summary
- **Status**: `draft`

Notes and suggestions regarding gas optimization.

# Changelog

Suggests utilizing methods that probably will minimize gas consumption. Especially:
* allocating to memory states that are reutilized
* Using custom errors instead of require
* decrease evaluations per call by sticking to simple GT or LT.

# Testing

No tests implemented, those should be next to define truthfulness of suggestions above.

# Outstanding issues


l-monninger:
For those who aren't familiar, EVM does not have opcodes for LTE and GTE: https://www.ethervm.io/#10

0xPrimata:
Within my tests, there are some unexpected results. I'll be investigating that, especially the absurdly high gas consumption for the current get functions that simply return states and using states during execution is consuming less gas then storing those variables in memory first then utilizing it through execution.

l-monninger:
Closing this as we resume discussion elsewhere.

PR #86, opened 2024-05-29
fix: add cors.

l-monninger:
# Summary
- **Status**: `draft` 

WIP. I believe this has fallen out a few times. It is needed for local testing with the browser OFC. 

# Changelog

# Testing

# Outstanding issues


0xPrimata:
@l-monninger right below line 366 it requires `.allow_credentials(true);`
explorer is able to read it if its added 🙏🏽 
image

0xPrimata (approved):

l-monninger:
@0xPrimata there are a few more policies they set in the upstream. I will check which apply and add as needed before we merge here.

l-monninger:
This is the note from upstream: https://github.com/movementlabsxyz/aptos-core/blob/ea91067b81f9673547417bff9c70d5a2fe1b0e7b/api/src/runtime.rs#L209

l-monninger:
@0xPrimata  If the above passes I will unmark as draft, then we can drop the other reviewers and merge.

0xPrimata:
`.allow_methods(vec![Method::GET, Method::POST]);` isn't really necessary since it's reducing the scope of possible methods. All methods are allowed by default. Maybe good to add though.
https://docs.rs/poem/latest/poem/middleware/struct.Cors.html#method.allow_method

l-monninger:
@0xPrimata It's just to match the upstream.

PR #89, opened 2024-05-30
Remove `Arc<RwLock<_>>` around `DbReaderWriter`

mzabaluev:
# Summary

`DbReaderWriter` itself is two (ludicrously redundant) `Arc` pointers, so it's cheaply cloneable and generally, there's no need to guard or share this object.

To be further fixed some day: https://github.com/aptos-labs/aptos-core/issues/13464

# Changelog

- Simplified management of `DbReaderWriter` handles inside executor state.

# Testing

Existing tests pass after the refactoring.


l-monninger (approved):
yep, lgtm

0xmovses:
Yes also agree, can't even remember why this got here.

Issue #63, opened 2024-05-16
Refactor node's public API to facilitate moving state into parallel tasks


mzabaluev:
The top-down design dictated by traits like `SuzukaFullNode` where the "god object" is borrowed immutably as `&self` by async methods does not work well with background tasks that ideally need to work on moved, rather than shared, bits of state. You end up needing `Arc<Mutex<_>>` for anything mutable even if conceptually it's only used in one task.

Consider refactoring to an API that will have two phases:

* Initializing the node through a non-async constructor method or builder API.
* The async method to run the node consuming the state, something like `pub async fn run(self) -> Result<(), NodeError>`.

_Originally posted by @mzabaluev in https://github.com/movementlabsxyz/movement/pull/58#discussion_r1601170509_

## Design discussion

The design guidelines have been refined since the original comment; see #340

## Implementation PRs

- [x] #308
- [ ] #404

musitdev:
I agree with you, and this pattern of using Arc<Mutex<_>> to manage concurrent access can lead to at best contention and at worth deadlock when more and more functionalities are added.
From my experience, what works well is a sort of supervisor loop that start task and manage data sharing using actor model or event driven pattern.
The idea is to dissociate the way you access the data (you access them by moving, cloning, locking depending on the context) from how you process them.

l-monninger:
@mzabaluev 
> An alternative approach is to construct two objects: a handle for invoking public methods on, and a future to drive the background implementation.

The advantages of this are presumably:
1. Signatures for public methods can remain `& self` because receiver-based tasks are being managed elsewhere.
2. Overall separation of concerns. 

The disadvantages are:
1. Readability.
2. API disunity. 

Additionally, any state in the background implementation should really only be accessed through the public methods. So, we are kind of disintegrating something that should be integrated. 

For me, ideally, these concerns are pushed down from the top-level implementation. I certainly don't care if we're doing the above somewhere. But, to the extent we can wrap things up in a nice bow, I still encourage us to do so. 

Ultimately, though, for me the ideal is still to have an API that manages this complexity and looks something like this w.r.t. background tasks: https://github.com/a16z/helios?tab=readme-ov-file#using-helios-as-a-library

@musitdev 
> The idea is to dissociate the way you access the data (you access them by moving, cloning, locking depending on the context) from how you process them.

[...]

PR #91, opened 2024-06-03
ENG-357: A view into the finalized state

mzabaluev:
# Summary
- **RFCs**: [RFC 23: State Finality Modes](https://github.com/movementlabsxyz/rfcs/blob/main/0023-m2-state-finality-modes/rfc-m2-state-finality-modes.md).
- **Requires**: https://github.com/movementlabsxyz/aptos-core/pull/13
- **Blocked by**: #94

Replace the "final executors" with the maptos-fin-view component that shares the storage with the optimistic executor, providing the API instance for the final ledger state.

# Testing

Add crate-level tests for fin-view to verify that updating the finalized block height makes ledger states visible if they are at or below that block height, or not visible otherwise even if they are already added in later versions.

# Outstanding issues

Integration with settlement: when a block is finalized, obtain its ledger info and update the `FinalityView`.


protocol-units/execution/executor/src/v1.rs
```diff
@@ -1,34 +1,49 @@
-// FIXME: glob imports are bad style
-use crate::*;
-use aptos_types::transaction::SignedTransaction;
-use maptos_opt_executor::Executor;
+use crate::{BlockMetadata, ExecutableBlock, Executor, HashValue, SignedTransaction};
```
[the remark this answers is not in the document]
l-monninger: Referring to dir/crate name, btw.

protocol-units/execution/executor/src/lib.rs
```diff
@@ -13,9 +13,10 @@ use aptos_api::runtime::Apis;
 use movement_types::BlockCommitment;
 
 use async_channel::Sender;
+use async_trait::async_trait;
 
-#[tonic::async_trait]
-pub trait SuzukaExecutor {
+#[async_trait]
+pub trait Executor {
```
l-monninger, line 20: Idk if you are trying to place this in this PR or save it for another one, but I don think the design we discussed is probably more readable. 

Either have separate methods for executing each `dyn`, `opt`, and `fin` block type or having one `execute_block` that demuxes an enum of those different blocks, would be ideal IMO.

l-monninger (changes requested):

l-monninger:
@mzabaluev For execution dir, I might go...

```
/execution
    /maptos-dof
        /executor -> maptos-dof-executor
        /opt-executor -> maptos-opt-executor
        /fin-view -> maptos-fin-view
```

`dof` instead of `dyn-opt-fin` if we want something more concise--though IMO we will want to update the RFC to have this be a well-known, proper noun.

protocol-units/execution/executor/src/v1.rs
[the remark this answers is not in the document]
mzabaluev: I was not happy with just "executor", either.
One idea I've had was "maptos-staged-execution". But if I understand your future-proofing idea, we could have differently staged executors.

protocol-units/execution/executor/src/v1.rs
```diff
@@ -1,34 +1,49 @@
...
+			&config.aptos,
+		)?;
+		Ok(Self::new(executor, finality_view, transaction_channel))
 	}
 }
 
-#[tonic::async_trait]
-impl MonzaExecutor for MonzaExecutorV1 {
+#[async_trait]
+impl Executor for ExecutorV1 {
```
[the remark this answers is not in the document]
mzabaluev: There's the `Executor` trait and this implementation type. I suppose the trait is useful for creating mocks, otherwise I'd just have inherent public methods on the type. I don't like naming traits with the `Trait` suffix (looking at you, Aptos), but we'll need to have a couple of names, otherwise having both named `Executor` in one crate may be too confusing.

protocol-units/execution/executor/src/lib.rs
mzabaluev: We will only execute `opt` blocks, though `dyn` might come later. `fin` is updated with `set_finalized_block_height`.

I don't see value in multiplexing these requests in a single method at this point, the way integration will be done in the full node. We could design a channel for sending these events to the executor task that runs in the background, but it's something that should be left for the general rearchitecting called for in #63.

mzabaluev:
Is it OK to name it DOF if we only have OF at this point? 😁

Cargo.toml
```diff
@@ -180,33 +176,33 @@ ethers-middleware = { version = "=2.0.10", default-features = false }
 # Aptos dependencies
 # We use a forked version so that we can override dependency versions. This is required
 # to be avoid depenedency conflicts with other Sovereign Labs crates.
-aptos-vm = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "e9b42128f8ed51e90d06beec72d32797693ab66c" }
-aptos-sdk = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "e9b42128f8ed51e90d06beec72d32797693ab66c" }
-aptos-consensus-types = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "e9b42128f8ed51e90d06beec72d32797693ab66c" }
-aptos-crypto = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "e9b42128f8ed51e90d06beec72d32797693ab66c", features = [
+aptos-vm = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "d1a2d26568a8804b2d73d4a25b5545c8e604a6fc" }
+aptos-sdk = { git = "https://github.com/movementlabsxyz/aptos-core", rev = "d1a2d26568a8804b2d73d4a25b5545c8e604a6fc" }
```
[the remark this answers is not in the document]
mzabaluev: > should we maybe have a field in the PR template for bumping upstream revisions?

Maybe, to summarize the changes we're doing the dependency bump for, preferably with references to PRs in dependencies' repos. I added a **Requires** field in this PR's description for a similar purpose, and it actually covers this dependency bump.

With workspace-wide dependencies, any dependency bumps should be concentrated in the top-level `Cargo.toml`, so the version changes themselves are easy to spot.

protocol-units/execution/executor/src/v1.rs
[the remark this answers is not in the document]
mzabaluev: I ended up with `maptos-dof-execution` in `execution/dof`. The trait is now named `DynOptFinExecutor` and the "production" type is `v1::Executor`.

networks/monza/monza-full-node/Cargo.toml
```diff
@@ -12,7 +12,7 @@ rust-version = { workspace = true }
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-monza-executor = { workspace = true }
+maptos-dof-execution = { workspace = true }
```
l-monninger, line 15: `dof` ftw

protocol-units/execution/fin-view/src/fin_view.rs
```diff
@@ -0,0 +1,200 @@
...
+		let apis = finality_view.get_apis();
+		let context = apis.accounts.context.clone();
+
+		let fin_ledger_info = context.get_latest_ledger_info_wrapped()?;
+		let fin_version = fin_ledger_info.version();
+
+		let state_key = state_key_from_address(account_addrs[1]);
+		let acc_data = context.get_state_value(&state_key, fin_version)?;
+		assert!(acc_data.is_some());
+		let state_key = state_key_from_address(account_addrs[2]);
```
l-monninger, line 194: Yeah, this is good. But, perhaps we should use different var names instead of overwriting, for clarity.

protocol-units/execution/fin-view/src/fin_view.rs
```diff
@@ -0,0 +1,200 @@
...
+			);
+			let block = ExecutableBlock::new(block_id.clone(), executable_transactions);
+			executor.execute_block(block).await?;
+		}
+
+		finality_view.set_finalized_block_height(2)?;
+
+		// Retrieve the executor's API interface and fetch the accounts
+		let apis = finality_view.get_apis();
+		let context = apis.accounts.context.clone();
```
l-monninger, line 186: This is using context, maybe let's do the same thing using `apis.accounts` itself for completeness--in case there's something the API does that we weren't expecting. That or let's link to the Aptos core implementation and leave a comment here that explains why such a test is unccessary.

l-monninger (changes requested):
Generally, lgtm. Lmk what you think about the added test case.

protocol-units/execution/fin-view/src/fin_view.rs
mzabaluev: The APIs in aptos-api expose methods that use path arguments and other scaffolding from the poem web server, so I called context directly to not have to deal with this.
The only exception I saw is `TransactionsApi::get_transaction_by_hash_inner`, perhaps defined as an escape hatch for similar testing purposes, and we use it in executor tests.
But you are right, I'll look into how to make it as close as possible to real life.

protocol-units/execution/fin-view/src/fin_view.rs
mzabaluev: `let` is not overwriting, but rebinding the name for another slot, and the older slot gets dropped. I prefer reusing names for repetitive code; Rust prevents mistakes because slots are immutable by default.

protocol-units/execution/fin-view/src/fin_view.rs
l-monninger: Sweet! As I said, the comment would suffice too.

protocol-units/execution/fin-view/src/fin_view.rs
l-monninger: Yeah, sorry, rebinding. I meant this purely for reusability as we are asserting conditions which are the duals of one another, I was a little confused the first time I read.

l-monninger: EDIT: readability, all the wrong words at this time of day

protocol-units/execution/fin-view/src/fin_view.rs
mzabaluev: aptos-api methods bound for OpenAPI are not public, so I exposed one and used it in https://github.com/movementlabsxyz/movement/pull/91/commits/f25dc248913593378a094acc62ecba74be4245d0

mzabaluev:
Some tests merged from main fail due to port configuration issues. This could be due to the changed environment variables used by `Config::try_from_env`.

6 hours to run CI is a bit much.

l-monninger:
@mzabaluev You're trying to bind the `fin` services to the same port as the `opt` services.

mzabaluev:
Cause of the latest CI error: `MAPTOS_PRIVATE_KEY` is not set when running the faucet in process-compose. I have checked that the faucet binary outputs a randomly generated key.

l-monninger (approved):

mzabaluev:
> What about the "Blocked by?"

That was resolved in the aptos-core PR.

PR #93, opened 2024-06-04
Well Known Files for ENV and Setup and Build Stages 

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Requires**: https://github.com/movementlabsxyz/aptos-core/pull/14

    - Modifies major `protocol-units` to manage setups through Rust binaries `<protocol-unit>-setup`.
    - Modifies major `protocol-units` to read from a well known `config.toml`. 

# Changelog

- Extensive changes to top-level start logic of `suzuka-full-node`.
- Extensive changes to top-level start logic of `monza-full-node`. 
- Extensive changes to top-level start logic of `m1-da-light-node`.
- Added `commander` crate for ergonomic process spawning. 
- Added setup scripts for  `<protocol-unit>-setup`. The most extensive is `m1-da-light-node-setup`. 

# Testing

- Unit tests and e2e tests are passing.
- `process-compose` e2e tests have been modified to account for new setup. 

# Outstanding issues
1. This is overall fairly scrappy. While Configs are reasonable well organized internally, the level of abstraction at which configs suited for reading from `$DOT_MOVEMENT_PATH/config.toml` varies. 
2. Overall, I suggest we adopt (as this PR currently and implicitly does) a standard of config parameters similar to that used with protobuf message types. That is fields, or at least wrapping structs should be globally unique. This allows for flattening at higher-order configs and thus for configs to continue to be read from a single `config.toml`. `m1-da-light-node` parameters can be improved in this regard. 
3. $DOT_MOVEMENT is used for all configs and state. Ideally, this can be changed to public and private configs and state. 
4. `docker-compose` orchestration with this setup in mind is to be tackled in a forthcoming PR.


l-monninger:
gali

l-monninger:
**Gali PR:** https://github.com/movementlabsxyz/infra/pull/45

l-monninger:
Once CI-passes this should be good to resolve merge.

l-monninger:
Oh, this also needs an update to our `README.md` to account for the new runner commans.

l-monninger:
Going to quiet the logging again.

.github/workflows/build-push-containers.yml
```diff
@@ -0,0 +1,331 @@
+name: Build Push Containers
+
+on:
+  push:
```
mzabaluev, line 4: Do we really want to do this on every push? This could be down to each throwaway commit pushed into a development branch.

.github/workflows/checks.yml
```diff
@@ -24,25 +24,101 @@ jobs:
...
       uses: DeterminateSystems/nix-installer-action@main
 
-    - name: Run MCR tests in nix environment
-      run: nix develop --command bash  -c "just mcr-contract-tests"
-
-    - name: Run MCR client tests in nix environment
-      run: CELESTIA_LOG_LEVEL=FATAL MOVEMENT_PREBUILT=false nix develop --command bash  -c "just mcr test.local -t=false"
-
     - name: Run Cargo Check in nix environment
       run: nix develop --command bash  -c "cargo check"
```
mzabaluev, line 28: `cargo check` builds in incremental mode by default.

This means all crates, including depenencies, will need to be built again by the test jobs.

It may be more efficient to do `cargo build --release` if that's what the test jobs are building with.
The full devil is in the details: some cargo flag thrown in by the test setup scripts may mean all crates will need to be rebuilt for the tests anyway. But disabling incremental builds is a worthwhile step to save some build time.

Cargo.toml
```diff
@@ -8,19 +8,15 @@ members = [
...
     "util/buildtime/buildtime-helpers",
     "util/buildtime/buildtime-macros",
-    "networks/monza/monza-client",
-    "networks/monza/monza-config",
-    "networks/monza/monza-full-node",
-    "networks/suzuka/suzuka-client",
-    "networks/suzuka/suzuka-config",
-    "networks/suzuka/suzuka-full-node",
+    "networks/monza/*",
+    "networks/suzuka/*",
```
mzabaluev, line 19: OK, as long as all relevant crates are listed in the workspace dependencies anyway.

.github/workflows/build-push-containers.yml
mzabaluev: Perhaps it should be filtered to the branch `main`, after you confirm that the CI workflow works in this branch.

No need to automatically push junky development images to Docker Hub.

networks/monza/faucet/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "monza-faucet-service"
+description = "Monza Faucet Service"
```
mzabaluev, line 3: This goes beyond switching to well-known files, but OK.

networks/monza/monza-config/src/lib.rs
```diff
@@ -1,20 +1,36 @@
-#[derive(Debug, Clone, PartialEq, Eq)]
+use serde::{Deserialize, Serialize};
+use std::path::PathBuf;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct Config {
+	#[serde(flatten)]
```
mzabaluev, line 6: This is not future-proof: if something needs to be added besides the execution units' config, the file format will need to be broken.

networks/monza/monza-config/src/lib.rs
```diff
@@ -1,20 +1,36 @@
...
+	#[serde(default = "Config::default_execution_config")]
 	pub execution_config: maptos_execution_util::config::Config,
```
mzabaluev, line 8: Could be renamed without information loss, to get more concise naming in the TOML schema in case we remove the `flatten` flag above.

```suggestion
	pub execution: maptos_execution_util::config::Config,
```

networks/monza/monza-full-node/src/main.rs
```diff
@@ -14,7 +14,10 @@ async fn main() -> Result<(), anyhow::Error> {
 			.init();
 	}
 
-	let executor = MonzaPartialNode::try_from_env()
+	let dot_movement = dot_movement::DotMovement::try_from_env()?;
```
mzabaluev, line 17: 😍

networks/monza/monza-config/src/lib.rs
l-monninger: > Overall, I suggest we adopt (as this PR currently and implicitly does) a standard of config parameters similar to that used with protobuf message types. That is fields, or at least wrapping structs should be globally unique. This allows for flattening at higher-order configs and thus for configs to continue to be read from a single config.toml. m1-da-light-node parameters can be improved in this regard.

Does this not cover that? Essentially, we have to assume that every field that needs to be read from a file does not conflict in the global namespace. So, it's flatten at every level where reading from `config.toml` is valid.

.github/workflows/build-push-containers.yml
l-monninger: These go to ghcr.io and are important for testing. We could pattern match the commit content and use an `if: matched` expression. This would show when container checks are skipped s.t. container correctness can still be verified.

Cargo.toml
l-monninger: Yeah, for now, that's what I'd like to strive for up at `networks/`.

networks/monza/faucet/Cargo.toml
l-monninger: Well... it kind of came as a necessity because there wasn't an easy way to make the old way of running the faucet respect the ENV variables.

networks/monza/monza-config/src/lib.rs
l-monninger: There are two patterns which I thought it might be nice to achieve in future commits.

1. **Always wrap `Config` fields in a globally uniquely named struct**: This improves readability of the TOML and is more like protobuf service standard--which this essentially needs to mimic. (We run into the same kind of problems that lead to the decisions made in protobuf standard.)
2. **`Config` and `Parameters`**: each unit knows exactly what it needs to run--perhaps sometimes in different modes. Each `Config` which has optional fields should `.try_into()` into a target `Parameters`. This not only cleans up unit implementations and separate concerns. But, it also means that we can know if the unit will fail slightly earlier in the process. We don't have to wait until it's trying to bind to an address for example.

networks/monza/monza-config/src/lib.rs
mzabaluev: Understood. As long as we manage the fields to be unique, this is workable.

mzabaluev (approved):

networks/monza/monza-config/src/lib.rs
mzabaluev: On point 2:

If I understand the idea correctly, `Config` provides the schema for human-readable configuration (via serde) without enforcing constraints like mandatory fields? And then `impl TryFrom<Config> for Parameters` would validate the parameters, provide domain-specific error diagnostics that would be difficult to shoehorn into serde, etc.

I think the separation of concerns is nice, but I'd like to morph `Parameters` into a builder API for protocol unit instances. We need the programmatic way of constructing the objects that's better than monstrous `new` methods with 6+ arguments.

PR #96, opened 2024-06-05
Services Containers nixed

l-monninger:
# Summary
- **Categories**: `cicd`

Containers for all services using a combination of Nix and Docker.  

ENG-480, ENG-482 & ENG-483

# Changelog
- Docker files for each service grouped in folders
- Updates to `flake.nix` creating build shells to workaround issue with vendor dependencies building

# Testing
Manual test of generating each image would be required, some do take long enough for a short nap.

# Outstanding issues
As mentioned, there is a workaround here, which works and actually produces some small images which will need to be fixed at some point.


l-monninger:
Just comment out or `if: false` the container workflow. It is being worked on here: https://github.com/movementlabsxyz/movement/pull/93.

andyjsbell (approved):

PR #97, opened 2024-06-05
create maptos-rest service

0xmovses:
# Summary
We need additional endpoints to able to query `state_root_hash` from an e2e test. 
Some formatting was applied automticallty, only relevant changes are inside `maptos-rest`. 

# Changelog
Added new package `maptos-rest`
Added simple units tests. 
Added endpoints

# Testing
`cargo test --package movement-rest `

# Outstanding issues


protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+
+#[handler]
+async fn state_root_hash() -> Response {
+	"the_state_root_hash".into_response()
+}
+
+pub fn create_routes() -> Route {
+	Route::new()
+		.at("/health", poem::get(health))
+		.at("/state_root_hash", poem::get(state_root_hash))
```
l-monninger, line 31: Use hyphens: https://restfulapi.net/resource-naming/

protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+	}
+}
+
+#[handler]
+async fn health() -> Response {
+	"healthy".into_response()
+}
+
+#[handler]
+async fn state_root_hash() -> Response {
```
l-monninger, line 24: You will want `Context` here. Are you trying to merge this PR before the actual implementation is done for organizational purposes?

l-monninger (changes requested):

protocol-units/maptos-rest/src/lib.rs
0xmovses: I was yeah, just to keep PRs small. But I'l just keep comitting here, till the whole e2e test is here.

networks/monza/monza-client/src/tests/mod.rs
```diff
@@ -134,5 +136,12 @@ async fn test_example_interaction() -> Result<()> {
 			.context("Failed to get Bob's account balance the second time")?
 	);
 
+<<<<<<< HEAD
```
andyjsbell, line 139: a missed conflict maybe

networks/suzuka/suzuka-full-node/src/lib.rs
```diff
@@ -13,6 +13,9 @@ pub trait SuzukaFullNode {
 	/// Runs the executor until crash or shutdown.
 	async fn run_executor(&self) -> Result<(), anyhow::Error>;
 
+	/// Runs the maptos rest apit service until crash or shutdown.
```
andyjsbell, line 16:
```suggestion
	/// Runs the maptos rest api service until crash or shutdown.
```

protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+	// More fields to be added here, log verboisty, etc.
+}
+
+impl MaptosRest {
+	pub const MAPTOS_REST_ENV_VAR: &'static str = "MAPTOS_REST_URL";
+
+	pub fn try_from_env(context: Arc<Context>) -> Result<Self, Error> {
+		let url = env::var(Self::MAPTOS_REST_ENV_VAR)
+			.unwrap_or_else(|_| "http://localhost:30832".to_string());
+		Ok(Self { url, context })
```
andyjsbell, line 27: maybe better as constant

protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+	)?;
+	let state_root_hash = txn_with_proof
+		.proof
+		.transaction_info
+		.state_checkpoint_hash()
+		.ok_or_else(|| anyhow::anyhow!("No state root hash found"))?;
+	Ok(state_root_hash.to_string().into_response())
+}
+
+//These feel more like integration tests, can be moved to a different
```
andyjsbell, line 71: Do we want to leave this here?

andyjsbell (commented):
looks good, a few nits

mzabaluev:
This needs to be rebased against or merged from recent main.

Cargo.toml
```diff
@@ -60,6 +61,7 @@ mcr-settlement-client = { path = "protocol-units/settlement/mcr/client" }
 mcr-settlement-manager = { path = "protocol-units/settlement/mcr/manager" }
 ## types
 movement-types = { path = "util/movement-types" }
+maptos-rest = { path = "protocol-units/maptos-rest" }
```
mzabaluev, line 64: Maybe add a separate "section" for this with comments, and sort it somewhere between mempool and sequencing.

Cargo.toml
```diff
@@ -12,6 +12,7 @@ members = [
     "protocol-units/mempool/*",
     "protocol-units/settlement/mcr/client",
     "protocol-units/settlement/mcr/manager",
+    "protocol-units/maptos-rest",
```
mzabaluev, line 15: Elsewhere, "maptos" is used in crate names because they are supposed to be global, but in the directory tree we don't need the prefix. Rename the dir to something like "api/rest"?

log_file.txt
```diff
@@ -0,0 +1,16 @@
+[2m2024-06-11T12:44:12.190727Z[0m [32m INFO[0m [2mbasicalicebob[0m[2m:[0m Scenario:1
```
mzabaluev, line 1: Spuriously added file, please remove.

networks/suzuka/suzuka-client/Cargo.toml
```diff
@@ -12,13 +12,24 @@ rust-version = { workspace = true }
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+async-channel = { workspace = true }
```
mzabaluev, line 15: What's `async-channel` for?

protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+	middleware::{AddData, Tracing},
+	web::{Data, Path},
+	EndpointExt, IntoResponse, Response, Route, Server,
+};
+use std::env;
+use std::sync::Arc;
+use tracing::info;
+
+#[derive(Debug)]
+pub struct MaptosRest {
```
mzabaluev, line 15: Here also, let's not promote "maptos" into an official naming prefix for actual item names. This quirkiness is only needed to disambiguate in flat namespaces like crate names (when published in a registry) and environment variables. Just use `Server` so it's imported as `use maptos_rest::Server`?

protocol-units/maptos-rest/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+		let maptos_rest = self.create_routes();
+		Server::new(TcpListener::bind(&self.url)).run(maptos_rest).await?;
+		Ok(())
+	}
+
+	pub fn create_routes(&self) -> impl EndpointExt {
+		Route::new()
+			.at("/health", get(health))
+			.at("/movement/v1/state-root-hash/:blockheight", get(state_root_hash))
+			.at("movement/v1/richard", get(richard))
```
mzabaluev, line 42: Does this endpoint have an OpenAPI description and an RFC? 🤔

test_result.txt
```diff
@@ -0,0 +1,11 @@
+{"timestamp":"2024-06-11T12:44:07.084573Z","level":"INFO","fields":{"msg":"1 Before alice fund"},"target":"exec"}
```
mzabaluev, line 1: Another spuriously added file.

util/movement-types/src/lib.rs
```diff
@@ -168,6 +168,15 @@ impl From<Commitment> for Vec<u8> {
 	}
 }
 
+impl fmt::Display for Commitment {
```
mzabaluev, line 171: This creates a canonical string representation for commitment values.
I think it's OK, but if it's only meant to be used in logging and debugging, it may be better to go with `Debug`.

mzabaluev (changes requested):
Lots of files were spuriously added under `protocol-units/bridge`, please remove them.

protocol-units/maptos-rest/src/lib.rs
mzabaluev: Also, perhaps `movement-rest`?

0xmovses: The name was suggested by Liam when I was asking about how to name the service. But I agree the quirkiness is not needed here!

test_result.txt
0xmovses: Will get rid of these

networks/suzuka/suzuka-client/src/tests/mod.rs
```diff
@@ -130,5 +131,29 @@ async fn test_example_interaction() -> Result<()> {
 			.context("Failed to get Bob's account balance the second time")?
 	);
 
+	sleep(Duration::from_secs(10)).await;
+
+	let anvil_rpc_port = "8545";
+	let anvil_rpc_url = format!("http://localhost:{anvil_rpc_port}");
+	let anvil_ws_url = format!("ws://localhost:{anvil_rpc_port}");
```
mzabaluev, line 138: I'm not sure if I'm comfortable with adding "unit" tests that require additional infra to run. I'd like `cargo test` in the root of the checked out workspace to not require any external setup. But this may be too late to complain about this if we have similar requirements for other crates' tests already.

I think such e2e-ish tests should be in their own crate, or perhaps they should be feature-gated to not be  run by `cargo test` by default.

util/movement-types/src/lib.rs
0xmovses: `Debug` is already implemented for `Commitment`, I needed to implement `Display` on it for some other reason, what that was now I can't quite recall!

networks/suzuka/suzuka-client/src/tests/mod.rs
0xmovses: Looks like this will be addressed in its own PR. But yes, agree.

networks/suzuka/suzuka-client/Cargo.toml
0xmovses: And Old thing I was doing, left over. Removed now :)

util/movement-types/src/lib.rs
mzabaluev: `Debug` is just a derive though, it prints decimal byte values. Maybe we should implement it to print hex too.

Cargo.toml
```diff
@@ -50,6 +50,8 @@ maptos-dof-execution = { path = "protocol-units/execution/dof" }
 maptos-opt-executor = { path = "protocol-units/execution/opt-executor" }
 maptos-fin-view = { path = "protocol-units/execution/fin-view" }
 maptos-execution-util = { path = "protocol-units/execution/util" }
+# infra
```
mzabaluev, line 53:
```suggestion
## infra
```
Otherwise it looks like a "first-level header" under which "mempool", "sequencing", etc. are grouped.

mzabaluev (commented):
This needs updating with the new config stuff from main, otherwise looks good.

Cargo.toml
```diff
@@ -140,28 +140,48 @@ secp256k1 = { version = "0.27", default-features = false, features = [
...
+#     "ws",
+# ] }
+# alloy-sol-types = { version = "0.7.2", features = ["json"] }
+# alloy_network = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-network" }
+# alloy_transport = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-transport" }
+# alloy_transport_http = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-transport-http" }
+# alloy_contract = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-contract" }
+# alloy_signer_wallet = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-signer-wallet" }
+# alloy_rpc_types = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-rpc-types" }
+# alloy_transport_ws = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-transport-ws" }
```
mzabaluev, line 179: This should better be removed.

Cargo.toml
```diff
@@ -140,28 +140,48 @@ secp256k1 = { version = "0.27", default-features = false, features = [
...
-alloy = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy", features = [
+alloy = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy", rev = "607359a8d7ba7543d702da5d406b7bc78eff3d9f", features = [
     "pubsub",
     "signers",
     "signer-wallet",
     "signer-yubihsm",
 ] }
 alloy-primitives = { version = "0.7.2", default-features = false }
-alloy_provider = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-provider", features = [
+alloy_provider = { git = "https://github.com/alloy-rs/alloy.git", package = "alloy-provider", rev = "607359a8d7ba7543d702da5d406b7bc78eff3d9f", features = [
```
mzabaluev, line 150: Since we keep `Cargo.lock` under version control, the purpose of this is to protect against inadvertent updates, correct?

mzabaluev (commented):
Looks good, except for some trivial cleanup.

Cargo.toml
0xmovses: oh whoops, intended to.

Cargo.toml
0xmovses: Yes, I ran into that an hour back. We got the latest upstream which broke our code.

protocol-units/movement-rest/src/lib.rs
```diff
@@ -0,0 +1,96 @@
...
+
+#[derive(Debug)]
+pub struct MovementRest {
+	/// The URL to bind the REST service to.
+	pub url: String,
+	pub context: Option<Arc<Context>>,
+	// More fields to be added here, log verboisty, etc.
+}
+
+impl MovementRest {
```
l-monninger, line 22: This will obviously have to change at some point soonish to use the `config.toml`.

l-monninger (approved):

PR #98, opened 2024-06-06
stlkMOVEToken

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Introduces the `stlkMOVE` a stakable locked MOVE token. 

# Changelog

- Thorough changes to custodial tokens in `protocol-units/settlement/mcr/contracts`. 

# Testing

- Extensive contract tests. No integration with rest of protocol in this PR. 

# Outstanding issues
- Integration work is being done in #104.


l-monninger:
@0xPrimata Right now, the initialization only works for non-multi-inheritance paths. I'm focusing on just getting one inheritance chain up. So, if you have time to actually set up for [`init` and `init_unchained`](https://docs.openzeppelin.com/contracts/4.x/upgradeable#further_notes) that would be great.

l-monninger:
Additionally, @0xPrimata 
- Add `GovernorTimelockControl`.
- I believe the storage for `LockedToken` should be refactored to use a more conventional storage contract pattern. 
- Suggest voting-based governance.

l-monninger:
@0xPrimata if we want to close for cleanup, I think this is good for review.

andyjsbell (approved):

PR #102, opened 2024-06-07
Suzuka Doc Comments

BenDavis36:
# Summary
Creation of doc comments for Suzuka

# Changelog
Adds comments for `struct SuzukaPartialNode<T>` and associated fields.
Adds comments for `pub struct Config`.

### DIFF
--- a/networks/suzuka/suzuka-full-node/src/partial.rs
+++ b/networks/suzuka/suzuka-full-node/src/partial.rs
@@ -24,11 +24,36 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
+/// Suzuka partial node has the ability to
 pub struct SuzukaPartialNode<T> {
+	/// The executor responsible for executing transactions and blocks.
+	///
+	/// 'T' must implement the `SuzukaExecutor` trait, allowing different
+	/// types of executors to be used with the `SuzukaPartialNode`.
 	executor: T,
+
+	/// Asynchronous channel sender for signed transactions.
+	///
+	/// This sender is used to send transactions to the transaction receiver for
+	/// processing.
 	transaction_sender: Sender<SignedTransaction>,
+
+	/// Asynchronous channel receiver for receiving signed transactions.
+	///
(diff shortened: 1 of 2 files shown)


0xmovses (changes requested):

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -24,11 +24,36 @@ use std::future::Future;
 use std::sync::Arc;
 use std::time::Duration;
 
+/// Suzuka partial node has the ability to
```
mzabaluev, line 27: This needs to be completed.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -24,11 +24,36 @@ use std::future::Future;
...
 pub struct SuzukaPartialNode<T> {
+	/// The executor responsible for executing transactions and blocks.
+	///
+	/// 'T' must implement the `SuzukaExecutor` trait, allowing different
+	/// types of executors to be used with the `SuzukaPartialNode`.
```
mzabaluev, line 32: Private members do not really need doc comments, as they won't be rendered in the API documents. A plain one-line comment describing the purpose should be enough.

0xmovses: There are tonnes of private members in aptos core that are doc commented. This is for IDE support more than rust crate docs I suppose, for eg.

```rust
/// See [AptosDB::get_transaction_by_version].
        ///
        /// [AptosDB::get_transaction_by_version]: ../aptosdb/struct.AptosDB.html#method.get_transaction_by_version
        fn get_transaction_by_version(
            &self,
            version: Version,
            ledger_version: Version,
            fetch_events: bool,
        ) -> Result<TransactionWithProof>;
        
        ```

mzabaluev: I just checked and you are right, this gives better quality of life in the IDE.

PR #103, opened 2024-06-08
Reduce faucet image size

andyjsbell:
# Summary
- **Categories**: `cicd`

Image size for `aptos-faucet-service` was > 1GB which is now 146MB

# Changelog

- Used ldd to build dependency tree

# Testing

# Outstanding issues


l-monninger (approved):
lgtm

PR #105, opened 2024-06-10
ENG 539 e2e test node settlement

musitdev:
# Summary

Your initial scenario will look more like:

Run something like Alice and Bob.

Fetch the state root from OPT.

Fetch the last accepted commitment from the settlement contract.

Check that the state root and block height match.

# Changelog

Some of the changes came from the PR#139 as it hasn't been merged in main and this PR use it.
The PR's changes are:

The e2e test is in `networks/suzuka/suzuka-client/src/tests/settlement_state.rs` file.

The MCR setup has been modified.
 - `protocol-units/settlement/mcr/setup/src/local.rs` contains the start Anvil part and config definition.
 - `protocol-units/settlement/mcr/setup/bin/one_validator_genesis_ceremonial.rs` do the genesis ceremony.
 
The MCR unit test has been updated to call the mcr setup to start Anvil and use `protocol-units/settlement/mcr/setup/bin/two_validator_genesis_ceremonial.rs` to do the test genesis ceremonial.

Process-compose for Suzuka node and mcr has been updated to call setup and the tests.

The [.github/workflows/checks.yml](https://github.com/movementlabsxyz/movement/pull/105/files#diff-3ea54af4839eb75404d71b28252bead7e7ec8f676b1f815e1cde02629a75c165) has been updated to start both test in the CI.

[.gitmodules](https://github.com/movementlabsxyz/movement/pull/105/files#diff-fe7afb5c9c916e521401d3fcfb4277d5071798c3baf83baf11d6071742823584) has been updated to add Solidity libs in MCR contracts.

The struct `MovementRest` has been removed (not needed) because only the Movement Rest services are used and integrated in Suzuka node opt and fin entry points.

I've created 2 services `get_current_commitment` and `get_finalized_block_info` to get commitment states.

Some unused files in MCR client has been removed.

I've augmented the gas increment in case of insufficient gas because there's a big diff between calculated gas and used ones. So to avoid retrying too much time, I've increased the increment (+25% more gas after each retry).

Remove now unused script that were starting Anvil and loading contract. 

I've updated the commander to spawn a task in the background.

# Testing

To run the test, Anvil must be installed. It's started during setup.

To run the e2e test, I use this command:
```
CELESTIA_LOG_LEVEL=FATAL MOVEMENT_PREBUILT=false nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just suzuka-full-node native e2etest.build.setup.local --keep-tui"
```
After each execution, I remove the `.movement` directory to remove the last execution generated config.

Anvil process should be killed at the end of execution of Suzuka node, but it's not always the case, so before starting the test verify that Anvil is not running and kill it if yes.

# Outstanding issues

For the smart contract to be started, after genesis ceremonial, the first commitment (height 1) must be submitted by hand because Suzuka node first commitment height start at 2.
If this first commitment is not submitted, the Suzuka node commitment is rejected silently.

The MCR test leave Anvil running after execution because when Tokio end, struct are not dropped. It can be solved but need to manage Tokio end execution. As the config and anvil start management will be changed, I leave it for now.

If an error occurs during Eth Tx send, the Suzuka node end because the error is not managed and stop the node. Eth Tx error are very common, a Tx retry logic after error is implemented in `eth_client.rs` and can be updated to manage more errors, but perhaps the node should have its own error logic when writing to the chain.


0xmovses:
Note to reviewers this is still in WIP. @musitdev will be rebasing in this work https://github.com/movementlabsxyz/movement/pull/97 for querying the `state_root_hash` at the rest api level.

networks/suzuka/suzuka-full-node/src/main.rs
```diff
@@ -14,13 +14,21 @@ async fn main() -> Result<(), anyhow::Error> {
 			.init();
 	}
 
+	// Load variables defined in .env file.
+	let movement_storage_path =
+		std::env::var("MOVEMENT_BASE_STORAGE_PATH").unwrap_or("".to_string());
+	let mut env_file_path = std::env::current_dir()?;
+	env_file_path.push(movement_storage_path);
+	env_file_path.push(".env".to_string());
+	dotenv::from_filename(env_file_path)?;
```
mzabaluev, line 23: I think we should not depend on `dotenv` in production code. Setting environment variables in the process is [generally unsafe](https://doc.rust-lang.org/std/env/fn.set_var.html#safety), and even if it's safe in this particular case, it's a deeply weird way to configure the node.

Can we just source the .env file in the suzuka-full-node shell script? Not that it should be even necessary if you can just get the prelude script to set the environment variables for process-compose.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -232,23 +239,71 @@ where
...
+		Self::bound(executor, light_node_client, settlement_client, maptos_rest)
 	}
 }
+
+pub const ETH_RPC_URL_ENV_VAR: &'static str = "ETH_RPC_URL";
+pub const ETH_WS_URL_ENV_VAR: &'static str = "ETH_WS_URL";
+pub const ETH_MCR_CONTRACT_ADDRESS_VAR: &'static str = "ETH_MCR_CONTRACT_ADDRESS";
+pub const ETH_DEFAULT_TX_GAS_LIMIT_VAR: &'static str = "ETH_DEFAULT_TX_GAS_LIMIT";
+pub const ETH_MAX_TX_SEND_RETRY_VAR: &'static str = "ETH_MAX_TX_SEND_RETRY";
+pub const ETH_VALIDATOR_PRIVATE_ADDRESS_VAR: &'static str = "ETH_VALIDATOR_PRIVATE_ADDRESS";
```
mzabaluev, line 277: To avoid repetitive and inconsistent `_ENV_VAR` and `_VAR` suffixes in Rust names, it might be better to group these constants under an `env_vars` module, like done in #116.

In general, Rust's module system and name aliases help avoid the C-like "every unit's name must contain its entire provenance and purpose" disease.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -232,23 +239,71 @@ where
...
+	let config = McrEthSettlementConfig { mrc_contract_address, gas_limit, tx_send_nb_retry };
+
+	let private_address = std::env::var(ETH_VALIDATOR_PRIVATE_ADDRESS_VAR)?;
+	let client =
+		McrEthSettlementClient::build_with_urls(&rpc_url, ws_url, &private_address, config).await?;
+	Ok(client)
+}
+
+#[cfg(test)]
+use mcr_settlement_client::mock::MockMcrSettlementClient;
```
mzabaluev, line 305: This use item could be in the function body below, no need for a separate `#[cfg(test)]` guard then.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -232,23 +239,71 @@ where
...
+	let rpc_url = std::env::var(ETH_RPC_URL_ENV_VAR).unwrap_or("http://localhost:8545".to_string());
+	let ws_url = std::env::var(ETH_WS_URL_ENV_VAR).unwrap_or("ws://localhost:8545".to_string());
+	let gas_limit = std::env::var(ETH_DEFAULT_TX_GAS_LIMIT_VAR)
+		.unwrap_or(DEFAULT_TX_GAS_LIMIT.to_string())
+		.parse()?;
+	let tx_send_nb_retry = std::env::var(ETH_MAX_TX_SEND_RETRY_VAR)
+		.unwrap_or(MAX_TX_SEND_RETRY.to_string())
+		.parse()?;
+	let mrc_contract_address =
+		std::env::var(ETH_MCR_CONTRACT_ADDRESS_VAR).unwrap_or(MCR_CONTRACT_ADDRESS.to_string());
```
mzabaluev, line 295: We have `McrEthSettlementConfig::try_from_env` reading from a differenlty defined set of environment variables, should we avoid duplication here?

This also invites a question as to why the URL and the private address parameters need to be outside `McrEthSettlementConfig` (another name that does not need to be this long in Rust), but I can see a case for exposing more polymorphism on the mcr-settlement-client level.

process-compose/suzuka-full-node/process-compose.yml
```diff
@@ -32,7 +32,7 @@ processes:
 
   suzuka-full-node:
     command: |
-      ./target/$CARGO_PROFILE/suzuka-full-node
+      RUST_LOG=debug ./target/$CARGO_PROFILE/suzuka-full-node
```
mzabaluev, line 35: Nitpick: it may be stylistically better to use the `env` configuration object in YAML.

process-compose/suzuka-full-node/process-compose.yml
```diff
@@ -32,7 +32,7 @@ processes:
...
     depends_on:
```
mzabaluev, line 36: The node should probably depend on `deploy-mcr` to make sure that the MCR configuration has been written.
But `deploy-mcr` is not defined in this file, only in the `anvil` overlay. 🤔

networks/monza/monza-client/src/tests/mod.rs
```diff
@@ -4,8 +4,10 @@ use crate::{
 	types::LocalAccount,
 };
 use anyhow::{Context, Result};
+use async_channel::Sender;
 use once_cell::sync::Lazy;
 use std::str::FromStr;
+use suzuka_executor::v1::SuzukaExecutorV1;
```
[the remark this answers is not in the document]
mzabaluev: This was fixed in #97, I believe.

networks/suzuka/suzuka-client/Cargo.toml
```diff
@@ -12,13 +12,32 @@ rust-version = { workspace = true }
...
+async-channel = { workspace = true }
+alloy-sol-types = { workspace = true  }
+alloy_provider = { workspace = true }
+alloy_signer_wallet = { workspace = true }
+alloy_network = { workspace = true }
+alloy-primitives = { workspace = true  }
+alloy_contract = { workspace = true }
+alloy_transport = { workspace = true }
+alloy_rpc_types = { workspace = true }
+
```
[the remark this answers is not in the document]
mzabaluev: Upon a better look, few of the dependencies are not used by production code; I've addressed this in #126.

networks/suzuka/suzuka-client/src/tests/settlement_state.rs
```diff
@@ -0,0 +1,290 @@
...
+use std::env;
+use std::str::FromStr;
+use url::Url;
+
+sol!(
+	#[allow(missing_docs)]
+	#[sol(rpc)]
+	MCR,
+	"../../../protocol-units/settlement/mcr/client/abi/MCR.json"
+);
```
mzabaluev, line 24: This generates the RPC client from `MCR.json` (BTW, this relative path will break if the crate is ever vendored outside the workspace).
In `eth_client.rs`, the ABI is compiled from `MCRLegacy.json`. I wonder if we can run into a discrepancy?

musitdev: Both `eth_client.rs` and [settlement_state.rs](https://github.com/movementlabsxyz/movement/pull/105/files/71b53cf387f2357fb818e9ba7885486208b321c6#diff-415c7420ab8b4c2875c6cbe7cc79806fef15cb260512f70bac492e61ad4d9ec1) compile using the same file `MCR.json` . Perhaps I'm missing something but the intention was to use the same ABI file for all clients.

mzabaluev: Right, it got moved to `MCRLegacy.json` in main, and there's a bunch of new eth client code that will use the changed contract `MCR.json`, but it's not in use yet. I have updated the file path in my merge branch.

mzabaluev:
Superseded by #460.

PR #107, opened 2024-06-10
CI cleanup for containers

mzabaluev:
# Summary
- **Categories**: `scripts`, `cicd`.

Various improvements in the CI workflow and scripts for building containers.

# Testing

CI should pass. The scripts that have failing commands inside them should fail, or not be invoked
(the container workflows are currently disabled?)

### DIFF
--- a/scripts/movement/build-push-image
+++ b/scripts/movement/build-push-image
@@ -1,3 +1,6 @@
 #!/bin/bash
+set -e
+
 FILE_PATH=./docker/build/$1/Dockerfile
+
 ./scripts/docker/build-push-image $FILE_PATH mvlbs $1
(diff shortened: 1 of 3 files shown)


scripts/movement/build-push-image
```diff
@@ -1,3 +1,6 @@
 #!/bin/bash
+set -e
+
 FILE_PATH=./docker/build/$1/Dockerfile
-./scripts/docker/build-push-image $FILE_PATH mvlbs $1
\ No newline at end of file
+
+./scripts/docker/build-push-image $FILE_PATH mvlbs $1
```
mzabaluev, line 6: I'm afraid we don't have a `movement/Dockerfile` in master any more. Should we remove this workflow or let it be disabled to put something else later instead?

Is there currently a value in building containers for every git push?

PR #112, opened 2024-06-11
Timelock Contracts Deployment Framework

0xPrimata:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `scripts`

Initializing a proper deployment flow for any Timelocked upgradeability among the Movement settlement stack.

Sample uses MCR but it's applicable to any other contract.

# Testing

Script runs against upgradeable MCR. Need to implement for the rest.
A generic implementation is likely not possible.

# Outstanding issues
Interfacing with all contracts. If there is any modification on their initialization, deployment flow needs to be updated.


0xmovses:
@0xPrimata can you list the commands I need to run the tests locally? CI is failing btw

PR #114, opened 2024-06-11
ENG-591 ETH side atomic bridge contract

0xmovses:
# Summary
- **RFCs**: [Link to RFC](https://github.com/movementlabsxyz/rfcs/blob/code8/ENG-338/atomic-swap-bridge/0040-atomic-bridge/rfc-0040-atomic-bridge.md), 
- **Categories**: 
`protocol-units`

Creates the solidity contract for the Initiator of the Atomic Bridge Swap.

# Changelog
- Creates `AtomicBridgeInitiator.sol`
- And `AtomicBridgeInitiator.t.sol`
- Creates a new directory `protocol-units/bridge/movement-birdge/`
- Intializes a foundry project 
- Adds the `IWETH19.sol` 

# Testing
`cd protocol-units/bridge/movement-bridge/ && forge test --fork-url https://ethereum-sepolia-rpc.publicnode.com -vv


protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+pragma solidity ^0.7.6;
+
+contract AtomicBridgeInitiator is IAtomicBridgeInitiator {
+    struct BridgeTransfer {
+        uint256 amount;
+        address originator;
+        address recipient;
+        bytes32 hashLock;
+        uint timeLock;
+        bool exists;
```
andyjsbell, line 14: what is `exists` used for?  We set it to true when we create a transfer which is then stored as a keyed value, hence it exists

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+            abi.encodePacked(
+                _originator, 
+                _recipient, 
+                _hashLock, 
+                _timeLock, 
+                block.timestamp
+        ));
+
+        // Check if the bridge transfer already exists
+        require(bridgeTransfers[_bridgeTransferId].amount == 0, "Bridge transfer already exists");
```
andyjsbell, line 56: Could bring this to the top to save gas on a failing transaction

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        // If msg.value is greater than 0, convert ETH to WETH and add to total amount
+        if (msg.value > 0) {
+            weth.deposit{value: msg.value}();
+            //Transfer WETH to this contract, revert if transfer fails
+            require(weth.transfer(address(this), msg.value), "WETH transfer failed");
+            totalAmount += msg.value;
+        }
+
+        // Ensure there is a valid total amount
+        require(totalAmount > 0, "Total amount must be greater than 0");
```
andyjsbell, line 44: This at the top maybe

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+            exists: true
+        });
+
+        emit BridgeTransferInitiated(_bridgeTransferId, _originator, _recipient, _hashLock, _timeLock);
+        return _bridgeTransferId;
+    }
+
+    function completeBridgeTransfer(bytes32 _bridgeTransferId, bytes32 _secret) external override {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[_bridgeTransferId];
+        require(bridgeTransfer.exists, "Bridge transfer does not exist");
```
andyjsbell, line 73: couldn't we use another member of the struct and remove `exists`?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: Yes I had removed this field for the same reason. It was in the @cde8's original design. But it does allow for this nice error message on L73. Without it user would get a generic revert error. But I'm happy to remove it if we want. 
`require(bridgeTransfer.exists, "Bridge transfer does not exist");`

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(keccak256(abi.encodePacked(_secret)) == bridgeTransfer.hashLock, "Invalid secret");
+
+        // Move the bridge transfer to completed
+        completedBridgeTransfers[_bridgeTransferId] = bridgeTransfer;
+
+        // Delete from active bridgeTransfers
+        delete bridgeTransfers[_bridgeTransferId];
+
+        payable(bridgeTransfer.recipient).transfer(bridgeTransfer.amount);
```
andyjsbell, line 82: we might want to check if this doesn't fail silently due to the recipient being a contract

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        });
+
+        emit BridgeTransferInitiated(_bridgeTransferId, _originator, _recipient, _hashLock, _timeLock);
+        return _bridgeTransferId;
+    }
+
+    function completeBridgeTransfer(bytes32 _bridgeTransferId, bytes32 _secret) external override {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[_bridgeTransferId];
+        require(bridgeTransfer.exists, "Bridge transfer does not exist");
+        require(keccak256(abi.encodePacked(_secret)) == bridgeTransfer.hashLock, "Invalid secret");
```
andyjsbell, line 74: Actually we could just check on this maybe

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+            hashLock: _hashLock,
+            timeLock: block.timestamp + _timeLock,
+            exists: true
+        });
+
+        emit BridgeTransferInitiated(_bridgeTransferId, _originator, _recipient, _hashLock, _timeLock);
+        return _bridgeTransferId;
+    }
+
+    function completeBridgeTransfer(bytes32 _bridgeTransferId, bytes32 _secret) external override {
```
andyjsbell, line 71: would this warrant a reentrancy modifier(openzepplin) as we use `payable`

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        uint _timeLock
+    ) external payable override returns (bytes32 _bridgeTransferId) {
+        console.log("initiateBridgeTransfer");
+        uint256 totalAmount = _wethAmount;
+
+        // If msg.value is greater than 0, convert ETH to WETH and add to total amount
+        if (msg.value > 0) {
+            weth.deposit{value: msg.value}();
+            //Transfer WETH to this contract, revert if transfer fails
+            require(weth.transfer(address(this), msg.value), "WETH transfer failed");
```
andyjsbell, line 39: Isn't this covered by the `weth.deposit` call already?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        delete bridgeTransfers[_bridgeTransferId];
+
+        payable(bridgeTransfer.recipient).transfer(bridgeTransfer.amount);
+
+        emit BridgeTransferCompleted(_bridgeTransferId, _secret);
+    }
+
+    function refundBridgeTransfer(bytes32 _bridgeTransferId) external override {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[_bridgeTransferId];
+        require(bridgeTransfer.exists, "Bridge transfer does not exist");
```
andyjsbell, line 89: could use `bridgeTransfer.timeLock > 0` here instead

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+        require(block.timestamp > bridgeTransfer.timeLock, "Timelock has not expired");
+
+        delete bridgeTransfers[_bridgeTransferId];
+
+        payable(bridgeTransfer.originator).transfer(bridgeTransfer.amount);
```
andyjsbell, line 94: check for silent failure

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+
+        emit BridgeTransferRefunded(_bridgeTransferId);
+    }
+
+    function getBridgeTransferDetail(bytes32 _bridgeTransferId) external view override returns (
```
andyjsbell, line 99: we get this out of the box I think so seems like we can remove this entire function

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+            bridgeTransfer.exists, 
+            bridgeTransfer.amount,
+            bridgeTransfer.originator,
+            bridgeTransfer.recipient,
+            bridgeTransfer.hashLock,
+            bridgeTransfer.timeLock
+        );
+    }
+
+    function getCompletedBridgeTransferDetail(bytes32 _bridgeTransferId) external view returns (
```
andyjsbell, line 118: same here also

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+
+    mapping(bytes32 => BridgeTransfer) public bridgeTransfers;
+    mapping(bytes32 => BridgeTransfer) public completedBridgeTransfers;
+    IWETH10 public weth;
+
+    constructor(address _weth) {
+        weth = IWETH10(_weth);
+    }
+
+    function initiateBridgeTransfer(
```
andyjsbell, line 25: we might want to check these params for validity to avoid accidental locking of funds...

andyjsbell: wouldn't the recipient but on the destination chain, which in our case is Movement, how would this type `address` translate exactly?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: Yes let's add it.

protocol-units/bridge/contracts/src/IAtomicBridgeInitiator.sol
```diff
@@ -0,0 +1,60 @@
...
+     * @dev Returns the details of a specific bridge transfer
+     * @param _bridgeTransferId Unique identifier for the bridge transfer
+     * @return exists Boolean indicating if the bridge transfer exists
+     * @return amount The amount of assets to be allocated and sent
+     * @return originator The address allowed to withdraw (claim) the funds
+     * @return recipient The address intended to receive the assets
+     * @return hashLock The hash of the secret that will unlock the funds
+     * @return timeLock The timestamp until which this BridgeTransfer is valid
+    **/
+    function getBridgeTransferDetail(bytes32 _bridgeTransferId) external view returns (
```
andyjsbell, line 52: Not needed

andyjsbell (commented):
good start I think, we can probably improve a few things

edit: do we need to include all the interfaces within `WETH` folder?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: Yes this was a todo in my mind. Would have to be a hardcoded string

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xPrimata: actually, contract is making the deposit!

0xmovses: Yes, this was so that it works with @cde8's traits and original design.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,137 @@
...
+                _timeLock, 
+                block.timestamp
+        ));
+
+        // Check if the bridge transfer already exists
+        require(bridgeTransfers[_bridgeTransferId].amount == 0, "Bridge transfer already exists");
+
+        bridgeTransfers[_bridgeTransferId] = BridgeTransfer({
+            amount: totalAmount,
+            originator: _originator,
```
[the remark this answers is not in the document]
0xmovses: yeah makes sense, if we move `_originator` from the fn sig it will break @cde8's traits for the service. Tagging here so here is aware, remove it from the fn input and `msg.sender`.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
andyjsbell: Does this mean it will be resolved now?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xPrimata: Believe we can keep it as it is since we are storing weth. This would also trigger a reentrancy, but bridgeTransfer.complete toggle prevents a reentrancy.

log_file.txt
andyjsbell, line 1: Do we need this file?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        external
+        payable
+        returns (bytes32 _bridgeTransferId)
+    {
+        address originator = msg.sender;
+        uint256 ethAmount = msg.value;
+        uint256 totalAmount = _wethAmount + ethAmount;
+        _bridgeTransferId = keccak256(abi.encodePacked(originator, _recipient, _hashLock, _timeLock, block.timestamp));
+        // Check if the bridge transfer already exists
+        if (bridgeTransfers[_bridgeTransferId].amount != 0) revert BridgeTransferExists();
```
cde8, line 45: would call this error `BridgeTransferInvalid` or something.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        // Ensure there is a valid total amount
+        if (totalAmount == 0) {
+            revert ZeroAmount();
+        }
+        // If msg.value is greater than 0, convert ETH to WETH
+        if (ethAmount > 0) weth.deposit{value: ethAmount}();
```
cde8, line 51: Are gas fees included in the ETH amount? Just to double-check, if not, it could lead to extra costs otherwise.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+
+        //Transfer WETH to this contract, revert if transfer fails
+        if (_wethAmount > 0) {
+            if (!weth.transferFrom(originator, address(this), _wethAmount)) revert WETHTransferFailed();
+        }
+        bridgeTransfers[_bridgeTransferId] = BridgeTransfer({
+            amount: totalAmount,
+            originator: originator,
+            recipient: _recipient,
```
cde8, line 60: remove the underscores

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+            hashLock: _hashLock,
+            timeLock: block.timestamp + _timeLock,
+            completed: false
```
cde8, line 63: Can we also check if the bridge implementation allows us to retrieve a list of all bridgeTransferIds from the bridgeTransfers map in the contract? I need this to rebuild the state of the bridge if it goes down. If we cannot iterate over the stored bridgeTransfers, then keeping past transfers may not be necessary. This is an important question to answer as it is crucial for rebuilding the state of the Bridge Service.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        });
+
+        emit BridgeTransferInitiated(_bridgeTransferId, originator, _recipient, totalAmount, _hashLock, _timeLock);
+        return _bridgeTransferId;
+    }
+
+    function completeBridgeTransfer(bytes32 _bridgeTransferId, bytes32 _secret) external onlyAuthorized {
```
cde8, line 70: Could you rename secret to pre_image and also remove the underscores here.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        // Retrieve the bridge transfer
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[_bridgeTransferId];
+        uint256 amount = bridgeTransfer.amount;
+        if (amount == 0) revert NonExistentBridgeTransfer();
```
cde8, line 74: This state could never occur due to initialize_bridge_transfer function, waste of gas?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        if (bridgeTransfer.completed) revert BridgeTransferHasBeenCompleted();
+        if (keccak256(abi.encodePacked(_secret)) != bridgeTransfer.hashLock) revert InvalidSecret();
+
+        bridgeTransfer.completed = true;
+
+        // Mock transfer to recipient by serializing the address
+        // this is supposed to be removed in production
+        address recip;
```
cde8, line 82: @0xPrimata, this is incorrect. On the initiator contract, the bridging service completes the BridgeTransfer, which means we claim the funds, and now the funds belong to the bridge. As such, no further transfer is needed.

cde8 (changes requested):
First iteration of the review; will continue later.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+        bytes32 hashLock;
+        uint256 timeLock;
+        bool completed;
+    }
+
+    mapping(bytes32 => BridgeTransfer) public bridgeTransfers;
+    IWETH10 public weth;
+    address public authority;
+
+    modifier onlyAuthorized() {
```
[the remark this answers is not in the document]
0xmovses: Additionaly, We are never checking `authority` at any point in the contract, so it seems redunant and will add gas. I'll remove, unless you can think of any maintance, or functionality uses that only an authorized wallet should be ablle to do.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,113 @@
...
+
+    function initiateBridgeTransfer(uint256 _wethAmount, bytes32 _recipient, bytes32 _hashLock, uint256 _timeLock)
+        external
+        payable
+        returns (bytes32 _bridgeTransferId)
+    {
+        address originator = msg.sender;
+        uint256 ethAmount = msg.value;
+        uint256 totalAmount = _wethAmount + ethAmount;
+        _bridgeTransferId = keccak256(abi.encodePacked(originator, _recipient, _hashLock, _timeLock, block.timestamp));
```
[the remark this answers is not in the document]
0xmovses: I've added a `nonce` which increments after all the checks pass here and use that `nonce` inside the `keccak256`. Let me know what you think.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: I checked on this. Gas fees are deducted from `msg.value` before the function call happens. So `ethAmount` would not include gas fees.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
0xmovses: It was being used in `completeBridgeTansfer`, but as discussed in spain, we want the function to be callable by anyone. And this would allow the service to be decentralized with incentives later down the line. 

I've added Open Zeppelin Access Control and used it to only restrict the `refundBridgeTransfer` call. This function would only be called if for some reason we needed to refund funds.

0xmovses:
We're going to switch WETH10.sol for WETH9.sol. Updates incoming

protocol-units/bridge/contracts/src/IAtomicBridgeInitiator.sol
```diff
@@ -0,0 +1,56 @@
...
+    error TimeLockNotExpired();
+    error ZeroAddress();
+    error Unauthorized();
+
+    /**
+     * @dev Creates a new atomic bridge transfer using native ETH
+     * @param _wethAmount The amount of WETH to send
+     * @param _recipient The address on the other chain to which to transfer the funds
+     * @param _hashLock The hash of the secret (HASH) that will unlock the funds
+     * @param _timeLock The number of blocks until which this BridgeTransfer is valid and can be executed
```
andygolay, line 33: This says number of blocks for `_timeLock` but in the code implementation it seems to be using time in seconds. It seems like research is needed to be sure about which approach will work best.

andygolay (commented):
Generally this looks like it represents the Initiator flow in the PDF shared. Working out whether it's number of blocks or seconds for the time lock is important to resolve, but maybe that can be decided in tandem with bridge service development.

protocol-units/bridge/contracts/src/IAtomicBridgeInitiator.sol
0xmovses: Yep, well noticed. On a discussion with @cde8  we've decided to go with `block.timestamp` on eth side and the now timestamp on move which is in UTC secs. It will only be in e2e testing where we see if there are any issues with it. We walked through it yesterday and it seems to be solid.

0xmovses:
@andygolay Yes this is for the Eth initiator contract. Eth counterparty will be a seperate contract.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: I looked into this. There is no way solidiy allows you to get all the keys from the mapping, so the only way is to store a seperate array of all the keys, then write a function
```
function getAllBridgeTransferIds() external view returns (bytes32[] memory) 
```
We need the getter function because the inbuilt get method on an array in solidity can only get a single element in the array by index. The getter function I've added will return the entire array.

protocol-units/bridge/contracts/test/AtomicBridgeInitiator.t.sol
```diff
@@ -0,0 +1,208 @@
...
+        vm.warp(block.timestamp + timeLock + 1);
+        vm.startPrank(originator);
+        atomicBridgeInitiator.refundBridgeTransfer(bridgeTransferId);
+
+        (,,,,, bool completed) = atomicBridgeInitiator.bridgeTransfers(bridgeTransferId);
+        assertTrue(completed);
+
+        vm.stopPrank();
+    }
+}
```
cde8, line 208: Please add a test showing how to iterate over the stored bridge transfers from your map by adding 10 or 20 of them. We should be able to iterate through all of these entries to assess functionality requirements for the bridge.

Ideally, to restore the atomic bridge state, remove entries once they are completed to streamline the state--this makes the restore of the system lighter.

If you want to "store" all bridge transfers, could you please provide the rationale why?
Also, if valid, may I ask you to consider using two separate maps to distinguish between completed and incomplete transfers? I need quick access to all incomplete items to restore blockchain state.

protocol-units/bridge/contracts/src/IAtomicBridgeInitiator.sol
```diff
@@ -0,0 +1,56 @@
...
+        payable
+        returns (bytes32 _bridgeTransferId);
+
+    /**
+     * @dev Completes the bridging Counterparty
+     * @param _bridgeTransferId Unique identifier for the BridgeTransfer
+     * @param pre_image The secret that unlocks the funds
+     *
+     */
+    function completeBridgeTransfer(bytes32 _bridgeTransferId, bytes32 pre_image) external;
```
cde8, line 48: add an underscore to pre_image

cde8 (changes requested):
The review comments largely involves on the ability to iterate over bridge transfers and the method by which we internally store BridgeTransfers. This is crucial for how the BridgingService will restore its state. It is not only pertinent to this contract but also, for example, on the move side. Are we able to retrieve and iterate over maps within those contracts.

protocol-units/bridge/contracts/src/WETH/README.md
```diff
@@ -0,0 +1,38 @@
+# WETH10
```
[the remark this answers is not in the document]
0xmovses: Left over readme, we're just using `WETH9.sol` in the tests now. Will remove

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,99 @@
...
+        // WETH remains stored in the contract
+        // Only to be released upon bridge transfer in the oppisite direction
+
+        bridgeTransfer.completed = true;
+        emit BridgeTransferCompleted(bridgeTransferId, pre_image);
+    }
+
+    function refundBridgeTransfer(bytes32 bridgeTransferId) external {
+        // Check that the calling account has the refunder role
+        require(hasRole(REFUNDER_ROLE, msg.sender), "Caller is not a minter");
```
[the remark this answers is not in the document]
0xmovses: This is weird, this was removed. It says outdated, I think you're commenting on outdated code.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (block.timestamp < bridgeTransfer.timeLock) revert TimeLockNotExpired();
+
+        bridgeTransfer.completed = true;
+
+        if (!weth.transfer(bridgeTransfer.originator, amount)) revert WETHTransferFailed();
+        emit BridgeTransferRefunded(bridgeTransferId);
+    }
+
+    // Get all bridge transfer ids, the solidity get method only returns one item by index
+    function getAllBridgeTransferIds() external view returns (bytes32[] memory) {
```
cde8, line 93: Opting for this solution might initially seem adequate; however, it is important to reconsider our approach to working with the completed component, as maintaining it could lead to unsustainable scalability in the long term. Consequently, we must carefully evaluate the maximum return limits to prevent potential errors.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,96 @@
...
+        uint256 amount;
+        address originator;
+        bytes32 recipient;
+        bytes32 hashLock;
+        uint256 timeLock;
+        bool completed;
+    }
+
+    mapping(bytes32 => BridgeTransfer) public bridgeTransfers;
+    bytes32[] public bridgeTransferIds;
```
cde8, line 19: When we opt for this solution, we should definitely avoid storing all  transfers, as this could become excessively large over time. Consequently, since you have chosen this approach, it means we cannot iterate over the mapping, as such my intuition was correct. Therefore, let's redesign the requirements here.

cde8 (changes requested):
Can I conclude that we cannot iterate over a mapping as one might expect? If that is the case, then we need to redesign how I begin to recover the bridge state. We should also make a decision about whether we want to store all bridge transfers or only the relevant ones, right? I'm not sure about storage, but storing everything seems excessive because it has no value if we cannot iterate through it. Perhaps we should discard the completed transfers and only keep the live ones.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: I'm not sure there is a max return limit? Or if there is, we would never hit that limit, not infinite but a very large number

0xmovses:
We can iterate over the values of a mapping, but no way to iterate over the keys. That's why I went for storing the keys in a seperate array.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,96 @@
...
+        if (keccak256(abi.encodePacked(preImage)) != bridgeTransfer.hashLock) revert InvalidSecret();
+        bridgeTransfer.completed = true;
+        emit BridgeTransferCompleted(bridgeTransferId, preImage);
+    }
+
+    function refundBridgeTransfer(bytes32 bridgeTransferId) external {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[bridgeTransferId];
+        uint256 amount = bridgeTransfer.amount;
+        if (bridgeTransfer.completed) revert BridgeTransferHasBeenCompleted();
+        if (block.timestamp < bridgeTransfer.timeLock) revert TimeLockNotExpired();
```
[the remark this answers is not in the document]
0xmovses: We discussed this yesterdar right? That there would be no way to sync block time between the chains. Maybe I misunderstood but I thought we settled on block.timestamp as sufficient, if comparing with the move timestamp now() method.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: But we need to store all this data on chain, so that indexers could show historical data. The cost of storing data in a mapping does not directly depend on the "height" or the number of entries already present in the mapping. The cost of storing a new entry in a mapping is constant, meaning it does not increase as more entries are added

 Many contracts store all their data in a mapping.

0xmovses: A mapping in solidity is a hash table

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: As discussed, remove as indexing will be done via blockheigh, replaying history and event listening. No need for this funciton anymore

0xmovses:
So as I mentioned in our discussion, I've opted *not* to delete any entries of `BridgeTransferDetails` this is so that a user can query and get a `BridgeTransfer`, with a key `bridgeTransferId`, how they got that key, would be via listening to events. 

We want this data to be queryable. We take no gas cost for storing all the data in the mapping because the cost of storing a new entry in a mapping is constant, meaning it does not increase as more entries are added.

I also added an extra assertion to check WETH balance after a refund. 

Indexing of historical data will be done via the normal strategy, historical playback of events within a blockheight range.

protocol-units/bridge/contracts/src/WETH/README.md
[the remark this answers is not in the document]
0xmovses: This was left on outdated code, it was removed

protocol-units/bridge/contracts/test/AtomicBridgeInitiator.t.sol
0xmovses: As discussed, we will not be going with the strategy for indexing historical data. This would be done by indexing historical data within a blockheight range and listening for those events.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
cde8: That is interesting. In my memory, we should look at actually, let's say, how time is actually secured on Ethereum. If that would be done by the validators, then we might do time, but also we get all this kind of time zone stuff going on. And thus we came to the conclusion that block height would be more absolute point in time because we know the moment the swap gets initiated what block height, what the height is of the block and we have an absolute number of blocks we will allow the bridge transfer to complete.

The other option would be kind of, let's say, to use an oracle, but then you're actually delegating the whole thing (responsibility) to oracles, right, which is again kind of an indirection. As such, we concluded, in my recollection, block height is probably the best way to go.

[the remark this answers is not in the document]
cde8: Let's  just use block height on both sides because it's the most robust thing. We can script the logic inside the bridging service to take into account the average block time plus minus something and we take the range big enough. It is the most objective most consistent way after thinking about it. It will prevent all kinds of issues down the line.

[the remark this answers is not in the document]
0xmovses: Ok that's done

andygolay (approved):
Approving given Andy B's guidance about being able to work out issues as we put the entire bridge together and not getting too hung up on perfectly implementing each piece in the first iterations.

cde8 (changes requested):
I find the statement, checking for amount, somewhat unusual, particularly since it describes a state that cannot be realized. Can we not just test for existence in solidity? Last one

0xPrimata (changes requested):
Dropped a few comments

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,90 @@
...
+        if (wethAmount > 0) {
+            if (!weth.transferFrom(originator, address(this), wethAmount)) revert WETHTransferFailed();
+        }
+
+        nonce++; //increment the nonce
+        bridgeTransferId =
+            keccak256(abi.encodePacked(originator, recipient, hashLock, timeLock, block.timestamp, nonce));
+
+        // Check if the bridge transfer already exists
+        if (bridgeTransfers[bridgeTransferId].amount != 0) revert BridgeTransferInvalid();
```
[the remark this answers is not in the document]
cde8: Additionally, the error message currently does not accurately reflect the violation. It should instead be labeled as "BridgeTransferAlreadyExists" to provide a specific error that enables the Bridge to manage it effectively.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,90 @@
...
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[bridgeTransferId];
+        if (bridgeTransfer.state == MessageState.COMPLETED) revert BridgeTransferHasBeenCompleted();
+        if (keccak256(abi.encodePacked(preImage)) != bridgeTransfer.hashLock) revert InvalidSecret();
+        bridgeTransfer.state = MessageState.COMPLETED; 
+        emit BridgeTransferCompleted(bridgeTransferId, preImage);
+    }
+
+    function refundBridgeTransfer(bytes32 bridgeTransferId) external {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[bridgeTransferId];
+        uint256 amount = bridgeTransfer.amount;
```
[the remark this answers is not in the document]
0xmovses: Do you mean `if (bridgeTransfer.state == MessageState.REFUNDED_revert BridgeTransferIsNotRefundable();`?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
0xmovses: we could check on timelock that's probably better. I think this was here in an earlier version ..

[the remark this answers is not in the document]
cde8: Given the characteristics of the hashing function and nonce, it is impossible to generate a previously existing bridgeTransferId. Therefore, the corresponding statement can be omitted.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,90 @@
...
+            state: MessageState.INITIALIZED 
+        });
+
+        emit BridgeTransferInitiated(bridgeTransferId, originator, recipient, totalAmount, hashLock, timeLock);
+        return bridgeTransferId;
+    }
+
+    function completeBridgeTransfer(bytes32 bridgeTransferId, bytes32 preImage) external {
+        BridgeTransfer storage bridgeTransfer = bridgeTransfers[bridgeTransferId];
+        if (bridgeTransfer.state == MessageState.COMPLETED) revert BridgeTransferHasBeenCompleted();
```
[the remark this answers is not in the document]
0xmovses: There is no PENDING state, assume you mean INITALIZED, I'll add that 

The function can be called at any time, its only on the counterparty side that timelock is checked on `complete_bridge_transfer`. No funds are actually moved here. @cde8 What do you think should we check the timelock to restrict the `MessageState` mutation?

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
0xmovses: yes no need for the extra compute. Every call here will generate a unique id.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
cde8: The function completeBridgeTransfer can only be invoked once it has been initialized. Additionally, this initialization permits the completion of transactions in the REFUNDED state.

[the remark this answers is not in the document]
0xmovses: Added `if (bridgeTransfer.state != MessageState.INITIALIZED) revert BridgeTransferHasBeenCompleted();`

No check on timelock as that's done on counterparty side, and save on compute. but can add it if we want.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
```diff
@@ -0,0 +1,90 @@
...
+        if (keccak256(abi.encodePacked(preImage)) != bridgeTransfer.hashLock) revert InvalidSecret();
+        bridgeTransfer.state = MessageState.COMPLETED; 
+        emit BridgeTransferCompleted(bridgeTransferId, preImage);
+    }
+
+    function refundBridgeTransfer(bytes32 bridgeTransferId) external {
```
cde8, line 76: Refunds are only permissible when they are in the initialized state.

cde8 (commented):
Control the transitions between states through gating.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: I just pushed changes, your commenting on outdated code. 

It's now just `if (bridgeTransfer.state != MessageState.INITIALIZED) revert BridgeTransferHasBeenCompleted();`

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: Updated

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
0xmovses: Added timelock check on this side 👌

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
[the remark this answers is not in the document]
0xPrimata: @cde8 if you access a non-existent resource, it will return default type values like 0, 0x0 or false.

e.g. a non-existent bridgeTransferId amount would return 0. Since bridge transfers must have over 0 amount, we can safely assume that if a bridgeTransferId has an amount different than 0, then it exists.

0xPrimata (approved):

Issue #115, opened 2024-06-11
Integrate block reversion


mzabaluev:
`BlockCommitmentEvent::Rejection` events are not yet handled by the node, this needs to be implemented.

PR #116, opened 2024-06-11
ENG-434: Integration of MCR settlement

mzabaluev:
# Summary
- **Categories**: `networks`.

Integrate the MCR settlement client and perform processing of block acceptance events by updating the `fin` view.
This gets us a minimally functional full node without block reversion.

# Changelog

* Minimal integration of a Suzuka full node without block reversion.

# Testing

No new tests. E2E tests should be responsible for testing the full node, to be added in #105.

# Outstanding issues

#115


networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -196,15 +199,23 @@ where
...
 	while let Some(res) = stream.next().await {
 		let event = res?;
 		match event {
 			BlockCommitmentEvent::Accepted(commitment) => {
 				debug!("Commitment accepted: {:?}", commitment);
+				executor.set_finalized_block_height(commitment.height)?;
 			}
 			BlockCommitmentEvent::Rejected { height, reason } => {
 				debug!("Commitment rejected: {:?} {:?}", height, reason);
+				// TODO: block reversion
```
0xmovses, line 218: More of a note for later but I think we should be adding some more `info!()` traces for logging in production. Here would be a useful place to add one.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -3,84 +3,61 @@ use crate::send_eth_tx::SendTxErrorRule;
...
 
-#[derive(Clone, Debug)]
-pub struct McrEthSettlementConfig {
-	pub mrc_contract_address: String,
+/// Configuration of the MCR settlement client.
+///
+/// This structure is meant to be used in serialization.
+/// Validation is done by the builder interface of the [`Client`].
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
```
mzabaluev, line 43: This could be moved into a separate micro-crate to trim dependencies for suzuka-config. Keeping it here to ease reviewing for this PR.

0xmovses (approved):

l-monninger (approved):

l-monninger:
This is good on the rust side. But, I think this is working mainly because the defaults still match up with the environment variables used in the deployer script. What we will want to do is rewrap the deployment in a rust `mcr-setup` binary.

PR #122, opened 2024-06-13
ENG-591 move side bridge modules and `moveth.move`

0xmovses:
# Summary
Adds the aptos-move modules for the counterparty aptos side bridge.

# Changelog
- Adds `moveth.move`
- Creates `moveth_tests.move`
- Creates `atomic_bridge.move` with in-file tests. 

@cde8 `complete_bridge_transfer` is now callable by any account that has the minter role for `moveth`, if its not a minter of moveth, the function will fail. On the move side we just need the deployer account of `moveth` to allow a client to mint tokens. 

By design move doesn't allow you to store a `&signer` type inside a struct, because it does not have the capability `store`. So therefor, it has to be the &signer that is calling `complete_bridge_transfer` that is a moveth minter. We can't access the deployer or some other minter signer internally from within the contract

# Testing
`cd protocol-units/bridge/move-modules && aptos move test --dev`

# Outstanding issues


protocol-units/bridge/move-modules/.aptos/config.yaml
andygolay, line 1: @0xmovses is this `.aptos` directory and `config.yaml` meant to be included in the PR? I would think it would make more sense to gitignore it. 

In this case it seems to me like in `config.yaml` the `rest_url` of `https://fullnode.devnet.aptoslabs.com` could be misleading if we're deploying to Movement rather than Aptos.

protocol-units/bridge/contracts/src/IAtomicBridgeInitiator.sol
andygolay, line 1: - As far as I can tell from how it is used in `AtomicBridgeInitiator`, `_timelock` is in seconds, not number of blocks. Correct `IAtomicBridgeInitiator.sol` comments to reflect this.

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
andygolay, line 1: - In `AtomicBridgeInitiator`, `BridgeTransfer` is defined with an Eth address field `recipient`. In the Solidity contract, `recipient` is the Eth address where WETH is sent to complete a transfer. However, on the Move side, MovETH is minted to a `recipient` Aptos address. Is `BridgeService` doing something in-between to get the Aptos address? Or should there be separate `eth_recipient` and `move_recipient` fields in the `BridgeTransfer` struct, as EVM and Aptos addresses aren't compatible?

protocol-units/bridge/move-modules/sources/atomic_bridge.move
andygolay, line 1: - On the Move side, I see `initiator: &signer,` which seems to indicitate that the person who is signing the transaction with a Move wallet using an Aptos address, is perhaps supposed to be the the initiator of the bridge transfer. But `BridgeTransfer` on the Eth side has an Eth address as `initiator`. I wonder if a more descriptive variable name than `initiator` therefore makes sense on the Move side, depending on whether `BridgeService` or the originator of the bridge transfer is the signer of a specific function call? 

It appears to me that `lock_bridge_transfer_assets` must be called by `BridgeService` and so it wouldn't be possible to infer the recipient Move address for `BridgeTransferDetails` from context provided when calling that function, unless that address is provided in the `BridgeTransfer` struct.

`complete_bridge_transfer` on the other hand must be called by the originator of the transfer because only they can provide `secret`. However, in that function, `recipient`'s value comes from the `BridgeTransferDetails` `details` and so it seems to me that the recipient should actually the signer in that case, because only the signer should be in possession of `secret`. 

- `complete_bridge_transfer` is called by `BridgeService` and it mints MovETH to `recipient`. However, the originator is the only one with `secret`. Doesn't the mint action need to be followed by the MovETH being locked up, and then there needs to be another function, `claim`, for the originator to submit `secret`? `BridgeService` should not be in possession of `secret` prior to `originator` receiving their MovETH.

As far as I can tell, `lock_bridge_transfer_assets` should be called by `BridgeService` and hsould mint and lock the MovETH. And `complete_bridge_transfer` should be called by the bridge transfer originator, who submits the secret in order to unlock and claim their MovETH.

Please let me know if I'm misunderstanding the flow.

andygolay, line 1: `complete_bridge_transfer` is called by `BridgeService` and it mints MovETH to `recipient`. However, the originator is the only one with `secret`. Doesn't the mint action followed by the MovETH being locked up, need to be done by `BridgeService`? 

And then there needs to be another function, `claim`, for the originator to submit `secret`? `BridgeService` should not be in possession of `secret` prior to `originator` receiving their MovETH.

Right now, `initiator` in `complete_bridge_transfer` is both minting the MovETH and submitting `secret`; it seems to me the minting and locking up should be one step, with `BridgeService` calling that function after detecting locked assets on Ethereum.

And claiming should be another step, called by the originator of the bridge transfer.

In `complete_bridge_transfer` the signer (`initiator`) mints the MovETH. So it seems like there must be some issue, because isn't `BridgeService` supposed to mint the MovETH? But `BridgeService` should not be in possession of the secret. 

Let me know if / how I'm misunderstanding this, if I'm getting it wrong?

I'm mostly going off the Eth -> Move RFC flow chart https://github.com/movementlabsxyz/rfcs/pull/40/files

protocol-units/bridge/move-modules/.aptos/config.yaml
0xmovses: Yes will remove the `.aptos` dir and update the `config.yaml`

protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol
0xmovses: The `BridgeService` handles generic address types. Upon implementation their conversation is handled. You can checkout the draft PR for the bridge service here https://github.com/movementlabsxyz/movement/pull/133

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+            amount,
+            hash_lock,
+            time_lock: timestamp::now_seconds() + time_lock
+        };
+
+        smart_table::add(&mut bridge_store.pending_transfers, bridge_transfer_id, details);
+        event::emit(
+            BridgeTransferAssetsLockedEvent {
+                bridge_transfer_id,
+                caller: signer::address_of(caller),
```
cde8, line 90: See comment above

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+
+        true
+    }
+
+    public fun complete_bridge_transfer(
+        caller: &signer,
+        bridge_transfer_id: vector<u8>,
+        secret: vector<u8>
+    ) acquires BridgeTransferStore, BridgeConfig {
+        let bridge_store = borrow_global_mut<BridgeTransferStore>(signer::address_of(caller));
```
cde8, line 106: The client is the one who will call this function. It can be called by anyone who possesses the secret, as you explained. Make sure this is callable by anyone.

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+        let details: BridgeTransferDetails = smart_table::remove(&mut bridge_store.pending_transfers, bridge_transfer_id);
+        assert!(details.recipient != @0x0, 1);
```
cde8, line 108: This cannot be the case

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+        caller: &signer,
+        bridge_transfer_id: vector<u8>,
+        secret: vector<u8>
+    ) acquires BridgeTransferStore, BridgeConfig {
+        let bridge_store = borrow_global_mut<BridgeTransferStore>(signer::address_of(caller));
+        let details: BridgeTransferDetails = smart_table::remove(&mut bridge_store.pending_transfers, bridge_transfer_id);
+        assert!(details.recipient != @0x0, 1); 
+
+        // check secret against details.hash_lock
+        let config = borrow_global<BridgeConfig>(signer::address_of(caller));
```
cde8, line 111: config is never used

cde8 (changes requested):
We need to redo this when changes have been integrated

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+        smart_table::add(&mut bridge_store.completed_transfers, bridge_transfer_id, details);
+        event::emit(
+            BridgeTransferCompletedEvent {
+                bridge_transfer_id,
+                secret,
+            },
+        );
+    }
+
+    public fun abort_bridge_transfer(
```
andygolay, line 129: If `abort_bridge_transfer` is meant to be called only after the time lock expires, shouldn't there be some logic in the function to check that the time lock has indeed expired?

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+            },
+        );
+
+        true
+    }
+
+    public fun complete_bridge_transfer(
+        caller: &signer,
+        bridge_transfer_id: vector<u8>,
+        secret: vector<u8>
```
andygolay, line 104: Rename `secret` to `pre_image` throughout, for consistency?

protocol-units/bridge/move-modules/sources/atomic_bridge.move
andygolay, line 106: It looks to me like it should be borrowing the `BridgeTransferStore` using the deployer address rather than the caller address, because anyone can call `complete_bridge_transfer`.

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,239 @@
...
+
+        // check secret against details.hash_lock
+        let config = borrow_global<BridgeConfig>(signer::address_of(caller));
+
+        // Check secret against details.hash_lock
+        let computed_hash = keccak256(secret);
+        assert!(computed_hash == details.hash_lock, 2);
+
+        // Mint moveth tokens to the recipient, caller must me a minter of moveth
+        moveth::mint(caller, details.recipient, details.amount);
```
andygolay, line 118: In `complete_bridge_transfer` we can't be sure that `caller` is a MovETH minter, right?

Should the contract be the minter in this case?

protocol-units/bridge/contracts/cache/solidity-files-cache.json
andygolay, line 1: Does this cache file need to be included in the GitHub repo? I'm wondering whether it might potentially contain sensitive information. From what I've read, it's usually recommended to `.gitignore` it.

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -0,0 +1,236 @@
...
+
+    public fun abort_bridge_transfer(
+        caller: &signer,
+        bridge_transfer_id: vector<u8>
+    ) acquires BridgeTransferStore, BridgeConfig {
+        // check that the signer is the bridge_module_deployer
+        assert!(signer::address_of(caller) == borrow_global<BridgeConfig>(signer::address_of(caller)).bridge_module_deployer, 1);
+        let bridge_store = borrow_global_mut<BridgeTransferStore>(signer::address_of(caller));
+        let details: BridgeTransferDetails = smart_table::remove(&mut bridge_store.pending_transfers, bridge_transfer_id);
+        assert!(block::get_current_block_height() > details.time_lock, 2);
```
andygolay, line 132: This test fails, apparently due to needing to initialize first.

See https://github.com/aptos-labs/aptos-core/issues/5201 for fixes

andygolay (changes requested):
Mind fixing the test for querying `block_height` and then I'll re-review?

protocol-units/bridge/contracts/cache/solidity-files-cache.json
0xmovses: Will remove

protocol-units/bridge/move-modules/sources/atomic_bridge.move
0xmovses: Soz was moving fast, they were passing and I pushed pre-emptively.

0xmovses:
Current weirdness when trying to aptos_framwork init blocks within test code. With the suggested steps it fails with:
```
error[E11001]: test failure
│    ┌─ /Users/movses/.move/https___github_com_aptos-labs_aptos-core_git_devnet/aptos-move/framework/aptos-framework/sources/block.move:97:9
│    │
│ 93 │     public(friend) fun initialize(aptos_framework: &signer, epoch_interval_microsecs: u64) {
│    │                        ---------- In this function in 0x1::block
│    ·
│ 97 │         move_to<CommitHistory>(aptos_framework, CommitHistory {
│    │         ^^^^^^^ Test was not expected to error, but it gave a RESOURCE_ALREADY_EXISTS (code 4004) error with error message: "Failed to move resource into 0000000000000000000000000000000000000000000000000000000000000001". Error originating in the module 0000000000000000000000000000000000000000000000000000000000000001::block rooted here
```

This isn't related to module implementation but some weirdness within aptos_framework. Timestamp test initialization works (of course). I'm moving on to the services for now.

andygolay (approved):
lgtm

cde8 (changes requested):
I have noticed a considerable number of cache and output files in the repository, which complicates the review process. Could you please remove them before I commence my review?

0xmovses:
@cde8 That's done now.

0xmovses:
I'm implementing a resource account for the minting of `moveth.move`, please hold off on reviews till next commit.

cde8 (approved):

cde8 (approved):
Approved, enabling future development of this functionality. However, please note that the code has not been thoroughly reviewed by me in its entirety. Approval  requested from @0xmovses

PR #123, opened 2024-06-13
Downgrade Runners

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `cicd`

To reduce costs, switches `movement-runner` to `buildjet-8vcpu-ubuntu-2204` and only uses `macos-latest` on `cargo check`.

# Changelog

See above.

# Testing

CI passes. 

# Outstanding issues
1. Likely we will want to move to self-hosted runners on AWS.


0xmovses (approved):
RD :shipit:

PR #119, opened 2024-06-12
Docker Compose Rewrite

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**:`protocol-units`

Extensive changes across all protocol units to accommodate `docker-compose` setups and process runtimes. 

# Changelog

- Well known configs from https://github.com/movementlabsxyz/movement/pull/93 modified to support non-native service dependencies. Generally, this means adding to the config `listen_` and `connection_` hostnames and ports. 
- General cleanup of #93.
- Rewrite of the actual `docker-compose` setup to no longer use the monolith.
- Minor modifications to build and publish flow.
- **REMOVES MONZA** as it time consuming to maintain throughout this process. 

# Testing

- Manual testing only; CI tests not yet added besides those for builds.
- To confirm behavior run `just suzuka-full-node docker-compose setup.local`

# Outstanding issues
1. `just suzuka-full-node docker-compose setup.local` shell does not exit cleanly. Leaves user unable to see additional input. 
2. Still pretty messy. Better standardization and patterns should be introduced as this continues to grow.
3. Replacement flags for debugging are pretty much useless as the setup will always be one ChainID ahead.

PR #129, opened 2024-06-15
State Timestamp Fix

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

While running `complex-alice` e2e tests against the CLI, we encountered an issue with timestamps that lead to further investigation and several discoveries.

- **Note:** we were able to confirm execution is otherwise reasonable by removing code which subtraction underflows in the CLI.

- **Note:** Aptos CLI is not yet in the Nix flake. For the moment, it is best to build from source on the host and allow impure in your nix shell.  

## Issues discovered
1. **`BlockMetadata` hack is not necessary:** Blocks can execute successfully with their own `BlockMetadata` contained, provided that the genesis has been properly transitioned w.r.t. to this. See `rollover_genesis` implementation for more details. 
2. **Current execution will not hold across epochs:** Likely because of missing `StateCheckpoints`, the current execution will not hold across epochs. This was tested by simulated advanced timestamps on blocks. Lines for this can be uncommented to test and confirm.
    - We are okay to have one large epoch; this is dicussed further below. 
    - However, it is likely missing `StateCheckpoints` are the culprit of the 
3. **We still cannot execute `StateCheckpoints`**: the "not last transaction in block issue persists." The current best guess is that this also has to do with a faulty genesis. 
4. **One long epoch needs upstream mods**: if we wish to create a single long epoch, we will need to make modifications on `aptos-core` to prevent unsafe arithmetic for calculating rewards (which do not affect our `maptos-opt-executor`). 
5. **Misinterpretation of timestamps**: DA integration assumed microseconds when the DA was outputting seconds. Block-times were inaccurate. 

# Changelog

# Testing

# Outstanding issues


l-monninger:
For information about CLI bug: https://github.com/aptos-labs/aptos-core/issues/13704

l-monninger:
Initially, I thought that using the `BlockMetadata` hack was causing transactions to fail to be assigned to the right block in the state API. I mis-confirmed this conclusion by identifying `"transactions" : null` in the REST API response for `block_by_height`. However, I later confirmed this behavior was the same on the upstream.

It still seems that either state or transactions are not appropriately assigned. But, it appears much more likely to be a `StateCheckpoint` issue.

l-monninger:
It appears this is where the `StateCheckpoint` is appended automatically: https://github.com/movementlabsxyz/aptos-core/blob/ea91067b81f9673547417bff9c70d5a2fe1b0e7b/execution/executor/src/components/apply_chunk_output.rs#L246

l-monninger:
> It appears this is where the StateCheckpoint is appended automatically: https://github.com/movementlabsxyz/aptos-core/blob/ea91067b81f9673547417bff9c70d5a2fe1b0e7b/execution/executor/src/components/apply_chunk_output.rs#L246

In our case, we are getting `None` for `state_checkpoint_to_add`. So, somehow, we are ending up with multiple `StateCheckpoints`, but the don't see to be added here. Otherwise, there's some other transaction being added that's messing this up.

l-monninger:
I am also investigating where committing transactions to the mempool upon completion to see if that changes client behavior, as this was something we used to do in M1.

l-monninger:
> I am also investigating where committing transactions to the mempool upon completion to see if that changes client behavior, as this was something we used to do in M1.

If this works, we probably don't need to solve the checkpointing issue. We can just go with a long epoch--though that will require other mods.

l-monninger:
> I am also investigating where committing transactions to the mempool upon completion to see if that changes client behavior, as this was something we used to do in M1.

Initial implementation does not seem to make a difference.

l-monninger:
Okay, so one thing I want to check is whether `aptos_framework::time` works despite the above. It may be some clock that sits outside of the actual block execution flow, that is not properly synced.

mzabaluev:
Whenever we discover surprising behavior in aptos-core, it makes sense to check if it remains in the upstream main branch. Some things may be solved by pulling in upstream changes, which we want to do for other reasons as well.

l-monninger:
> Whenever we discover surprising behavior in aptos-core, it makes sense to check if it remains in the upstream main branch. Some things may be solved by pulling in upstream changes, which we want to do for other reasons as well.

This was in large part our mistake. We did not have correct timestamps on `LedgerInfo`.

l-monninger:
Closing in favor of #119

PR #133, opened 2024-06-18
ENG-560: Atomic Bridge Shared Crate

cde8:
# Summary
- **RFCs**: RFC040 Atomic bridge
- **Categories**: protocol-units

## Structure
Implemented the logic for managing "structure" by utilizing the atomic swap primitive. This shared crate incorporates a rigorously tested architecture for transferring assets between two blockchains through trait interfaces.

It is presumed that asset transfer is feasible when there are contracts present on both sides of the blockchain:

Blockchain 1:
Initiator
Counterparty

Blockchain 2:
Initiator
Counterparty

Blockchain 1 Initiator will work with Blockchain 2 Counterparty

and

Blockchain 2 Initiator will work with Blockchain 2 Counterparty

## Conversions
In scenarios involving multiple blockchains and distinct Rust blockchain clients, variations in hash and address formats are encountered. To address this, developers using this crate are required to encapsulate hash and address types and define traits for facilitating conversion, ensuring equality comparisons through PartialEq and Eq implementations. This approach enables seamless comparison of addresses and hashes across diverse blockchain environments.

## Testing
Thorough emphasis is placed on rigorously testing the bridging functionality, recognizing its pivotal role in ensuring operational integrity. I have developed an AbstractBlockchain framework that enables manipulation of time, facilitating comprehensive testing scenarios. These tests are designed to simulate diverse client behaviors, including false positives and random errors, enhancing reproducibility for debugging purposes through the use of a seed. This approach enables efficient detection of abnormalities within the bridging functionality, effectively simulating prolonged operational conditions in a condensed timeframe.

## Multiple sources of trust
Diversifying sources of trust is a critical element in the context of bridging, as the bridging service will interact with events that are consumed by client libraries. By ensuring the implementation of the contract and monitoring traits, the integration of multiple sources of truth can effectively function in conjunction with the respective client libraries.

## Integration
Initially, the integration of Ethereum and Aptos-like chains will involve the implementation of two additional crates utilizing their respective client libraries. These clients must adhere to the corresponding traits, namely BridgeInitiatorContract and BridgeCounterpartyContract. Furthermore, it is essential to implement the monitoring traits. Upon successful implementation of these traits within the crates, the BridgingService structure can be properly configured, resulting in a functional bridging service. This setup will also facilitate seamless expansion to other blockchain networks.

# Changelog
1. Added bridge-cli, bridge-service, and bridge-shared packages to handle different aspects of blockchain bridging.
2. Updated Cargo.toml files to include new workspace members for bridge protocol units.
3. Introduced new dependencies in Cargo.lock for added packages ensuring compatibility across asynchronous tasks and cryptographic functions.
4. Created main executables for CLI and service units with initial greeting message.
5. Developed foundational structures for managing blockchain events and interactions within the bridge-shared unit.

# Testing
- Ensured all unit tests cover the new functionalities introduced in the bridge-cli, bridge-service, and bridge-shared units.
- Added integration tests to simulate real-world transaction scenarios across blockchain services.
- Performed regression tests to confirm that existing functionality remains unaffected by the new updates.

# Outstanding issues
- Further integration testing with external blockchain services to guarantee robust performance under various network conditions.
- Expansion of test coverage to include error handling and failure scenarios.


protocol-units/bridge/shared/tests/bridge_service_error_handling.rs
andygolay, line 1: This all looks amazing, @cde8. One small comment / question: I notice you have this error handling file, and you also have `bridge_service/error_handling.rs`. Would it make sense to organize them together on the same level, as they both appear to deal with bridge service error handling?

protocol-units/bridge/cli/src/main.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
```
l-monninger, line 1: Was this meant to be left in?

protocol-units/bridge/shared/src/blockchain_service.rs
```diff
@@ -0,0 +1,179 @@
...
+			(_, Poll::Ready(Some(event))) => {
+				Poll::Ready(Some(ContractEvent::CounterpartyEvent(event)))
+			}
+			_ => Poll::Pending,
+		}
+	}
+}
+
+// Practical implementation
+
```
l-monninger, line 62: Can we get doc comments on these?

protocol-units/bridge/shared/src/bridge_service/active_swap.rs
```diff
@@ -0,0 +1,526 @@
...
+				state: ActiveSwapState::LockingTokens(
+					call_lock_bridge_transfer_assets::<BFrom, BTo>(counterparty_contract, details)
+						.boxed()
+						.timeout(Delay::new(self.config.contract_call_timeout)),
+					0,
+				),
+			},
+		);
+
+		self.waker.wake();
```
l-monninger, line 201: Can you explain the need in comments to implement our own wakers/runtime?

protocol-units/bridge/shared/tests/abstract_blockchain.rs
```diff
@@ -0,0 +1,163 @@
...
+
+	let transaction = Transaction::Initiator(InitiatorCall::InitiateBridgeTransfer(
+		initiator_address.clone(),
+		recipient_address.clone(),
+		amount,
+		time_lock.clone(),
+		hash_lock.clone(),
+	));
+
+	blockchain.transaction_sender.unbounded_send(transaction).unwrap();
```
l-monninger, line 74: I've commented this on other folks PRs. But, I prefer that we avoid unwraps even in tests. In tests because people often use tests as examples and copy code accordingly.

protocol-units/bridge/shared/tests/blockchain_service.rs
```diff
@@ -0,0 +1,47 @@
...
+			InitiatorAddress("initiator"),
+			RecipientAddress::from("recipient"),
+			HashLock("hash_lock"),
+			TimeLock(100),
+			Amount(1000),
+		)
+		.await
+		.expect("initiate_bridge_transfer failed");
+
+	let mut cx = Context::from_waker(futures::task::noop_waker_ref());
```
l-monninger, line 31: Again, it seems like our own runtime might be overkill here. If we're doing this for some kind ordering, shouldn't the assumption be that the `BlockchainService` will (and in fact must) provide its own ordering.

protocol-units/bridge/shared/tests/bridge_service.rs
```diff
@@ -0,0 +1,267 @@
...
+	let completed_event_counterparty =
+		completed_event_counterparty.B2C_ContractEvent().expect("Not a B2C event");
+	tracing::debug!(?completed_event_counterparty);
+	assert_eq!(
+		completed_event_counterparty,
+		&BridgeContractCounterpartyEvent::Completed(CompletedDetails {
+			bridge_transfer_id: Convert::convert(transfer_initiated_event.bridge_transfer_id()),
+			recipient_address: RecipientAddress::from(BC2Address("recipient")),
+			hash_lock: HashLock(BC2Hash::from("hash_lock")),
+			secret: HashLockPreImage(b"hash_lock".to_vec()),
```
l-monninger, line 122: "Preimage" is one word in mathematics and cryptography. I don't think the camel-case is needed here.

l-monninger (changes requested):

l-monninger:
There aren't any concrete client implementations yet, right? @cde8

protocol-units/bridge/cli/src/main.rs
cde8: Hey Liam,  this was intentionally included. Think of it as a temporary placeholder for the to be developed bridge command line interface.

l-monninger:
Overall, there's a lot of abstraction here and not a lot doc comments. I think to give a proper review I'd like to see more doc comments, particularly as pertain to the usage of various portions of the API.

protocol-units/bridge/shared/tests/bridge_service_error_handling.rs
cde8: Great find @andygolay I included this as a rough sketch for the error handling process. I will remove it.

cde8: This is more a test of the client interface, as it should fail with an invalid secret; this is actually assumed and could be removed completely for clarity. I will remove this file.

protocol-units/bridge/shared/src/bridge_service/active_swap.rs
cde8: Could you clarify what you mean by implementing our own Waker or runtime?

protocol-units/bridge/shared/tests/abstract_blockchain.rs
cde8: I understand your perspective. The code in question is designed primarily for testing purposes, and it is not intended for broader use. The potential failure mode identified would predominantly occur due to insufficient memory. Converting this to an `expect` could be considered, although the necessity of such a change might be minimal. However, if you deem the modification essential, I am prepared to implement this adjustment. Please advise on how you wish to proceed.

protocol-units/bridge/shared/tests/blockchain_service.rs
cde8: I'm unsure about your references to "ordering" and "runtime," Liam. Could you clarify what you mean? However, I'm not clear on how this relates to implementing a runtime. There is no runtime implementation here. Only the check that items get cleaned up as they should. Could you please explain further?

protocol-units/bridge/shared/tests/bridge_service.rs
cde8: I have included the case for clarity. Would you prefer it to be renamed to HashLockPreimage? The same logic could also be applied to a HashLock for example. You would also consider that one word. Please give me your preferences and I will rename it towards whatever you like.

cde8:
> There aren't any concrete client implementations yet, right? @cde8

The concrete implementation will be done as soon as the movement and Aptos crates are developed, reviewed, and merged according to the original planning.

protocol-units/bridge/shared/src/bridge_service/active_swap.rs
l-monninger: You have implemented what looks something like coroutines--with wakers and polls--and with very similar semantics to how you would implement async runtimes in rust.

protocol-units/bridge/shared/tests/abstract_blockchain.rs
l-monninger: I will comment it again on a later PR or put it in an issue. I suppose we're going ahead with the merge here to unblock @0xmovses

protocol-units/bridge/shared/tests/blockchain_service.rs
l-monninger: You're implementing poll and waker logic at a lower level than simply relying on async methods. Why?

protocol-units/bridge/shared/tests/bridge_service.rs
l-monninger: It's a nit.

l-monninger (approved):

l-monninger:
@cde8 On second review, I see why you chose to implement `Future` and `Stream` and release manipulations of the `Waker` seem to be confined to tests.

protocol-units/bridge/shared/tests/blockchain_service.rs
cde8: If you are aiming for a higher degree of control beyond the use of async methods, particularly for detailed testing, it becomes necessary to adopt a more nuanced implementation approach. 

This subject might warrant a thorough, dedicated discussion on design tradeoffs beyond the scope of a pull request.

PR #135, opened 2024-06-18
Refactor setup APIs using `DocMovement`

mzabaluev:
# Summary

* Pass `DocMovement` by reference down the async call stacks where the object is not modified or moved.
* De-warn the `*SetupOperations` traits by desugaring the `async fn` definition (impls don't need to be changed), rename them all to `Setup` in the Rusty less verbose style.
* Rename a getter method in `DocMovement`, the `get_` prefix is un-Rusty.

# Changelog

# Testing

Purely a refactoring, existing tests should pass.


l-monninger:
Sorry, we were making similar changes at the same time.

l-monninger:
I may have an even more generalized and easier to use setup that I'll be looking to endeavor over the weekend. So, maybe let's try to pause on these.

PR #137, opened 2024-06-19
Godfig

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `misc`.

A KV store meant for use in configuration tasks. 

# Changelog

1. Introduces `godfig` crate.
2. Introduces `flocks` crate for file locks used to build godfig.
3. Provides an example usage of `godfig`-based manager pattern in `m1-da-light-node`. 

# Testing

1. Unit tests have been added to `godfig`. 
2. e2e tests have been run with updated `godfig` `m1-da-light-node` config.

# Outstanding issues
1. Not all units have been updated to use `godfig` for setup and runtime. The objective in a forthcoming PR would be to refactor according to `godfig`.
2. Better standards can be used and better ergonomics for `godfig` can be provided:
    1. No options in configs.
    2. Config version enums.
    3.  Enums for top-level key names. 
    4. Perhaps a macro for `godfig` top-level fields that derefs a lower_underscore_case field with the same name as the struct.
3. `godfig` should probably expose the backend streaming for more complex manager patterns which would pause or restart runners on changes. 
4. `flocks` can be improved by holding a global mapping to fd locks. Currently, mutual exclusion is only preserved between users of the same file handle and between processes. Within a process, users who generate a new lock would not be mutually excluded.


andyjsbell:
Great name, I should have trademarked it 😉

Cargo.toml
```diff
@@ -72,6 +72,7 @@ commander = { path = "util/commander" }
 suzuka-config = { path = "networks/suzuka/suzuka-config" }
 monza-config = { path = "networks/monza/monza-config" }
 # util
+flocks = { path = "util/flocks" }
```
0xmovses, line 75: Can we just names that are more self descriptive, `file-locks` is immediedtly understandable `flocks`, one has to remember what it means. To a new developer they would have to check or look up a glossary.

Cargo.toml
```diff
@@ -72,6 +72,7 @@ commander = { path = "util/commander" }
...
 godfig = { path = "util/godfig" }
```
0xmovses, line 76: Same with `godfig` its not immedietly apparent to the reader what this means. We could simply call it `system-config` or `dynamic-config`.

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+    #[error("Lock is not available")]
+    LockNotAvailable,
+    #[error("File error: {0}")]
+    FileError(#[from] std::io::Error),
+}
+
+/// A file-based read-write lock.
+/// This only mutually excludes processes trying to violate the lock, not the same process--which is not considered contention.
+/// If you want to prevent contention within the same process, you should wrap this in your preferred synchronization primitive.
+pub struct FrwLock<T: AsFd> {
```
0xmovses, line 25: Let's rename to: FileRwLock

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+pub enum FrwLockError {
+    #[error("Lock is not available")]
+    LockNotAvailable,
+    #[error("File error: {0}")]
+    FileError(#[from] std::io::Error),
+}
+
+/// A file-based read-write lock.
+/// This only mutually excludes processes trying to violate the lock, not the same process--which is not considered contention.
+/// If you want to prevent contention within the same process, you should wrap this in your preferred synchronization primitive.
```
0xmovses, line 24: This comment doesn't read very well. It's ok to add more lines to explain this point clearly.

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+
+impl<T: AsFd> FrwLock<T> {
+    pub fn new(file: T) -> Self {
+        Self {
+            cell: UnsafeCell::new(file),
+        }
+    }
+
+    pub(crate) fn try_write(&self) -> Result<FrwLockWriteGuard<T>, FrwLockError> {
+        let file = unsafe { &*self.cell.get() };
```
0xmovses, line 37: The raw pointer obtained from `self.cell.get()` must be valid. Maybe this is the only way, but it seems unnecessary to me to be using `unsafe` blocks for a config. 

I didn't realise that with this design we would need to actually read and write to files themselves. Just some state in memory (the dynamic KV part). Would this be possible?

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+        match flock(file, FlockOperation::NonBlockingLockExclusive) {
```
0xmovses, line 38: `match file_lock`

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+// FrwLock<T>.
+unsafe impl<T> Send for FrwLock<T> where T: AsFd + Sized + Send {}
+unsafe impl<T> Sync for FrwLock<T> where T: AsFd + Sized + Send + Sync {}
+// NB: These impls need to be explicit since we're storing a raw pointer.
+// Safety: Stores a raw pointer to `T`, so if `T` is `Sync`, the lock guard over
+// `T` is `Send`.
+unsafe impl<T> Send for FrwLockReadGuard<T> where T: AsFd + Sized + Sync {}
+unsafe impl<T> Sync for FrwLockReadGuard<T> where T: AsFd + Sized + Send + Sync {}
+unsafe impl<T> Send for FrwLockWriteGuard<T> where T: AsFd + Sized + Sync {}
+unsafe impl<T> Sync for FrwLockWriteGuard<T> where T: AsFd + Sized + Send + Sync {}
```
0xmovses, line 101: I understand its fine to have these, as long as these invariants hold up. But I still think this is too heavy for the problem we are trying to solve, that is, some kind of dynamic state where different services can read or write vals. Why does it have to be a file and not some other "place" in memory?

0xmovses: We shouldn't need to be handling thread safety and unsafe impls for simple config initialization. Just seems to much to me.

util/flocks/src/tfrwlock/mod.rs
```diff
@@ -0,0 +1,56 @@
...
+            FrwLockError::LockNotAvailable => TfrwLockError::LockNotAvailable,
+            FrwLockError::FileError(e) => TfrwLockError::FileError(e),
+        }
+    }
+}
+
+/// A file-based read-write lock.
+/// This only mutually excludes processes trying to violate the lock, not the same process--which is not considered contention.
+/// If you want to prevent contention within the same process, you should wrap this in your preferred synchronization primitive.
+pub struct TfrwLock<T: AsFd> {
```
0xmovses, line 33: What todes the T stand for here? In the name `Tfrw`

util/flocks/src/tfrwlock/read_guard.rs
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
+
+impl<'a, T: AsFd> Deref for TfrwLockReadGuard<'a, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        &*self.inner
+    }
+}
```
0xmovses, line 23: We shouldn't be needing to roll our own read write guards. Is there no crate that gives us a file based `RwLock` ? I mean even a lightweight DB like reddis would be easier. We get everything out of the box with that.

0xmovses (commented):
Why couldn't we use `redis` instead of rolling our own state manager (and all the assoiciated read write thread safety issues that come along with that?

My main worry with this is that we are creating more tech-debt when we might not need to. There might be a simpler way to solve our cross service config / init situation. And committing to maintaining this code might take us away from more important, rollup focused things. 

https://docs.rs/redis/latest/redis/

0xmovses:
Just to add further point here, I think the overall design here _is_ right. I think we can just simplify the implementation by using some KV state on a seperate process like reddis and get ourselves out of a world of pain potentially down the line. 

The code here was interesting to read, but the unsafe impl blocks seems a bit hacky for production! <3

Cargo.toml
l-monninger: `flocks` is the Unix term though. But, I suppose this is more more general. If @mzabaluev votes your way, then I'll change it.

Cargo.toml
l-monninger: `godfig` I think is branded. Right, @andyjsbell ?

l-monninger: Also, I generally subscribe to the doctrine of using novel names when creating opinionated crates and similar.

util/flocks/src/frwlock/mod.rs
l-monninger: This is following the same convention as `RwLock` (ReadWriteLock).

util/flocks/src/frwlock/mod.rs
l-monninger: You still need unsafe blocks for `UnsafeCell`. 

We need to persist, so that other processes can access and update.

util/flocks/src/frwlock/mod.rs
l-monninger: I didn't name this one flock. It comes from `rustix` which is wrapping Unix APIs: https://linux.die.net/man/2/flock

util/flocks/src/frwlock/mod.rs
l-monninger: Process A starts writing to the config.

Process B attempts to read whilst Process A is writing. 

Process B reads an invalid config. Process B crashes. 

Process A finishes writing.

util/flocks/src/tfrwlock/mod.rs
l-monninger: Thread

util/flocks/src/tfrwlock/read_guard.rs
l-monninger: On Reddit, some had claimed this took over: https://github.com/alfiedotwtf/file-lock. But, it doesn't seem well maintained. Might as well eat our own toast. 

With Redis, you would then need to run the Redis service.

0xmovses: Yes I meant Redis! How about using the Redis KV DB for a store?

util/flocks/src/frwlock/mod.rs
0xmovses: Line breaks!

util/flocks/src/tfrwlock/read_guard.rs
0xmovses: Also just found this, a transactional key-value store, pure rust implementation https://github.com/khonsulabs/nebari

util/flocks/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+rust-version = { workspace = true }
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+serde = { workspace = true, features = ["derive"] }
+anyhow = { workspace = true }
+tokio = { workspace = true }
+futures = { workspace = true }
+rustix = { workspace = true, features = ["fs"]}
```
mzabaluev, line 19: We're walking into Unix-specific territory here. For Windows, you'd have to use a different Win32 API. Hopefully this is intended.

util/flocks/src/frwlock/mod.rs
0xmovses: I wasn't aware of the term, in that case I don't mind, if we go with this implementation.

musitdev (commented):
I have mostly 2 main general remarks.
First, why all the API is async. Async don't add much in file processing. Flock is not async. The issue with async is that it force everything to be async. We won't be able to use this API in a non async program. Related to it, I don't think that Frwlocks should be sync and even send. We don't need to pass Godfish between thread, it's only used to read a config and to protect the access from other process. Without async and sync trait, it will simplify the whole code and API.
The second is about the marker trait contract for `Godfig<Contract, Backend>` that is a good idea, but It don't add constrain on config key use. In `V1::manager,` the config is declared as follows: `godfig: Godfig<Config, ConfigFile>,` where Config is the one of m1_da, but any key can be used to retrieve it. What would be interesting is to use the Contract to add constrain on the config key use to get and set.
Related to this question, I don't fully understand why you use a Vec<String> to set and get key. I thought it was to search in the tree of keys, but this Vec is mostly to get the config of a manager that will be mostly at the root of the tree.
So perhaps we can define a sort of config entry that is use as a contract and the parameter for get/set the manager config.

l-monninger:
> The code here was interesting to read, but the unsafe impl blocks seems a bit hacky for production, atleast for us (and future devs) to maintain ourselves! <3

Then unsafe blocks is intended to be using the same pattern as `tokio::sync`: https://github.com/tokio-rs/tokio/tree/master/tokio/src/sync. Then, the synchronization is up to the usage of  Unix `flock`--assuming there weren't mistakes in the above. 

As I mention in the PR, we are missing synchronization within the thread, but where the wrapper around the file descriptor is stored at a different address. 

Another alternative would be to try and have a `thing.lock` file-based lock and essentially implement a file-based semaphore. However, filesystems don't have good guarantees around `write()` and `read()`, so you'd likely to still have race conditions.

util/flocks/src/frwlock/mod.rs
```diff
@@ -0,0 +1,200 @@
...
+/// This only mutually excludes processes trying to violate the lock, not the same process--which is not considered contention.
+/// If you want to prevent contention within the same process, you should wrap this in your preferred synchronization primitive.
+pub struct FrwLock<T: AsFd> {
+    cell: UnsafeCell<T>,
+}
+
+impl<T: AsFd> FrwLock<T> {
+    pub fn new(file: T) -> Self {
+        Self {
+            cell: UnsafeCell::new(file),
```
mzabaluev, line 32: There should be comments explaining why do we have to have the unsafety and what are the lifetime guarantees for the contained object.

util/flocks/src/frwlock/mod.rs
mzabaluev, line 101: You don't need the `Sized` bound for generic parameters, it's implied.

util/flocks/src/frwlock/read_guard.rs
```diff
@@ -0,0 +1,25 @@
+use std::ops::Deref;
+use rustix::{
+    fs::{flock, FlockOperation},
+    fd::AsFd
+};
+
+pub struct FrwLockReadGuard<T: AsFd> {
+    pub(crate) data: *const T
+}
```
mzabaluev, line 9: What's the guarantee that the raw pointer will stay valid for the lifetime of the guard object?

util/flocks/src/frwlock/write_guard.rs
```diff
@@ -0,0 +1,31 @@
+use std::ops::{Deref, DerefMut};
+use rustix::{
+    fs::{flock, FlockOperation},
+    fd::AsFd
+};
+
+pub struct FrwLockWriteGuard<T: AsFd> {
+    pub(crate) data: *mut T
+}
```
mzabaluev, line 9: Also here, I think we're just hiding the unsafety creating a massive source of undefined behavior.

mzabaluev (changes requested):

l-monninger:
> First, why all the API is async. Async don't add much in file processing. Flock is not async. The issue with async is that it force everything to be async. We won't be able to use this API in a non async program. Related to it, I don't think that Frwlocks should be sync and even send. 

It's async so that we can use async polling against the Posix file lock bytes. IMO, we ideally feature flag `async` and `sync` APIs. I just started with async because that made sense within the existing code. 

> We don't need to pass Godfish between thread, it's only used to read a config and to protect the access from other process.

In the manager pattern where we keep checking the config, we do pass it between threads--unless we want to clone everywhere which would potentially lead to two versions of state. 

> The second is about the marker trait contract for Godfig<Contract, Backend> that is a good idea, but It don't add constrain on config key use.

Generally, the current implementation uses `Into<Vec<String>>` for a key, so that you can have a clear key "path"--i.e., without having to work with delimiters. `Godfig<Contract, Backend>` does away with this because you can of course convert `Into<Vec<String>>` into `Vec<String>` and just store that. 

I had entertained using other sorts of keys, but it complicates the model. 

> So perhaps we can define a sort of config entry that is use as a contract and the parameter for get/set the manager config.

I'd thought about doing some indirection on the keys, so that you would have properties like this. But, it seems overcomplicated.

@musitdev

util/flocks/src/frwlock/read_guard.rs
l-monninger: Very important catch. I had rewritten to `OwnedGuards` because the lifetimes were becoming quite difficult to manage: https://github.com/tokio-rs/tokio/blob/06582776a564c88a7b4c6f9e3d7c0ebd0ef3f34b/tokio/src/sync/rwlock.rs#L905. But, I need to change the receivers to Arc<Self> and probably rename these. 


Otherwise, yes this needs...
```
pub(crate) data: *mut T,
pub(crate) marker: PhantomData<&'a mut T>
```

util/flocks/src/frwlock/write_guard.rs
l-monninger: Same as above.

l-monninger:
I think I can make you guys happier if just implement `Trfw` and instead of using `UnsafeCell`, I actually just compose with an `RwLock` and build guards from tokio guards.

util/flocks/src/frwlock/read_guard.rs
mzabaluev: The normal Rust way to avoid the unsafety for temporary RAII objects is to use lifetimes, and `rustix` provides some additional help:

```rust
use rustix::fd::BorrowedFd;

pub struct FrwLockReadGuard<T, 'a> {
    fd: BorrowedFd<'a>
}
```

`BorrowedFd` satisfies `Send + Sync`. I think the lifetime bound would let you code async blocks safely without allowing any locks to "escape".

util/godfig/src/lib.rs
```diff
@@ -1,3 +1,7 @@
+pub mod backend;
+pub mod godfig;
+pub use godfig::*;
```
mzabaluev, line 3: Perhaps make `godfig` a private mod if everything in it can be imported from the crate root.

util/godfig/src/backend/mod.rs
```diff
@@ -1,57 +1,56 @@
...
+impl From<TfrwLockError> for GodfigBackendError {
+    fn from(error: TfrwLockError) -> Self {
+        GodfigBackendError::BackendError(error.into())
     }
+}
 
-    /// Tries to set a key-value pair.
-    async fn try_set<K, T>(&self, key: K, value: T) -> Result<(), anyhow::Error>
+pub trait BackendOperations {
+    async fn try_get<K, T>(&self, key: K) -> Result<Option<T>, GodfigBackendError>
```
mzabaluev, line 30:
```suggestion
    fn try_get<K, T>(&self, key: K) -> impl Future<Output = Result<Option<T>, GodfigBackendError>> + Send
```

May want to do what the compiler warnings suggest up front. Note that the implementations don't need this desugaring.

util/flocks/src/frwlock/read_guard.rs
mzabaluev: A phantom marker is not enough, you have to guarantee the validity of the pointee object. An `Arc` over the file object would guarantee that. But I expect owned locks to be bug-prone in general.

l-monninger:
I also just remembered as discovery I made that let me to implement that weird half-step between an owned guard and one with proper lifetimes...

It's that if you try to hold a guard on an AsFd beyond its lifetime, when the FrwLock drops, it will panic with an invalid file descriptor error. So, you actually would very ungracefully get the same guarantees--which is kind of cool.
But, I should not have left that there.

0xPrimata:
Also makes me wonder how much tech debt is being created here. godfig resulting in 25 files changes, 1k lines for one project being configed seems worrisome.

util/flocks/src/frwlock/mod.rs
mzabaluev: I'd go for `FileRwLock`, too: "RW" (which is what "Rw" refers to in camel-case notation) is a common abbreviation for read/write especially in conjunction with "Lock", but "F" is not related and not sufficiently clear.

util/flocks/Cargo.toml
radupopa369: I am fine with not running blockchain on windows :)

l-monninger:
This has already made its way in under a variety of other PRs.

PR #139, opened 2024-06-20
ENG 601 update setup to start Anvil and generate MCR config.

musitdev:
# Summary

Start all necessary services / process to run settlement client in local.

Open for review before rebase of main branch.

# Changelog

Update Settlement Local Setup process to start Anvil and publish the MCR smart contract
Update Settlement config accordingly with anvil startup and SC deployement.
Add Suzuka Node gracefull shutdown to stop anvil process if any and call all destructor.

# Testing
Before starting Suzuka node the config.toml file in .movement must be deleted otherwise the anvil process is not started because a private key is already defined (defined by the last execution).

 To run use the Suzuka client tests :
`CELESTIA_APP_PATH=/usr/local/bin/celestia-appd CELESTIA_LOG_LEVEL=FATAL MOVEMENT_PREBUILT=false nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just suzuka-full-node native test.build.setup.local --keep-tui"`

# Outstanding issues

Settlement client tests hasn't been updated and still use bash script to start anvil.

The config.toml file is not deleted at the end of execution. It will be resolved with the next config management process.


networks/suzuka/suzuka-full-node/src/main.rs
```diff
@@ -14,16 +18,52 @@ async fn main() -> Result<(), anyhow::Error> {
...
+	// End the program to shutdown gracefully when a signal is received.
+	let (stop_tx, mut stop_rx) = watch::channel(());
+	tokio::spawn({
+		let mut sigterm = signal(SignalKind::terminate()).context("Can't register to SIGTERM.")?;
+		let mut sigint = signal(SignalKind::interrupt()).context("Can't register to SIGKILL.")?;
+		async move {
+			loop {
+				select! {
+					_ = sigterm.recv() => println!("Receive SIGTERM"),
+					_ = sigint.recv() => println!("Receive SIGTERM"),
```
mzabaluev, line 32: These should not be unconditional printouts, use `tracing::debug!`?

networks/suzuka/suzuka-full-node/src/main.rs
```diff
@@ -14,16 +18,52 @@ async fn main() -> Result<(), anyhow::Error> {
...
+					tracing::warn!("Can't update stop watch channel because :{err}");
+					return Err::<(), anyhow::Error>(anyhow::anyhow!(err));
+				}
+			}
+		}
+	});
 
-	tokio::spawn(background_task);
+	//Start suzuka node process
+	let (gb_jh, run_jh) = {
```
mzabaluev, line 43: The block seems unnecessary, but this is just stylistics.

protocol-units/da/m1/setup/src/local.rs
```diff
@@ -1,16 +1,15 @@
+use crate::M1DaLightNodeSetupOperations;
+use celestia_types::nmt::Namespace;
 use commander::run_command;
 use dot_movement::DotMovement;
+use rand::Rng;
```
mzabaluev, line 5: Nitpick: I like to group use statements by dependencies in the same namespace, external crates, and `std`/`core`.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -24,13 +20,17 @@ use alloy_signer_wallet::LocalWallet;
...
 use alloy_transport_ws::WsConnect;
-
 use anyhow::Context;
+use mcr_settlement_config::Config;
+use movement_types::BlockCommitment;
+use movement_types::{Commitment, Id};
+use serde_json::Value as JsonValue;
+use std::array::TryFromSliceError;
+use std::fs;
+use std::path::Path;
```
mzabaluev, line 30: And this is why I don't like the ungrouped imports, everything is mashed together. Ah well.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -65,6 +82,7 @@ pub struct Client<P> {
 	send_tx_error_rules: Vec<Box<dyn VerifyRule>>,
 	gas_limit: u128,
 	num_tx_send_retries: u32,
+	kill_anvil_process: Option<AnvilKillAtDrop>,
```
mzabaluev, line 85: I don't think this belongs in the client.

In the production use case, we don't run Anvil, so let's not make this a wart on `Client`, but compose this helper with `Client` somewhere in test infrastructure?

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -252,6 +275,45 @@ where
 	}
 }
 
+pub struct AnvilAddressEntry {
+	pub address: String,
+	pub private_key: String,
+}
+
+/// Read the Anvil config file keys and return all address/private key.
+pub fn read_anvil_json_file_addresses<P: AsRef<Path>>(
```
mzabaluev, line 284: @l-monninger did not document the separation of concerns when adding the *-setup infrastructure crates, but as I understand it, this setup functionality belongs precisely there, not in the client crate. The integration tests would then consume the configuration created by the e2e setup.

util/commander/src/lib.rs
```diff
@@ -76,3 +77,54 @@ pub async fn run_command(command: &str, args: &[&str]) -> Result<String> {
 
     Ok(stdout_output)
 }
+
+/// Runs a command, piping its output to stdout and stderr, and returns the stdout output if successful.
+pub async fn spawn_command(command: String, args: Vec<String>) -> Result<(Option<u32>, JoinHandle<Result<String, anyhow::Error>>)> {
+
+    // print command out with args joined by space
+    println!("spawn command: {} {}", command, args.join(" "));
```
mzabaluev, line 85: Use `tracing`?

protocol-units/settlement/mcr/setup/src/local.rs
```diff
@@ -37,26 +36,140 @@ impl Default for Local {
...
-			info!("setting up MCR Ethereum client");
+		tracing::info!("Init Settlement local conf");
 
+		async move {
 			if config.rpc_url.is_none() {
 				config.rpc_url = Some(format!("http://localhost:{}", self.eth_rpc_port));
 			}
 			if config.ws_url.is_none() {
-				config.ws_url = Some(format!("http://localhost:{}", self.eth_ws_port));
+				config.ws_url = Some(format!("ws://localhost:{}", self.eth_ws_port));
```
mzabaluev, line 54: 👍

protocol-units/settlement/mcr/client/src/eth/mod.rs
```diff
@@ -1,2 +1,3 @@
 pub mod mcr;
 // pub mod staking;
+pub mod utils;
```
mzabaluev, line 3: This is dead code and does not refer to anything yet.

protocol-units/settlement/mcr/config/src/lib.rs
```diff
@@ -22,16 +22,17 @@ pub struct Config {
 	#[serde(default = "default_mcr_contract_address")]
 	pub mcr_contract_address: String,
 	#[serde(default = "default_gas_limit")]
-	pub gas_limit: u128,
+	pub gas_limit: u64,
 	#[serde(default = "default_tx_send_retries")]
 	pub tx_send_retries: u32,
+	pub anvil_process_pid: Option<u32>,
```
mzabaluev, line 28: Can we make management of infrastructure processes part of the orchestration layer, without adding dynamic warts to the config?

If this field is present in the config on file, but the process has already finished and possibly another one has taken its PID, this can get messy.

mzabaluev (changes requested):
Thanks for making it work.
I only have an objection about entangling the settlement client and the config with the vagaries of running an Anvil client process.

protocol-units/settlement/mcr/client/src/eth_client.rs
```diff
@@ -57,6 +57,23 @@ sol!(
...
+
+impl Drop for AnvilKillAtDrop {
+	fn drop(&mut self) {
+		tracing::info!("Killing Anvil process");
+		if let Err(err) = std::process::Command::new("kill").args(&[&self.pid.to_string()]).spawn()
+		{
+			tracing::info!("warn, an error occurs during Anvil process kill : {err}");
+		}
+	}
+}
```
mzabaluev, line 75: 1. This a very happy-path solution for minding a child process. Let's leave it to process-compose or whatever other systemic solution we come up with.
2. There are in-process ways to kill a process, no need to run a command.
3. This is also not portable to non-Unix systems.

networks/suzuka/suzuka-full-node/src/main.rs
l-monninger: Since you won't be worry about merging on main until after this sprint, maybe we do want to spend a bit more time on style though.

protocol-units/settlement/mcr/config/src/lib.rs
mzabaluev: Filed this concern as #143 to unblock this PR.

networks/suzuka/suzuka-full-node/src/main.rs
musitdev: Yes removed, it was for test.

networks/suzuka/suzuka-full-node/src/main.rs
musitdev: yes I remove it.

protocol-units/da/m1/setup/src/local.rs
musitdev: I'm used to remove all blank line in use statement because rustfmt sort them automatically if there are all in the same block. Sorted use is very convenient when you search for one. But I can change and keep your separate block.

protocol-units/settlement/mcr/client/src/eth/mod.rs
musitdev: yes remove it.

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: I don't agree here. They are sorted and you can find one easily. If you separate them like in this example. Alloy is separated. Automatic import when they are missing (my IDE add them automatically) will be added to the alloy section which result a real mash up where you have two sorted collections of import and need to search in both. And if you add more separate collection, in the end everything is mashed. Otherwise you've to very each time you import that the new import is in the right place that remove all the interest of automatic import.

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: I'm agree. I do it like that because, in the std lib I didn't find a way to do it and I didn't want to import a new crate because it's a temporary solution until we settlement the config and process management. So it's working for mac and linux that are the current OS that we target that why I use this very easy solution.

protocol-units/settlement/mcr/client/src/eth_client.rs
musitdev: Yes it's a temporary solution until process management is settlement. Currently I've no other way to indicate a process from one process to the other.

protocol-units/settlement/mcr/config/src/lib.rs
musitdev: It's for local testing in a managed env. If anvil stop it won't be restarted unless the user explicitly do it.

util/commander/src/lib.rs
musitdev: Yes good catch.

protocol-units/da/m1/setup/src/local.rs
mzabaluev: Yes, sorting within groups is fine, but I like to keep the imports grouped by provenance as well.

protocol-units/settlement/mcr/client/src/eth_client.rs
mzabaluev: OK, this seems less of a headache with automatically added imports.

protocol-units/settlement/mcr/client/src/eth_client.rs
mzabaluev: To be further addressed in #105.

mzabaluev (approved):
Looks good to go, all concerns have been discussed and filed away to be fixed later if necessary.

networks/suzuka/suzuka-config/src/lib.rs
l-monninger, line 1: This will always be used within Suzuka, so I think not flattening as you've done is fine.

protocol-units/settlement/mcr/client/src/eth_client.rs
l-monninger: I also prefer to group imports a bit.

protocol-units/settlement/mcr/client/src/eth_client.rs
l-monninger, line 1: There are still typos here. Please install the Grammarly extension and check the blue squiggles.

protocol-units/settlement/mcr/setup/src/local.rs
l-monninger, line 1: Looks reasonable enough.

networks/suzuka/setup/src/local.rs
```diff
@@ -33,7 +28,12 @@ impl Local {
 
 		// Update the config with the new m1_da_light_node_config
 		config.m1_da_light_node = new_m1_da_light_node_config;
- 
+
+		tracing::info!("Running mcr_settlement_setup");
+		let mcr_settlement_config: mcr_settlement_config::Config = config.mcr.clone();
+		config.mcr =
+			self.mcr_settlement_strategy.setup(&dot_movement, mcr_settlement_config).await?;
```
l-monninger, line 35: Yep, looks good.

util/commander/src/lib.rs
```diff
@@ -76,3 +77,54 @@ pub async fn run_command(command: &str, args: &[&str]) -> Result<String> {
 
     Ok(stdout_output)
 }
+
+/// Runs a command, piping its output to stdout and stderr, and returns the stdout output if successful.
+pub async fn spawn_command(command: String, args: Vec<String>) -> Result<(Option<u32>, JoinHandle<Result<String, anyhow::Error>>)> {
```
l-monninger, line 82: Yeah, reasonable way to do this.

l-monninger (changes requested):
Please fix the typos, then we can merge.

l-monninger (approved):

mzabaluev:
What causes the setup failure here: https://github.com/movementlabsxyz/movement/actions/runs/9775222670/job/26986389472?pr=139

PR #140, opened 2024-06-20
Graceful Restarts for Aptos Execution

l-monninger:
# Summary

Adds graceful restarts for Aptos execution, i.e., can start from the current DB. 

We still create a new DA network. I will try to address this in a forthcoming PR, but it is more involved. However, it doesn't matter for the current deployment which has one node because heights are assigned by the executors themselves and settled on. 

This would matter if nodes were started at different times and on different networks because then the bootstrapping would potentially miss transactions from one network (ofc). 

Excluding fixes to the DA to keep this PR small. 

# Changelog

1. Very small change to `Executor::maybe_bootstrap_empty_db` in `maptos-opt-executor`. 

# Testing

1. Ran `suzuka-full-node` e2e back-to-back. Can improve upon tests by checking for state from before, but this is also a bit more complicated and will be reserved for a future PR. 

# Outstanding issues

1. DA graceful restarts.
2. Restart testing for state.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -46,7 +46,10 @@ jobs:
 
     - name: Run M1 DA Light Node tests in nix environment
       # adjust the log level while debugging
-      run: CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"  
+      run: |
+        CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"  
+        # see if restarts are graceful
+        CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"  
 
   m1-da-light-node:
     if: false # this is effectively tested by the above
--- a/protocol-units/execution/opt-executor/src/executor/initialization.rs
+++ b/protocol-units/execution/opt-executor/src/executor/initialization.rs
@@ -70,30 +70,45 @@ impl Executor {
 		(genesis, test_validators)
 	}
 
-	pub fn bootstrap_empty_db(
+	pub fn maybe_bootstrap_empty_db(
 		db_dir: &PathBuf,
 		chain_id: ChainId,
 		public_key: &Ed25519PublicKey,
+		maptos_config: &Config,
 	) -> Result<(DbReaderWriter, ValidatorSigner), anyhow::Error> {
+
+		let db_rw = DbReaderWriter::new(AptosDB::new_for_test(db_dir));
 		let (genesis, validators) =
 			Self::genesis_change_set_and_validators(chain_id, Some(1), public_key);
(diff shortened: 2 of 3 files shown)


protocol-units/execution/opt-executor/src/executor/initialization.rs
```diff
@@ -70,30 +70,45 @@ impl Executor {
 		(genesis, test_validators)
 	}
 
-	pub fn bootstrap_empty_db(
+	pub fn maybe_bootstrap_empty_db(
```
mzabaluev, line 73: It would be good to document what "maybe" means here. Or clarify with naming `bootstrap_db_if_empty`?

Generally, fallible functions do not need to be prefixed with `try_` or `maybe_` unless it's needed to disambiguate from the infallible variant present alongside.

l-monninger: I'll commit once this runner stops, so that I don't bog it down.

.github/workflows/checks.yml
```diff
@@ -46,7 +46,10 @@ jobs:
 
     - name: Run M1 DA Light Node tests in nix environment
       # adjust the log level while debugging
-      run: CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"  
+      run: |
+        CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"  
+        # see if restarts are graceful
+        CELESTIA_LOG_LEVEL=FATAL nix develop --command bash  -c "just suzuka-full-node native build.setup.test.local -t=false"
```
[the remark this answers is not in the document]
l-monninger: Yeah, but how else would you test it?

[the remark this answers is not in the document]
l-monninger: It shouldn't be twice the "running" time though because currently a big portion of the "running" time is actually build time. So, we would skip all of that and just start up the network.

protocol-units/execution/opt-executor/src/executor/initialization.rs
```diff
@@ -70,30 +70,46 @@ impl Executor {
 		(genesis, test_validators)
 	}
 
-	pub fn bootstrap_empty_db(
+	/// Bootstrap a database with a genesis transaction if it is empty.
+	pub fn maybe_bootstrap_empty_db(
 		db_dir: &PathBuf,
 		chain_id: ChainId,
 		public_key: &Ed25519PublicKey,
+		maptos_config: &Config,
```
musitdev, line 78: Why you add `maptos_config`, you don't use it in the function.

musitdev (approved):

protocol-units/execution/opt-executor/src/executor/initialization.rs
l-monninger: Good catch. I had that in there before I simplified and must have forgotten to remove.

l-monninger: Also helps me catch that my Rust Analyzer is not doing much haha.

PR #146, opened 2024-06-24
MCR batch timeout

mzabaluev:
# Summary

Add the batch timeout setting to the configuration, make the mcr-settlement-manager honor it by flushing the accumulated commitments as a batch on timeout.

# Changelog

* `batch_timeout` parameter added to the MCR settlement configuration.

# Testing

Add a test exercising timeout behavior with a repetition.


musitdev (approved):
The future pending/timout seems to me a little overcomplicated. It works but with just a timeout that send the current batch when fired and been reset when the batch is sent when max_height is reach.
Otherwise, the pattern is interesting.
I think by just using a timeout, you can avoid the copy of the code that send the batch (mem::replace and client post) and put everything in timeout_ready by triggering the timeout when max_height is reach.
I let you decide if it's interesting or not.

protocol-units/settlement/mcr/manager/src/manager.rs
```diff
@@ -53,6 +60,7 @@ fn process_commitments<C: McrSettlementClientOperations + Send + 'static>(
 		let mut ahead_of_settlement = false;
```
0xmovses, line 60: When is this ever true.

0xmovses (approved):
Interesting use of `BTreeMap` here for `commitments_to_settle`. The test checks out. 
Out of curiosity, is this a pattern you invented, or something you've used or seen in the wild?

:shipit:

protocol-units/settlement/mcr/manager/src/manager.rs
mzabaluev: It's set to true when we get a commitment with height above what the client says it can accept.

mzabaluev:
> I think by just using a timeout, you can avoid the copy of the code that send the batch (mem::replace and client post) and put everything in timeout_ready by triggering the timeout when max_height is reach.

I thought of that, but in this place, we already have one new commitment that we can't send yet, so we need to flush the batch and start a new one.

I tried a helper async function to avoid duplicating the code, but borrowing the client instance forces an extra `Sync` bound on the generic parameter because the future of the helper fn, which internalizes the client reference, needs to be `Send` (I don't fully understand why the current unrolled code does not require `Sync`, though). So it's a choice between a public API wart and a little ugliness in internal code.

mzabaluev:
> Interesting use of `BTreeMap` here for `commitments_to_settle`. The test checks out. Out of curiosity, is this a pattern you invented, or something you've used or seen in the wild?

It was the quickest thing I thought about when mapping a potentially sparse (?) sequence of block heights. It's sorted so we can quickly check the largest key for that remaining corner case that still has the `todo!` on it.

PR #148, opened 2024-06-25
Flocks w/ Better Lifetimes

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `misc`

Does away with anyone owned guards in `flocks`. 

# Changelog

1. Updates `FrwLock` to use lifetimes, ensuring no outliving guards without using owned guard pattern.
2. Updates `TfrwLock` to use `tokio::RwLock` internally and likewise simplifies lifetimes.

# Testing

1. `cargo test -p flocks`
2. `cargo test -p godfig`

# Outstanding issues

1. Still can acquire locks on the same file descriptor in the same process if not using a single resource for said access.


util/flocks/src/frwlock/read_guard.rs
```diff
@@ -3,20 +3,22 @@ use rustix::{
     fs::{flock, FlockOperation},
     fd::AsFd
 };
+use std::marker::PhantomData;
 
-pub struct FrwLockReadGuard<T: AsFd> {
-    pub(crate) data: *const T
+pub struct FrwLockReadGuard<'a, T: AsFd> {
+    pub(crate) data: *const T,
+    pub(crate) _marker: PhantomData<&'a T>
```
mzabaluev, line 10: I don't see the point of making it this kludgy. If you can guarantee the `'a` lifetime for the object as you claim you do on the type system level, why not just make the `&'a T` borrow the inner value of the guard? If this is for relaxing the bounds on the standard-defined `Sync`/`Send` for `&'a T` (as the custom impls above suggest), the need to do it should be carefully scrutinized, because we may be violating soundness of the pointer access.

The `T` object must own the resource behind the file descriptor. If it's allowed to drop before the lock object, the file descriptor may be reclaimed by another object, e.g. a different file under advisory locking in our system, so dropping the lock guard will call `flock` on that.

util/flocks/src/tfrwlock/mod.rs
```diff
@@ -18,39 +20,176 @@ pub enum TfrwLockError {
...
+    }
+
+    pub async fn write(&self) -> Result<TfrwLockWriteGuard<'_, T>, TfrwLockError> {
+        loop {
+            match self.try_write() {
+                Ok(guard) => return Ok(guard),
+                Err(TfrwLockError::LockNotAvailable) => {
+                    // Yield control to other tasks
+                    yield_now().await;
+                },
```
mzabaluev, line 93: This will busy-loop when the file is locked by another process. To play nice with the OS, you'd need to use a blocking `flock`. To make it properly async-friendly, this lacks a way to wake up the task when locking status on the file changes. I'm not aware of any polling API that could be used for this purpose, though.

util/flocks/src/frwlock/read_guard.rs
mzabaluev: As already suggested in https://github.com/movementlabsxyz/movement/pull/137#discussion_r1652463085,
`BorrowedFd` might be the solution we need here.

l-monninger: `BorrowedFd` sounds like it would make sense.

util/flocks/src/tfrwlock/mod.rs
l-monninger: We could always sleep a little and just assert this is a low-performance lock. Some number of microseconds could be reasonable as file writes should be on that order. 

I thought about using `inotify`, for which there are async stream implementations. But, I figured that might wind up being be heavier and it requires paths.

POSIX `aio` isn't well-supported ofc.

l-monninger: > inotify may help if you use the file presence as the lock or "weak-lock" (i.e. the lock holder must unlock and remove the file). But this is less flexible and adds another failure mode: a process may die and leave the file, so inotify will never trigger.

Placing it on `xattrs` would be a bit better wouldn't it? Either way, yes, you would still need to poll as well. 

> This may be something that can only be solved by spawn_blocking.
You mean just for the thread pools? You don't think it's overkill. `flock` is blocking, but it should be a short operation, no? It just checks some bytes in an inode, no?

mzabaluev: If `flock` is called in the blocking mode, it's blocked while the lock is held. That's the dilemma: if it's blocking and the call waits on the lock in async context, the whole thread is blocked in the work-stealing runtime. If it's non-blocking, there is no nice way to schedule a wakeup for the next good time to poll for the lock, so you either busy-loop or sleep; each of these strategies have their own disadvantages.

`spawn_blocking` is not that expensive, but it might be pending longer and less predictably because the thread pool may be fully booked.

l-monninger: I actually wasn't even considering not using `LOCK_NB`.

When the lock actually gets acquired, that's always a blocking operation, but a very brief one, right? I mean it has to be atomic at least. `LOCK_NB` is just to do non-blocking while checking the lock, right?

My inclination would be to just do `LOCK_NB` with polling and accept the small overhead for the block when the lock is actually acquired. But, I guess I don't really mind doing `spawn_blocking` if you think that's the right way to go. Just want a little more info here.

util/flocks/src/frwlock/mod.rs
```diff
@@ -95,18 +97,18 @@ unsafe impl<T> Sync for FrwLock<T> where T: AsFd + Sized + Send + Sync {}
 // NB: These impls need to be explicit since we're storing a raw pointer.
 // Safety: Stores a raw pointer to `T`, so if `T` is `Sync`, the lock guard over
 // `T` is `Send`.
-unsafe impl<T> Send for FrwLockReadGuard<T> where T: AsFd + Sized + Sync {}
-unsafe impl<T> Sync for FrwLockReadGuard<T> where T: AsFd + Sized + Send + Sync {}
-unsafe impl<T> Send for FrwLockWriteGuard<T> where T: AsFd + Sized + Sync {}
-unsafe impl<T> Sync for FrwLockWriteGuard<T> where T: AsFd + Sized + Send + Sync {}
+unsafe impl<T> Send for FrwLockReadGuard<'_, T> where T: AsFd + Sized + Sync {}
+unsafe impl<T> Sync for FrwLockReadGuard<'_, T> where T: AsFd + Sized + Send + Sync {}
+unsafe impl<T> Send for FrwLockWriteGuard<'_, T> where T: AsFd + Sized + Sync {}
```
mzabaluev, line 102: I believe `T` needs a `Send` bound as well.

util/flocks/src/lib.rs
```diff
@@ -1,2 +1,37 @@
...
+    use thiserror::Error;
+
+    #[derive(Debug, Error)]
+    pub enum AsyncFlockError {
+        #[error("Lock is not available")]
+        JoinError(#[from] tokio::task::JoinError),
+        #[error("File error: {0}")]
+        IOError(#[from] rustix::io::Errno),
+        #[error("Misc: {0}")]
+        MiscError(String),
```
mzabaluev, line 19: Style: I'd avoid repeating `Error` suffixes on variants if the enum already is named `Error`.

util/flocks/src/lib.rs
```diff
@@ -1,2 +1,37 @@
...
+        fs::{flock as sync_flock, FlockOperation}
+    };
+    use thiserror::Error;
+
+    #[derive(Debug, Error)]
+    pub enum AsyncFlockError {
+        #[error("Lock is not available")]
+        JoinError(#[from] tokio::task::JoinError),
+        #[error("File error: {0}")]
+        IOError(#[from] rustix::io::Errno),
```
mzabaluev, line 17:
```suggestion
        Io(#[from] rustix::io::Errno),
```

Pedantically, camel-case style only capitalizes the first letter in acronyms; IIRC std and tokio has similar variants.

util/flocks/src/lib.rs
```diff
@@ -1,2 +1,37 @@
...
+    use rustix::{
+        fd::AsFd,
+        fs::{flock as sync_flock, FlockOperation}
+    };
+    use thiserror::Error;
+
+    #[derive(Debug, Error)]
+    pub enum AsyncFlockError {
+        #[error("Lock is not available")]
+        JoinError(#[from] tokio::task::JoinError),
```
mzabaluev, line 15: Future-proofing: this might be leaking too much implementation detail. If we decide we don't need `spawn_blocking`, this variant will have to be changed. Maybe just an opaque `NotAvailable` variant with a message would do.

util/flocks/src/lib.rs
mzabaluev, line 17: This leaks `rustix` into public API.
As `Errno` is infallibly convertible into `std::io::Error`, perhaps it's better to use the standard type.

util/flocks/src/tfrwlock/mod.rs
```diff
@@ -5,52 +5,210 @@ pub use write_guard::TfrwLockWriteGuard;
...
-
+use rustix::{
+    fs::FlockOperation,
+    fd::AsFd,
+};
+use crate::asynchronous::flock;
+use crate::asynchronous::AsyncFlockError;
 
 #[derive(Debug, Error)]
 pub enum TfrwLockError {
```
mzabaluev, line 17: I don't like: 1) prefixes used where module paths can disambiguate instead; 2) cryptic prefixes.

Since the flocks API user is unlikely to mix the two kinds of locks in the same piece of code, I think it's better to give them the same name: `FileRwLock`, likewise for the ancillary guard and error types. And distinguish the two kinds by module paths: `sync::FileRwLock` for the sync version, `tokio::FileRwLock` for the async-friendly variant (to highlight that it requires the tokio runtime).

util/flocks/src/tfrwlock/mod.rs
mzabaluev: > When the lock actually gets acquired, that's always a blocking operation, but a very brief one, right?

Yes, uncontested operation is not of concern.

> My inclination would be to just do `LOCK_NB` with polling

If you can find an efficient way to wake up the task, that would be ideal for async. `xattrs` with `inotify`? This is not portable, adds a persistent failure mode if the xattr is not cleaned up by the holding process, and imposes some extra requirements on the file system (to think of it, `flock` may do as well, but that's in POSIX and is dependent only on the OS kernel support).

musitdev (approved):
I still think we should use a sync API for Flock/ConfigFile and if we need, we can easily provide an async version for ConfigFile.
It's not really so important, so I approve and you decide.

util/flocks/src/lib.rs
```diff
@@ -1,7 +1,6 @@
-pub mod frwlock;
 pub mod tfrwlock;
```
mzabaluev, line 1: This module could be merged with `tokio` as well.

mzabaluev (approved):
The changes look good, some minor concerns remain.

mzabaluev:
> I still think we should use a sync API for Flock/ConfigFile and if we need, we can easily provide an async version for ConfigFile. It's not really so important, so I approve and you decide.

Depends if we have genuinely synchronous code to use it in, so that asyncness is not forced by this functionality alone.

util/flocks/src/lib.rs
mzabaluev: Or maybe, make it private and re-export `FileReadWriteLock*` at the top level.

util/flocks/src/tfrwlock/mod.rs
```diff
@@ -1,56 +1,214 @@
 pub mod write_guard;
 pub mod read_guard;
```
mzabaluev, line 2: No need to expose these modules publicly.

util/flocks/src/tfrwlock/read_guard.rs
```diff
@@ -1,23 +1,28 @@
...
-use crate::frwlock::{FrwLock, FrwLockReadGuard};
+use rustix::{
+    fs::{flock, FlockOperation},
+    fd::AsFd
+};
 
-pub struct TfrwLockReadGuard<'a, T: AsFd> {
-    _outer: RwLockReadGuard<'a, FrwLock<T>>,
-    inner: FrwLockReadGuard<T>,
+pub struct FileRwLockReadGuard<'a, T: AsFd> {
```
mzabaluev, line 8: As said elsewhere, I generally avoid lengthy prefixed names in favor of terser names disambiguated by module structure or `as` aliases if necessary: `use flocks::tokio::rwlock::ReadGuard as FileReadGuard;`

PR #151, opened 2024-06-26
chore: update tests in the workspace to compile

mzabaluev:
# Summary
- **RFCs**: $\emptyset$.

Update the tests in workspace crates that had compilation failures
because they needed adaptation to the new config and setup APIs,
so that the tests build.

# Testing

`cargo check --all-targets`
`cargo check -p m1-da-light-node-verifier --tests --features integration-tests`

No effort to make the updated tests pass yet.
This change is only to resolve build failures in `cargo check --all-targets` that is used by IDEs to show build problems.

# Outstanding issues

Need to produce workable test setups and add test runs to the CI so that the crate tests are run. This should be a subject for a separate planned task.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -25,7 +25,7 @@ jobs:
       uses: DeterminateSystems/nix-installer-action@main
 
     - name: Run Cargo Check in nix environment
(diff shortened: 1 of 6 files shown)


l-monninger (approved):
LTGM

PR #153, opened 2024-06-27
ENG-633 implement bridge services

0xmovses:
# Summary
NB. There are still some missing pieces to the entire implementation and these have been noted in seperate issues listed in the feature branch https://github.com/movementlabsxyz/movement/pull/215#issue-2413213805

It's worth noting also that Config and how initialise the service may have to change. What's here is mainly for convinience with regard to config and init, these decisions were made so that I quickly e2e test my code, likely most of the config will be refactored for a more generalised and ergomonic purpose.

The eth side e2e test currently fails because event logging is not implemented in this PR. Initially I attempted to return the value from chain, but I realised this would violate the method sigs and that the returned `BridgeTransferId` should be captured via events, not logs. 

In either case the e2e test deploys the eth contract and calls the functions correctly until the complete (where later it will succeed). 

For e2e testing aptos, it was agreed between @cde8 and I that we would leave this out until #222 is implemented as this will likely effect how we e2e test. Some tooling around deploying the move modules to the network would be required and we thought it wise not to duplicate efforts. 

# Changelog
- Adds the crates `"protocol-units/bridge/chains/ethereum",` `"protocol-units/bridge/chains/movement",` and `"protocol-units/bridge/e2e",`
- Adds CLI logic for calling eth contract 
- Adds e2e test 
- Adds various utililty logic for both eth and movement sides.
  
# Testing
`cargo  run -p bridge e2e-tests` works up till complete step, as described above. 

# Outstanding issues
Described above


protocol-units/bridge/cli/src/eth.rs
```diff
@@ -0,0 +1,193 @@
...
+#[derive(Parser)]
+#[command(name = "eth-bridge-cli")]
+#[command(about = "CLI for interacting with the Ethereum bridge client", long_about = None)]
+struct EthCli {
+	#[command(subcommand)]
+	command: EthCommands,
+}
+
+#[derive(Subcommand)]
+pub enum EthCommands {
```
l-monninger, line 22: Nice

protocol-units/bridge/cli/src/movement.rs
```diff
@@ -0,0 +1,141 @@
...
+		Amount, BridgeTransferId, HashLock, HashLockPreImage, InitiatorAddress, RecipientAddress,
+		TimeLock,
+	},
+};
+use clap::{Parser, Subcommand};
+use movement_bridge::{utils::MovementAddress, MovementClient};
+
+#[derive(Parser)]
+#[command(name = "movement-cli")]
+#[command(about = "CLI for interacting with the Movement client", long_about = None)]
```
l-monninger, line 14: Should we perhaps make this read as "Movement Bridge client?"

protocol-units/bridge/contracts/src/WETH9.sol
```diff
@@ -0,0 +1,74 @@
...
+
+    function transferFrom(address src, address dst, uint wad)
+        public
+        returns (bool)
+    {
+        require(balanceOf[src] >= wad, "Insufficient balance");
+
+        if (src != msg.sender && allowance[src][msg.sender] != type(uint).max) {
+            require(allowance[src][msg.sender] >= wad, "Allowance exceeded");
+            allowance[src][msg.sender] -= wad;
```
l-monninger, line 63: I think you may want to include the rest of the license: https://github.com/gnosis/canonical-weth/blob/master/contracts/WETH9.sol

protocol-units/bridge/e2e/src/main.rs
```diff
@@ -0,0 +1,142 @@
...
+	if !build_output.status.success() {
+		eprint!("Failed to build contracts: {}", String::from_utf8_lossy(&build_output.stderr));
+		return Err(anyhow::anyhow!("Failed to build contracts"));
+	} else {
+		println!("forge build output:");
+		println!("{}", String::from_utf8_lossy(&build_output.stdout));
+	}
+
+	//Start Anvil
+	let _ = TokioCommand::new("anvil").stdout(Stdio::null()).stderr(Stdio::null()).spawn()?;
```
l-monninger, line 46: This probably conceals a fair bit of useful debugging info, no?

l-monninger:
@cde8 @0xmovses should we update the CODEOWNERS to place you two as the owners on the bridge?

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+
+		let seed = [3u8; 32];
+		let mut rng = rand::rngs::StdRng::from_seed(seed);
+		let signer = LocalAccount::generate(&mut rng);
+
+		Ok(MovementClient {
+			initiator_address: Vec::new(), //dummy for now
+			rest_client,
+			faucet_client,
+			counterparty_address: DUMMY_ADDRESS,
```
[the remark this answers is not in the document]
0xmovses: Yes, we will need to resolve how we get the `initiator_address` (eth side) into this client struct upon init. Probably some better top-level config for the bridge service.

protocol-units/bridge/e2e/src/main.rs
0xmovses: Yeah good point, I'll update.

0xmovses:
>@cde8 @0xmovses should we update the CODEOWNERS to place you two as the owners on the bridge?

I think it's beneficial for any of us to review and look at the bridge code, even if reviewers are not the main ones working on this part of the codebase.

protocol-units/bridge/cli/src/main.rs
```diff
@@ -1,3 +1,42 @@
...
+#[tokio::main]
+async fn main() -> Result<()> {
+	let cli = Cli::parse();
+
+	match &cli.command {
+		Commands::Eth { eth_command } => {
+			eth::run(eth_command).await?;
+		}
+		Commands::Movement { movement_command } => {
+			todo!()
```
andygolay, line 37: Seems ok to replace `todo!()` with `movement::run(movement_command).await?` considering how well fleshed-out `movement.rs` is?

protocol-units/bridge/cli/src/movement.rs
```diff
@@ -0,0 +1,141 @@
...
+			initiator,
+			bridge_transfer_id,
+			hash_lock,
+			time_lock,
+			recipient,
+			amount,
+		} => {
+			lock_assets(bridge_transfer_id, hash_lock, *time_lock, recipient, *amount).await?;
+		}
+		MovementCommands::Complete { config_path, bridge_transfer_id, pre_image } => {
```
andygolay, line 81: @0xmovses how is `config_path` used in these functions? It's not looking like it's passed in as an arg; would you pls explain the purpose of it? Is it the dot movement path?

andygolay (commented):
Just a couple preliminary comments / questions for @0xmovses as I've begun testing the CLI.

protocol-units/bridge/cli/src/movement.rs
0xmovses: dot movement, and any suzuka config is, for now, completly un-related to the bridge. I added `config_path` so that later we could add some .yaml or .json files to set some config values, but for now this is just pre-emptive, so this param may be removed before the feature branch gets merged.

andygolay:
Ah, got it, thanks.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+			send_transaction_error_rules,
+		})
+	}
+}
+
+impl<P> Clone for EthClient<P> {
+	fn clone(&self) -> Self {
+		todo!()
+	}
+}
```
cde8, line 207: This is interesting and necessary as the bridge clones the trait impls. I would suggest implementing this and either create a new connection on each clone or do something with a connection pool. Connection pool would be preferred. With a max of 32 connections.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+
+	async fn initiate_bridge_transfer(
+		&mut self,
+		_initiator_address: InitiatorAddress<Self::Address>,
+		recipient_address: RecipientAddress<Vec<u8>>,
+		hash_lock: HashLock<Self::Hash>,
+		time_lock: TimeLock,
+		amount: Amount,
+	) -> BridgeContractInitiatorResult<()> {
+		let contract = AtomicBridgeInitiator::new(self.initiator_address.0, &self.rpc_provider);
```
cde8, line 225: Initiator address from the point of the trait impls is the signer, not the initiator contract address

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+		let recipient_bytes: [u8; 32] = recipient_address.0.try_into().unwrap();
+		let call = contract.initiateBridgeTransfer(
+			U256::from(amount.0),
+			FixedBytes(recipient_bytes),
+			FixedBytes(hash_lock.0),
+			U256::from(time_lock.0),
+		);
```
cde8, line 232: I am missing here the initiator address

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+		let _ = send_transaction(
+			call,
+			&self.send_transaction_error_rules,
+			self.num_tx_send_retries,
+			self.gas_limit as u128,
+		)
```
cde8, line 238: We are pulling here functionality from the mcr crate. When we are building the UX and CLI we do not want to pull in the mcr crate. We should pull this in locally or have a shared crate for these send_transaction functions.

Also the bridge handles the retries so just pull this in.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+		Ok(())
+	}
+
+	async fn complete_bridge_transfer(
+		&mut self,
+		bridge_transfer_id: BridgeTransferId<Self::Hash>,
+		pre_image: HashLockPreImage,
+	) -> BridgeContractInitiatorResult<()> {
+		let pre_image: [u8; 32] = utils::vec_to_array(pre_image.0)
+			.unwrap_or_else(|_| panic!("Failed to convert pre_image"));
```
cde8, line 249: Try to avoid the panic!

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+		let contract = AtomicBridgeInitiator::new(self.initiator_address.0, &self.rpc_provider);
```
cde8, line 250: Same here

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+		bridge_transfer_id: BridgeTransferId<Self::Hash>,
+	) -> BridgeContractInitiatorResult<()> {
+		let contract = AtomicBridgeInitiator::new(self.initiator_address.0, &self.rpc_provider);
+		let call = contract.refundBridgeTransfer(FixedBytes(bridge_transfer_id.0));
+		let _ = send_transaction(
+			call,
+			&self.send_transaction_error_rules,
+			self.num_tx_send_retries,
+			self.gas_limit as u128,
+		)
```
cde8, line 274: See above, initator address and the send_transaction

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+			initiator_address: InitiatorAddress(eth_details.originator),
+			recipient_address: RecipientAddress(eth_details.recipient.to_vec()),
+			hash_lock: HashLock(eth_details.hash_lock),
+			time_lock: TimeLock(eth_details.time_lock.wrapping_to::<u64>()),
+			amount: Amount(eth_details.amount.wrapping_to::<u64>()),
+		};
+
+		Ok(Some(details))
+	}
+}
```
cde8, line 306: Interesting read, as I conclude when the smart contract changes the slot number change. Complicated stuff, just avoid the panics.

cde8 (changes requested):
Looks great, some changes needed for the bridge to operate optimally with the service crates. Will do another review when these are addressed.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+use thiserror::Error;
+use utils::EthAddress;
+
+pub mod utils;
+
+const INITIATOR_ADDRESS: &str = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266";
+const COUNTERPARTY_ADDRESS: &str = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"; //Dummy val
+const RECIPIENT_ADDRESS: &str = "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266";
+const DEFAULT_GAS_LIMIT: u64 = 10_000_000_000;
+const MAX_RETRIES: u32 = 5;
```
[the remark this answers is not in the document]
0xmovses: Yep, as discussed on our huddle. Config to be added later.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+	InitiatedBridgeTransfer(BridgeTransferDetails<A, H>),
+	CompletedBridgeTransfer(BridgeTransferId<H>, HashLockPreImage),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum AbstractBlockainEvent<A, H> {
+	InitiatorContractEvent(SCIResult<A, H>),
+	CounterpartyContractEvent(SCCResult<A, H>),
+	Noop,
+}
```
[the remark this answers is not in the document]
0xmovses: Yep these types are for event logging I'll remove them from this PR.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+pub struct Config {
+	pub rpc_url: Option<String>,
+	pub ws_url: Option<String>,
+	pub chain_id: String,
+	pub signer_private_key: String,
+	pub initiator_address: String,
+	pub counterparty_address: String,
+	pub recipient_address: String,
+	pub gas_limit: u64,
+	pub num_tx_send_retries: u32,
```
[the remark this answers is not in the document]
0xmovses: The other Config types have been update but `signer_private_key` field seems better as a `String` because the type it parses to `LocalSigner<SigningKey<Secp256k1>>` is not `Deserialize`. And `Config needs to be deserializable if we want to do:
```
fn load_config(path: &str) -> Result<Config> {
	match std::fs::read_to_string(path) {
		Ok(config_str) => {
			let config: Config = serde_json::from_str(&config_str)?;
			Ok(config)
		}
		Err(_) => {
			println!("Config file not found, Using default config values");
			Ok(Config::default())
		}
	}
}
```
Which allows a config to be loaded from a path.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+	Noop,
+}
+
+///Configuration for the Ethereum Bridge Client
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
+	pub rpc_url: Option<String>,
+	pub ws_url: Option<String>,
+	pub chain_id: String,
+	pub signer_private_key: String,
```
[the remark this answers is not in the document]
0xmovses: The other Config types have been updated but `signer_private_key` field seems better as a `String` because the type it parses to `LocalSigner<SigningKey<Secp256k1>>` is not `Deserialize`. And `Config needs to be deserializable if we want to do:
```
fn load_config(path: &str) -> Result<Config> {
	match std::fs::read_to_string(path) {
		Ok(config_str) => {
			let config: Config = serde_json::from_str(&config_str)?;
			Ok(config)
		}
		Err(_) => {
			println!("Config file not found, Using default config values");
			Ok(Config::default())
		}
	}
}
```
Which allows a config to be loaded from a path.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+	CompletedBridgeTransfer(CounterpartyCompletedDetails<A, H>),
+}
+
+#[derive(Debug, Error, Clone, PartialEq, Eq)]
+pub enum MoveCounterpartyError {
+	#[error("Transfer not found")]
+	TransferNotFound,
+	#[error("Invalid hash lock pre image (secret)")]
+	InvalidHashLockPreImage,
+}
```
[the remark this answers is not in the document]
0xmovses: Yes these are for the event PR, left over here. Removed.

protocol-units/bridge/chains/ethereum/src/lib.rs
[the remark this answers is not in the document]
0xmovses: The fields on the client have been updated to their native chain types. However we still need the fields as `send_transaction` requires them 

```rust
let _ = send_transaction(
			call,
			&self.send_transaction_error_rules,
			self.num_tx_send_retries,
			self.gas_limit as u128,
		)
```
The values never get changed. So if we don't want them on the client they would need to be stored somewhere else in memory, and then passed in. For now  I think its easy and ergonomic enough for them to remain on the client. Maybe after we're done we will see somewhere where they can be accessed.

0xmovses:
Btw @cde8 I tried to create a connection pool by making the `EthClient` struct provider fields of these types 
```rust
pub struct EthClient<P> {
	rpc_providers: Arc<Mutex<Vec<P>>>,
	ws_providers: Arc<Mutex<Vec<RootProvider<PubSubFrontend>>>>,
	...
```
Then just push into the Vec of thread save connections. 

However, it wasn't a quick fix to implement `Clone` on it because to create a new provider is an async op 
```
let rpc_provider = ProviderBuilder::new()
			.with_recommended_fillers()
			.wallet(EthereumWallet::from(signer.clone()))
			.on_builtin(&rpc_url)
			.await?;
```
We can't do an async call inside `fn clone`, so we need to create a proper connection manager and then Clone can safetly clone connections. I opt for this in a seperate PR as it's not trivial and not blocking this from going in.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+	async fn get_bridge_transfer_details(
+		&mut self,
+		bridge_transfer_id: BridgeTransferId<Self::Hash>,
+	) -> BridgeContractInitiatorResult<Option<BridgeTransferDetails<Self::Address, Self::Hash>>> {
+		let mapping_slot = U256::from(0); // the mapping is the zeroth slot in the contract
+		let key = bridge_transfer_id.0;
+		let storage_slot = self.calculate_storage_slot(key, mapping_slot);
+		let storage: U256 = self
+			.rpc_provider
+			.get_storage_at(self.initiator_address.0, storage_slot)
```
cde8, line 258: This makes me think this is the intiator contract address not the inititiator address which makes sense.

protocol-units/bridge/chains/ethereum/src/utils.rs
```diff
@@ -0,0 +1,174 @@
...
+					// Verify all rules. If one rule return true or an error stop verification.
+					// If true retry with more gas else return the error.
+					if rule.verify(&err)? {
+						//increase gas of 10% and retry
+						estimate_gas += (estimate_gas * 10) / 100;
+						tokio::time::sleep(tokio::time::Duration::from_secs(1)).await;
+						continue;
+					}
+				}
+				return Err(BridgeContractInitiatorError::UnknownRpcError.into());
```
cde8, line 135: it should be generic error, is specific

protocol-units/bridge/chains/ethereum/src/utils.rs
```diff
@@ -0,0 +1,174 @@
...
+						format!(
+						"Send commitment Transaction fail, abort Transaction, receipt:{transaction_receipt:?}"
+					),
+					)
+					.into());
+				}
+			}
+			Ok(_) => return Ok(()),
+			Err(err) => {
+				return Err(BridgeContractInitiatorError::RpcTransactionExecutionError(
```
cde8, line 161: Should be generic error is specific

cde8 (changes requested):
Improving, we need the counterparty implementation for the Abstract bridge to be initialized.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+	pub time_lock: U256,
+	pub state: u8, // Assuming the enum is u8 for now..
+}
+
+pub struct EthClient<P> {
+	rpc_provider: P,
+	chain_id: String,
+	ws_provider: RootProvider<PubSubFrontend>,
+	initiator_address: EthAddress,
+	signing_address: EthAddress,
```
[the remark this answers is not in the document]
0xmovses: We need the signer_address to initialise the contract `let contract = AtomicBridgeInitiator::new(self.signing_address.0, &self.rpc_provider);`, this is not always the `initiator_address` because it can be called by any client.

protocol-units/bridge/chains/ethereum/src/utils.rs
0xmovses: Ok will change those.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+	Result<BridgeContractCounterpartyEvent<A, H>, BridgeContractCounterpartyError>;
+
+///Configuration for the Ethereum Bridge Client
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
+	pub rpc_url: Option<String>,
+	pub ws_url: Option<String>,
+	pub chain_id: String,
+	pub signer_private_key: String,
+	pub initiator_address: EthAddress,
```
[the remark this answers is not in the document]
0xmovses: I gues we can just use the signing address here instead `let contract = AtomicBridgeInitiator::new(initiator_address.0 .0, &self.rpc_provider);`. I thought it might be useful to store the initiator address and the signer address as different fields. But maybe its redunant.

protocol-units/bridge/chains/ethereum/src/lib.rs
0xmovses: What do you mean? It's passed in here `let contract = AtomicBridgeInitiator::new(initiator_address.0 .0, ..`

protocol-units/bridge/chains/ethereum/src/utils.rs
```diff
@@ -0,0 +1,174 @@
...
+}
+
+pub(crate) struct ProviderArgs {
+	pub rpc_provider: AlloyProvider,
+	pub ws_provider: RootProvider<PubSubFrontend>,
+	pub initiator_address: EthAddress,
+	pub signing_address: EthAddress,
+	pub counterparty_address: EthAddress,
+	pub gas_limit: u64,
+	pub num_tx_send_retries: u32,
```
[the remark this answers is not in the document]
0xmovses: We need this value for `send_transaction`. If we want to refactor the transaction fn, we can but I think it's not that important for now?

```
let _ = send_transaction(
			call,
			&self.send_transaction_error_rules,
			self.num_tx_send_retries,
			self.gas_limit as u128,
		)
```

protocol-units/bridge/chains/ethereum/src/utils.rs
```diff
@@ -0,0 +1,174 @@
...
+>(
+	base_call_builder: CallBuilder<T, &&P, D, Ethereum>,
+	send_transaction_error_rules: &[Box<dyn VerifyRule>],
+	number_retry: u32,
+	gas_limit: u128,
+) -> Result<(), anyhow::Error> {
+	//validate gas price.
+	let mut estimate_gas = base_call_builder.estimate_gas().await?;
+	// Add 20% because initial gas estimate are too low.
+	estimate_gas += (estimate_gas * 20) / 100;
```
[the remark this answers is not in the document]
0xmovses: Can we address this in a seperate issue / PR? I need this to go into the feature branch for my event logging work kinda soon.

0xmovses (commented):
@cde8 The counterparty implementation is done (movement side) under `chains/movement`, its in this PR. Keen to get your comments in there so we can get this merged in. Thanks!

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+			recipient_address: RecipientAddress(eth_details.recipient.to_vec()),
+			hash_lock: HashLock(eth_details.hash_lock),
+			time_lock: TimeLock(eth_details.time_lock.wrapping_to::<u64>()),
+			amount: Amount(eth_details.amount.wrapping_to::<u64>()),
+		};
+
+		Ok(Some(details))
+	}
+}
+
```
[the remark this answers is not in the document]
0xmovses: You mean on eth side? On movement side it is of course. And all the aptos counterparty code is here in this PR 👍

protocol-units/bridge/chains/ethereum/src/lib.rs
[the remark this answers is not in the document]
0xmovses: I sure can, but I'd say this is beyond the scope of this PR which is just to get the crate implementations in. Config should be handled later as its non-trivial. And will likely change quite a bit.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+		}
+	}
+}
+
+impl Config {
+	fn default_for_private_key() -> String {
+		let random_wallet = PrivateKeySigner::random();
+		random_wallet.to_bytes().to_string()
+	}
+}
```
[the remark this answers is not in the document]
0xmovses: For e2e testing, `default()` will allow us to gen a contract address with ease and without providing a config file at a specified path. I added it for that reason. Can remove if you like.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,294 @@
...
+
+///Configuration for the Ethereum Bridge Client
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct Config {
+	pub rpc_url: Option<String>,
+	pub ws_url: Option<String>,
+	pub chain_id: String,
+	pub signer_private_key: String,
+	pub initiator_address: EthAddress,
+	pub recipient_address: String,
```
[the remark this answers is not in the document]
0xmovses: That's been removed.

0xmovses:
Regarding the refactor of the `send_transaction` function, to not use use MCR function but create our own. This is a seperate PR / issue. 

The same for Config init and subrelated comments, these are imporvements that shouldn't block us for now. It's only when the feature branch goes into main itself that these issues need to be resolved for.

protocol-units/bridge/chains/ethereum/src/utils.rs
```diff
@@ -0,0 +1,174 @@
...
+		Ok(vec.into())
+	}
+}
+
+pub(crate) struct ProviderArgs {
+	pub rpc_provider: AlloyProvider,
+	pub ws_provider: RootProvider<PubSubFrontend>,
+	pub initiator_address: EthAddress,
+	pub signing_address: EthAddress,
+	pub counterparty_address: EthAddress,
```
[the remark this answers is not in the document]
0xmovses: Removed now.

0xmovses:
@cde8 With the TODO's added let's turn them into issues. And track them in the feature branch PR.

0xmovses:
The e2e tests aren't passing now with the `build_with_provider` methods removed. I run the tests with `cargo r -p bridge-e2e-tests`. 

I would say we keep that part in so that we can test all the calls are working as we continue. On commit `e2583e541ded442c5bf4b6368f227848c05cfe6c` the tests were running and various calls were being made successfully to the contracts.

protocol-units/bridge/chains/ethereum/src/lib.rs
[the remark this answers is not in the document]
cde8: Yes, the implementation of the abstract bridge necessitates completion on both sides concurrently. If only one side is implemented, it will be necessary to separate it.

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -108,14 +108,13 @@ where
...
-		// TODO: (richard) we are missing the initiator address here
-		_initiator_address: InitiatorAddress<Self::Address>,
+		initiator_address: InitiatorAddress<Self::Address>,
 		recipient_address: RecipientAddress<Vec<u8>>,
 		hash_lock: HashLock<Self::Hash>,
 		time_lock: TimeLock,
 		amount: Amount,
 	) -> BridgeContractInitiatorResult<()> {
-		let contract = AtomicBridgeInitiator::new(self.initiator_contract.0, &self.rpc_provider);
+		let contract = AtomicBridgeInitiator::new(initiator_address.0 .0, &self.rpc_provider);
```
cde8, line 117: @0xmovses Is the `AtomicBridgeInitiator` not the contract? As such, it should be initialized with the contract address.

cde8 (changes requested):
Question on the address provided

protocol-units/bridge/chains/ethereum/src/lib.rs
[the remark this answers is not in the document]
0xmovses: Hmm I thought the whole plan was to stand up the abstract bridge and E2E test it in one direction first. If this is the case, we will need to prioritise the contracts for the other direct. What's your preference?

protocol-units/bridge/chains/ethereum/src/lib.rs
0xmovses: ah yes, I think the only reason why the tests were passing as it was the same address being used to deploy and then to call. Updating

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -0,0 +1,199 @@
...
+			.on_builtin(&rpc_url)
+			.await?;
+
+		Ok(EthClient { rpc_provider, initiator_contract: config.initiator_contract })
+	}
+}
+
+impl<P> Clone for EthClient<P> {
+	fn clone(&self) -> Self {
+		todo!()
```
cde8, line 97: No need to implement it, otherwise the abstract bridge will not function.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+	///Address of the counterparty moduke
+	counterparty_address: AccountAddress,
+	///Address of the initiator module
+	initiator_address: Vec<u8>,
+	///The Apotos Rest Client
+	rest_client: Client,
+	faucet_client: FaucetClient,
+	///The signer account
+	signer: LocalAccount,
+}
```
cde8, line 46: I notice several project-specific configurations, such as the Suzuka configuration and dot movement settings, within our main framework that compromise its modularity by adding unnecessary complexity. 

To maintain the architectural integrity and clarity of our system, I propose that we either remove or refactor these configurations. Please let me know your thoughts on this matter.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		let faucet_listen_port = suzuka_config
+			.execution_config
+			.maptos_config
+			.client
+			.maptos_faucet_rest_connection_port;
+		let faucet_url = format!("http://{}:{}", faucet_listen_address, faucet_listen_port);
+		let faucet_url = Url::from_str(faucet_url.as_str()).unwrap();
+
+		let rest_client = Client::new(node_connection_url.clone());
+		let faucet_client = FaucetClient::new(faucet_url, node_connection_url.clone());
```
cde8, line 79: What is the necessity of integrating this into our client-side models?

protocol-units/bridge/chains/movement/src/lib.rs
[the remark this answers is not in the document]
cde8: Did we already resolve this, would like to get this out of the code

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+			faucet_client,
+			counterparty_address: DUMMY_ADDRESS,
+			signer,
+		})
+	}
+}
+
+impl Clone for MovementClient {
+	fn clone(&self) -> Self {
+		todo!()
```
cde8, line 97: This needs to be implemented for the bridge to function.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		amount: Amount,
+	) -> BridgeContractCounterpartyResult<()> {
+		//@TODO properly return an error instead of unwrapping
+		let args = vec![
+			self.to_bcs_bytes(&initiator.0).unwrap(),
+			self.to_bcs_bytes(&bridge_transfer_id.0).unwrap(),
+			self.to_bcs_bytes(&hash_lock.0).unwrap(),
+			self.to_bcs_bytes(&time_lock.0).unwrap(),
+			self.to_bcs_bytes(&recipient.0).unwrap(),
+			self.to_bcs_bytes(&amount.0).unwrap(),
```
cde8, line 122: Another home-rolled binary serialization format, as if we don't have enough already. 🙂

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		let payload = utils::make_aptos_payload(
+			self.counterparty_address,
+			COUNTERPARTY_MODULE_NAME,
+			"lock_bridge_transfer_assets",
+			self.counterparty_type_args(Call::Lock),
+			args,
+		);
+		let _ = utils::send_aptos_transaction(&self.rest_client, &mut self.signer, payload)
+			.await
+			.map_err(|_| BridgeContractCounterpartyError::LockTransferAssetsError);
```
cde8, line 133: You are mapping an error but neither using it nor effectively short-circuiting it. So why are you doing this? It has no value.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		let payload = utils::make_aptos_payload(
+			self.counterparty_address,
+			COUNTERPARTY_MODULE_NAME,
+			"complete_bridge_transfer",
+			self.counterparty_type_args(Call::Complete),
+			args,
+		);
+		let _ = utils::send_aptos_transaction(&self.rest_client, &mut self.signer, payload)
+			.await
+			.map_err(|_| BridgeContractCounterpartyError::CompleteTransferError);
```
cde8, line 156: You are mapping an error but neither using it nor effectively short-circuiting it. So why are you doing this? It has no value.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		let payload = utils::make_aptos_payload(
+			self.counterparty_address,
+			COUNTERPARTY_MODULE_NAME,
+			"abort_bridge_transfer",
+			self.counterparty_type_args(Call::Abort),
+			args,
+		);
+		let _ = utils::send_aptos_transaction(&self.rest_client, &mut self.signer, payload)
+			.await
+			.map_err(|_| BridgeContractCounterpartyError::AbortTransferError);
```
cde8, line 177: You are mapping an error but neither using it nor effectively short-circuiting it. So why are you doing this? It has no value.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+		&mut self,
+		_bridge_transfer_id: BridgeTransferId<Self::Hash>,
+	) -> BridgeContractCounterpartyResult<Option<BridgeTransferDetails<Self::Hash, Self::Address>>>
+	{
+		// let _ = utils::send_view_request(
+		// 	self.rest_client,
+		// 	self.counterparty_address,
+		// 	"atomic_bridge_counterparty".to_string(),
+		// );
+		todo!();
```
cde8, line 191: Needs to be implemented for the client implementation to work

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+}
+
+impl MovementClient {
+	fn counterparty_type_args(&self, call: Call) -> Vec<TypeTag> {
+		match call {
+			Call::Lock => vec![TypeTag::Address, TypeTag::U64, TypeTag::U64, TypeTag::U8],
+			Call::Complete => vec![TypeTag::Address, TypeTag::U64, TypeTag::U8],
+			Call::Abort => vec![TypeTag::Address, TypeTag::U64],
+			Call::GetDetails => vec![TypeTag::Address, TypeTag::U64],
+		}
```
cde8, line 202: These should be utility functions as self is never used.

protocol-units/bridge/chains/movement/src/utils.rs
```diff
@@ -0,0 +1,170 @@
...
+	/// Optional sequence data that is useful during indexing
+	pub sequence: Option<u32>,
+}
+
+/// Send Aptos Transaction
+pub async fn send_aptos_transaction(
+	rest_client: &RestClient,
+	signer: &mut LocalAccount,
+	payload: TransactionPayload,
+) -> Result<AptosTransaction> {
```
cde8, line 74: Avoid using anyhow::Error in libs

protocol-units/bridge/chains/movement/src/utils.rs
```diff
@@ -0,0 +1,170 @@
...
+		.into_inner();
+	Ok(response)
+}
+
+// This is not used for now, but we may need to use it in later. It's used for estimating gas
+pub async fn simulate_aptos_transaction(
+	aptos_client: &MovementClient,
+	signer: &mut LocalAccount,
+	payload: TransactionPayload,
+) -> Result<TransactionInfo> {
```
cde8, line 100: Avoid using anyhow::Error in libs

protocol-units/bridge/chains/movement/src/utils.rs
```diff
@@ -0,0 +1,170 @@
...
+
+/// Send View Request
+pub async fn send_view_request(
+	aptos_client: &MovementClient,
+	package_address: String,
+	module_name: String,
+	function_name: String,
+	type_arguments: Vec<MoveType>,
+	arguments: Vec<serde_json::Value>,
+) -> Result<Vec<serde_json::Value>, anyhow::Error> {
```
cde8, line 154: Avoid using anyhow::Error in libs

cde8 (changes requested):
Movement side first review

protocol-units/bridge/cli/src/main.rs
```diff
@@ -1,3 +1,42 @@
...
+		Commands::Eth { eth_command } => {
+			eth::run(eth_command).await?;
+		}
+		Commands::Movement { movement_command } => {
+			movement::run(movement_command).await?;
+		}
+	}
+
+	Ok(())
 }
```
cde8, line 42: It appears that the command line interface developed is primarily suited for testing purposes rather than for client use in asset exchange. We will need to change this component accordingly. This might also influence the e2e test. Maybe rename this to a testing cli or something.

protocol-units/bridge/chains/ethereum/src/lib.rs
0xmovses: See issue (with its own PR) https://github.com/movementlabsxyz/movement/issues/250

protocol-units/bridge/chains/movement/src/lib.rs
0xmovses: Yes fine, I was just doing what I needed to stand this up, again everything config can be done later and merged into the feature bracn at a later date.

protocol-units/bridge/chains/movement/src/lib.rs
0xmovses: We will need a faucet for the tests, so that a user can get tokens to deploy.

protocol-units/bridge/cli/src/eth.rs
```diff
@@ -0,0 +1,193 @@
...
+
+async fn initiate_transfer(
+	config: Config,
+	initiator_address: &str,
+	recipient_address: &str,
+	hash_lock: &str,
+	time_lock: u64,
+	amount: u64,
+) -> Result<()> {
+	println!("config {:?}", config);
```
cde8, line 132: remove the println! use tracing

protocol-units/bridge/chains/movement/src/utils.rs
0xmovses: Yes again, mostly just placeholder code to be handled properly in this issue https://github.com/movementlabsxyz/movement/issues/224#issue-2413712253

I think its ok to merge some of this placeholder code into the feature branch (but not into main!)

protocol-units/bridge/chains/movement/src/lib.rs
0xmovses: Yes.

cde8 (changes requested):

cde8 (approved):
I'm going to accept this PR. I might add some changes while I'm implementing the command line interface.

PR #156, opened 2024-07-01
feat: suzuka with manager and godfig.

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Uses `godfig` with manger pattern to run Suzuka Full Node.

# Changelog

1. Uses `godfig::try_wait_for` to await the appropriate config setting in Suzuka. 

# Testing

1. `just suzuka-full-node native build.setup.test.local`. 

# Outstanding issues
1. Still starts services one after another. This will not be modified until Suzuka Full Node go around. 
2. More clean can be cone, but pausing here to keep PR smallish.

### DIFF
--- a/networks/suzuka/suzuka-full-node/src/manager.rs
+++ b/networks/suzuka/suzuka-full-node/src/manager.rs
@@ -1,0 +1,43 @@
+use crate::SuzukaFullNode;
+use anyhow::Context;
+use super::partial::SuzukaPartialNode;
+use godfig::{
+    Godfig,
+    backend::config_file::ConfigFile
+};
+use suzuka_config::Config;
+use maptos_dof_execution::v1::Executor;
+
+#[derive(Clone)]
+pub struct Manager<Dof>
+    where 
+    Dof : SuzukaFullNode {
+    godfig: Godfig<Config, ConfigFile>,
+    _marker : std::marker::PhantomData<Dof>,
+}
+
+// Implements a very simple manager using a marker strategy pattern.
+impl Manager<SuzukaPartialNode<Executor>> {
+    pub async fn new(file : tokio::fs::File) -> Result<Self, anyhow::Error> {
+        let godfig = Godfig::new(ConfigFile::new(file), vec![]);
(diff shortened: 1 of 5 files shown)


networks/suzuka/suzuka-full-node/src/manager.rs
```diff
@@ -0,0 +1,43 @@
...
+impl Manager<SuzukaPartialNode<Executor>> {
+    pub async fn new(file : tokio::fs::File) -> Result<Self, anyhow::Error> {
+        let godfig = Godfig::new(ConfigFile::new(file), vec![]);
+        Ok(Self {
+            godfig,
+            _marker: std::marker::PhantomData,
+        })
+    }
+
+    pub async fn try_run(&self) -> Result<(), anyhow::Error> {
```
musitdev, line 29: I think the `Manager<SuzukaPartialNode<Executor>> should only implement `try_run(file : tokio::fs::File)` and the Manager<Dof> struct shouldn't have the Godfile as a field.
The try_run never end, so there's no need to create a struct to only start a long-running task, and it allows the other manager impl to use different config if needed.
The other possibility is to share the same config and propose a try_run for each executor like:
```
impl Manager<Dof> {
    pub async fn new(file : tokio::fs::File) -> Result<Self, anyhow::Error> {
        let godfig = Godfig::new(ConfigFile::new(file), vec![]);
        let config = self.godfig.try_wait_for_ready().await?;
        Ok(Self {
            config,
            _marker: std::marker::PhantomData,
        })
    }
impl Manager<SuzukaPartialNode<Executor>> {
       ... Suzuka node impl
    }
}
impl Manager<MonzaPartialNode<Executor>> {
       ...Monza node impl
    }
}
 ````

mzabaluev, line 29: I think it's fine to name it `run`, there is no name convention that says you should distinguish all fallible methods with a `try_` prefix.

Also perhaps, pass the recipient as `self` to not constrain the implementation into interior mutability.

PR #169, opened 2024-07-05
L monninger/godfig results

l-monninger:
# Summary

# Changelog

# Testing

# Outstanding issues


util/godfig/src/backend/mod.rs
```diff
@@ -53,4 +56,11 @@ pub trait BackendOperations {
         T: serde::de::DeserializeOwned + serde::Serialize + Send,
         F: FnOnce(Option<T>) -> Fut + Send,
         Fut: std::future::Future<Output = Result<Option<T>, GodfigBackendError>> + Send;
+
+    async fn try_transaction_with_result<K, T, R, F, Fut>(&self, key: K, callback: F) -> Result<R, GodfigBackendError>
```
mzabaluev, line 60: Can this be documented? There are five template parameters, so it's a bit tricky to decipher. What are the expectations for the closure? It works on `Option<T>` and returns `(Option<T>, R)`, I guess as a way to insert or remove or change the value and return some result.

The `try_` prefix is not necessary.

The `try_transaction` method (or, better, `transaction`) can now have a default implementation in terms of `try_transaction_with_result`.

util/godfig/src/godfig/mod.rs
```diff
@@ -42,6 +42,16 @@ where
         res
     }
 
+    pub async fn try_transaction_with_result<R, F, Fut>(&self, callback: F) -> Result<R, GodfigBackendError>
+    where
+        F: FnOnce(Option<Contract>) -> Fut + Send,
+        Fut: std::future::Future<Output = Result<(Option<Contract>, R), GodfigBackendError>> + Send
```
mzabaluev, line 48: Why is this one specific for `Contract`?

util/godfig/src/godfig/mod.rs
```diff
@@ -69,7 +79,7 @@ pub mod test {
         let backend = ConfigFile::new(tempfile.into());
         let godfig : Godfig<Test, ConfigFile> = Godfig::new(backend, vec!["test".to_string()]);
 
-        godfig.try_transaction(|data| async move {
+        godfig.try_transaction(|_data| async move {
```
mzabaluev, line 82: It would be good to exercise a transaction on an existing value and make sure it's passed to the closure.

util/godfig/src/backend/mod.rs
```diff
@@ -53,4 +56,11 @@ pub trait BackendOperations {
...
+        where
+        K: Into<Vec<String>> + Send,
+        T: serde::de::DeserializeOwned + serde::Serialize + Send,
```
mzabaluev, line 63: All import paths should be in use statements at the top of the file, this adds a lot of clutter.

PR #170, opened 2024-07-08
Alloy `msg.sender` is Zero w/ Anvil

l-monninger:
# Summary
- **RFCs**: $\emptyset$
- **Categories**: `debug`

`msg.sender` is getting decoded as 0 by deployed Anvil contracts when using Alloy providers. 

## Reproduce
The easiest way to repdroduce and  debug is to following this approach.

1. When you have made modifications that affect core contract deployment or contract logic, run `just mcr-client native build.local -t=false` 
2. In a separate shell run, `cargo test -p mcr-settlement-client test_genesis_ceremony`. 

You should now see output similar to the following:

```
2024-07-08T07:59:40.076743Z TRACE ReqwestTransport{url=http://0.0.0.0:8545/}: alloy_transport_http::reqwest_transport: response body body={"jsonrpc":"2.0","id":3,"error":{"code":3,"message":"execution reverted: ","data":"0xe2517d3f00000000000000000000000000000000000000000000000000000000000000009f2df0fed2c77648de5860a4cc508cd0818c85b8b8a1ab4ceeef8d981c8956a6"}}
test tests::e2e::genesis_ceremony::test_genesis_ceremony ... FAILED

failures:

---- tests::e2e::genesis_ceremony::test_genesis_ceremony stdout ----
Error: Governor failed to mint for alice

Caused by:
    server returned an error response: error code 3: execution reverted: 


failures:
    tests::e2e::genesis_ceremony::test_genesis_ceremony
```

Which you can decode via `forge selectors list MintableToken` to:

```
| Error    | AccessControlUnauthorizedAccount(address,bytes32)   | 0xe2517d3f     
```

This is unexpected and seemingly logically inconsistent because our call to `hasMinterRole` appropriately returned true for the governor address. And, that is the only role restriction on the method.

```solidity
function mint(
        address to,
        uint256 amount
    ) public virtual onlyRole(MINTER_ROLE) {
        _mint(to, amount);
}
```

To further debug, go ahead and remove `onlyRole(MINTER_ROLE)` and restart the network.

```solidity
function mint(
        address to,
        uint256 amount
    ) public virtual {
        _mint(to, amount);
}
```

You should now pass the mint step, but should get an error code at the `alice.approve` step which matches this:

```
| Error    | ERC20InvalidApprover(address)                       | 0xe602df05      
```

If you refer to the OpenZeppelin source, you'll see this only occurs when the owner address `msg.sender == 0`: https://github.com/search?q=repo%3AOpenZeppelin%2Fopenzeppelin-contracts%20ERC20InvalidApprover&type=code

## Behavior
In general, we have the following behavior:
1. All view methods not using `msg.sender` behave as expected. 
2. All other methods do not, appearing to have `msg.sender = 0`.

## Attempts to Debug and Recommendations
1. At first this seemed like potentially a proxy issue, I've tried both `UUPS` and `TransparentProxy` in a variety of different deployment scripts to no avail. But, importantly, I am not using the `Upgrades` module from `OpenZeppelin` directly.
2. I've tried playing around with the `alloy::providers::ProviderBuilder` configs a bit--particularly thinking this might be a `chain_id` issue. But, it does not appear so. 
3. There is a special `anvil` provider for `alloy` perhaps this is needed. 
4. There are various lines commented `// debug: ` in the code. Search these to help identify debug points. 

# Changelog

1. Included is a small section of a larger PR where this is occurring throughout.

# Testing

# Outstanding issues


protocol-units/settlement/mcr/setup/src/local.rs
```diff
@@ -1,65 +1,207 @@
...
+				anyhow!("Governor address not found in Anvil addresses"),
+			)?.address.clone();
+
+			// todo: make sure this shows up in the docker container as well
+			let mut solidity_path = std::env::current_dir()?;
+			solidity_path.push("protocol-units/settlement/mcr/contracts");
+
+			let solidity_path = solidity_path.to_string_lossy();
+			tracing::info!("solidity_path: {:?}", solidity_path);
+			let output_exec = run_command(
```
l-monninger, line 102: Add it above here.

l-monninger: `anvil build`

.gitmodules
andygolay, line 1: I'm currently getting 

```
[eth    ] Caused by:
[eth    ]     Command forge failed with args ["script", "DeployMCRDev", "--root", "/Users/andygmove/Downloads/repos/movement/protocol-units/settlement/mcr/contracts", "--broadcast", "--chain-id", "1358", "--sender", "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266", "--rpc-url", "http://0.0.0.0:8545", "--private-key", "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80"]
[eth    ]     Error Output: Error: 
[eth    ]     failed to resolve file: "/Users/andygmove/Downloads/repos/movement/protocol-units/settlement/mcr/contracts/lib/forge-std/src/Script.sol": No such file or directory (os error 2); check configured remappings
[eth    ]       --> /Users/andygmove/Downloads/repos/movement/protocol-units/settlement/mcr/contracts/script/DeployMCRDev.s.sol
[eth    ]       forge-std/Script.sol
```

and other such errors when I run 

```
just mcr-client native build.local -t=false
```

Not sure if it's related to this change; the `just` command was running for me about an hour ago on the same branch.

l-monninger:
```
forge compile
```
in the contract dir, sorry.

l-monninger: That's what I was commenting to @0xPrimata above

0xPrimata: To resolve @andygolay 's issue, you need to cd protocol-units/settlement/mcr/contracts && forge compile, it will download all required libs. Had the same issue.

0xPrimata:
Having an issue with

```
---- tests::e2e::genesis_ceremony::test_genesis_ceremony stdout ----
Error: Dot movement path not provided
```

echo $DOT_MOVEMENT_PATH doesn't return anything while just is running. I'm not seeing it declared on flake.nix but I can see it in `docker-compose.local.yml`. Is there anything I need to do to get it to actually get declared?

protocol-units/settlement/mcr/setup/src/local.rs
0xPrimata: what did you mean here?

andygolay:
> Having an issue with
> 
> ```
> ---- tests::e2e::genesis_ceremony::test_genesis_ceremony stdout ----
> Error: Dot movement path not provided
> ```
> 
> echo $DOT_MOVEMENT_PATH doesn't return anything while just is running. I'm not seeing it declared on flake.nix but I can see it in `docker-compose.local.yml`. Is there anything I need to do to get it to actually get declared?

Yeah I'm getting this as well. 

I tried 

```
export DOT_MOVEMENT_PATH=""
```

for a few paths that I thought might work, but that only made the test hang. It seems to me like `dot-movement` is supposed to create the path automatically but it isn't, for some reason?

l-monninger:
```
DOT_MOVEMENT_PATH="$(pwd)/.movement" RUST_LOG=trace cargo test -p mcr-settlement-client test_genesis_ceremony
```

l-monninger:
And, delete in between runs for the moment.

protocol-units/settlement/mcr/setup/src/local.rs
l-monninger: See how there's a `run_command` for the `forge sciprt` script. Add a `forge build/compile` above it.

andygolay:
I was able to fiddle with debugging to the point of getting the same approval failure error.

```
---- tests::e2e::genesis_ceremony::test_genesis_ceremony stdout ----
Error: Alice failed to approve MCR
```

It does not make sense to me that `governer_address` would be 0, nor do the tests show that my `alloy_signer_wallet::LocalWallet;` is 0. 

So I tend to think Alloy and Anvil are not communicating nicely with each other for some odd reason.

@l-monninger you mentioned "There is a special anvil provider for alloy perhaps this is needed."

And you mentioned bumping the Alloy version.

At this point I think both of those are worth a try. Lmk if you'd like me to take a stab at it?

If so, any info you have about that Anvil provider for Alloy would be helpful.

andygolay:
I tried modifying the minting to Alice as follows:

```
let mint_call = governor_token.mint(alice_address, U256::from(100));
let send_future = mint_call.send();
let mint_tx = send_future.await.context("Governor failed to mint for alice")?;

let watch_future = mint_tx.watch();
let mint_receipt = watch_future.await.context("Failed to get transaction receipt for minting")?;
info!("Mint transaction receipt: {:?}", mint_receipt);
```

So, no progress by switching from `call()` to `send()` and `watch()`.

0xPrimata (approved):
Test is passing, no conflicts with main LGTM, went through code changes earlier today.

Issue #167, opened 2024-07-04
Faucet Service not providing a response


0xPrimata:
**Describe the bug**
It seems like Aptos TS SDK expects a message being returned once the fundAccount() request is made but it never reaches it. Tracing it back has shown to be relatively hard since it points to ANS (?) module, something not present in our network as far as I know.
Aptos Devnet, afaik, does not have the same issue. I hit a blocker because I got rate limited. Will report tomorrow again on this issue if I can reproduce it on their network.

**To Reproduce**
1. [Build a Typescript SDK FundAccount request](https://aptos.dev/en/build/sdks/ts-sdk/quickstart)
2. Config it to request to any of our networks
3. You should get an error: `Cannot use 'in' operator to search for 'message' in Not Found`

**Expected behavior**

**Screenshots**
```
chunk TypeError: Cannot use 'in' operator to search for 'message' in <!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Cannot POST /fund</pre>
</body>
</html>

    at Bn (/Users/primata/movement/networks/suzuka/suzuka-client/src/tests/primata-entropy/army/node_modules/@aptos-labs/ts-sdk/src/core/crypto/keyless.ts:177:45)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Es (/Users/primata/movement/networks/suzuka/suzuka-client/src/tests/primata-entropy/army/node_modules/@aptos-labs/ts-sdk/src/internal/ans.ts:243:7)
    at async gn.fundAccount (/Users/primata/movement/networks/suzuka/suzuka-client/src/tests/primata-entropy/army/node_modules/@aptos-labs/ts-sdk/src/internal/ans.ts:284:33)
```

**Desktop (please complete the following information):**
 - OS: Mac

**Additional context**

PR #171, opened 2024-07-08
Sequencer w/ Sequence Numbers

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Adds sequence numbers to sequencer transaction types and respects them in ordering. 

Introduced in part to combat issues related to: https://github.com/movementlabsxyz/movement/issues/167

# Changelog

1. Modifies `movement_types::Transaction` to include a a sequencer number:
```rust
pub struct Transaction {
	pub data: Vec<u8>,
	pub sequence_number: u64,
}
```
2. Updates ordering of mempool transactions to respect said sequencer number. Both in memory:
```rust
impl Ord for MempoolTransaction {
	fn cmp(&self, other: &Self) -> Ordering {
		// First, compare by slot_seconds
		match self.slot_seconds.cmp(&other.slot_seconds) {
			Ordering::Equal => {}
			non_equal => return non_equal,
		}

		// If slot_seconds are equal, then compare by sequence number
		match self.transaction.sequence_number.cmp(&other.transaction.sequence_number) {
			Ordering::Equal => {}
			non_equal => return non_equal,
		}

		// If sequence number is equal, then compare by transaction on the whole
		self.transaction.cmp(&other.transaction)
	}
}
```
And, in `rocksdb`:
```rust
pub fn construct_mempool_transaction_key(transaction: &MempoolTransaction) -> String {
  // pad to 32 characters for slot_seconds
  let slot_seconds_str = format!("{:032}", transaction.timestamp);
  
  // pad to 32 characters for sequence number
  let sequence_number_str = format!("{:032}", transaction.transaction.sequence_number);
  
  // Assuming transaction.transaction.id() returns a hex string of length 32
  let transaction_id_hex = transaction.transaction.id(); // This should be a String of hex characters
  
  // Concatenate the two parts to form a 80-character hex string key
  let key = format!("{}:{}:{}", slot_seconds_str, sequence_number_str, transaction_id_hex);
  
  key
}

```
3. Updates `suzuka-full-node::partial` to use provide the `sequence_number` from an `aptos_transaction`:
```rust
let serialized_aptos_transaction = serde_json::to_vec(&transaction)?;
let movement_transaction = movement_types::Transaction {
  data : serialized_aptos_transaction,
  sequence_number : transaction.sequence_number()
};
let serialized_transaction = serde_json::to_vec(&movement_transaction)?;
transactions.push(BlobWrite { data: serialized_transaction });
```

## Conceptual Considerations
1. Should this work when there is more than one full-node sending transactions to the DA?
   - Mostly. Foremost, the client is responsible for sending the right sequence number, and the client should interact with only one full-node in quick succession. So, that full-node should be able to provide it the right sequence number repeatedly when queried. This would not work if the client decides to send transactions to different full-nodes, requesting different sequence numbers from each.m 

# Testing

1. Unit tests modified to match new type defintion.
2. Added new unit tests to `move-rocks` to check that new ordering conditions are satisifed.
3. e2

# Outstanding issues
1. None


musitdev (commented):
I have 2 main remarks:
 * the Tx are ordered by timestamp/seqnumber/tx_content. I think this ordering doesn't guaranty the sequence number order in the block. Nothing guaranty that the Tx will arrive in the order of the seq number attribution. Seq num are attributed by the client and between the attribution and when the Tx arrive to the mempool (when the ts is defined) they can be a lot of reordering.
 * The seq num ordering is needed only if there's 2 tx with the same sender in the block. We don't need to order all Tx by seq number. It's a very rare case that we've to manage.
For these reason, I think, changing the current ordering in the db with the key like it's done will be complex to do and at first it will be easier to reorder the Tx when the block is build if there's some Tx with the same sender.
We can put this ordering here https://github.com/movementlabsxyz/movement/blob/37b071ed8ba108d6e59dd18e0b76e1566d121454/protocol-units/mempool/util/src/lib.rs#L65, and we don't have to change how Tx are added to the mempool. Swapping Tx with incorrect seq number should work. I wonder how Aptos is doing, perhaps they have an easier or clever way to do.

Not related to this PR but as I look at this part of Suzuka node I think there are some inefficiencies: 
 * using json to serialized Tx raw data in the Db is not efficient. Currently, the received json Tx is deserialized then reserialized in json before written in the db. If we deserialize the json we should serialize it in a more compact and efficient way. Tx are mostly binary data.
 * the way the mempool Transaction key is constructed can be optimized.
Currently, with a few Tx it's not important, but if we want to process thousands Tx per block it will become important.

l-monninger:
> the Tx are ordered by timestamp/seqnumber/tx_content. I think this ordering doesn't guaranty the sequence number order in the block. Nothing guaranty that the Tx will arrive in the order of the seq number attribution. Seq num are attributed by the client and between the attribution and when the Tx arrive to the mempool (when the ts is defined) they can be a lot of reordering.

Not sure what you mean here. To clarify:

1. The `sequence_number` used here is the same as the one assigned by the client, having asked for the next sequence number via the API which reads the Mempool and `AptosDB`.
2. A block would be constructed at least once per slot (if there are available transactions). By this ordering, within the slot you would construct a block that is ordered by sequence number. If the client sends bad sequence numbers within a slot, then that should just be faulty logic on their end.

l-monninger:
> The seq num ordering is needed only if there's 2 tx with the same sender in the block. We don't need to order all Tx by seq number. It's a very rare case that we've to manage.
For these reason, I think, changing the current ordering in the db with the key like it's done will be complex to do and at first it will be easier to reorder the Tx when the block is build if there's some Tx with the same sender.
We can put this ordering here

1. Ordering all by sequence number also means that for a given account its transactions will be ordered by sequence number.
2.  (Re-)ordering the transaction when the block is built is essentially what's being done here. If you get two transactions that arrive out of order with respect to their sequence number in the same slot, they will be reordered according to their sequence number.

l-monninger:
> using json to serialized Tx raw data in the Db is not efficient. Currently, the received json Tx is deserialized then reserialized in json before written in the db. If we deserialize the json we should serialize it in a more compact and efficient way. Tx are mostly binary data.

1. Yes, we can consider a different format at some point. However, protobuf for the transaction data itself is kind of out of the cards until we break further from the upstream. `bcs` is probably the best bet in the short run.
2. I don't think there's a great way to avoid deserializing and the reserializing while preserving the API as we do currently.

l-monninger:
> the way the mempool Transaction key is constructed can be optimized.
Currently, with a few Tx it's not important, but if we want to process thousands Tx per block it will become important.

Do you mean this?

```
pub fn construct_mempool_transaction_key(transaction: &MempoolTransaction) -> String {
  // pad to 32 characters for slot_seconds
  let slot_seconds_str = format!("{:032}", transaction.timestamp);
  
  // pad to 32 characters for sequence number
  let sequence_number_str = format!("{:032}", transaction.transaction.sequence_number);
  
  // Assuming transaction.transaction.id() returns a hex string of length 32
  let transaction_id_hex = transaction.transaction.id(); // This should be a String of hex characters
  
  // Concatenate the two parts to form a 80-character hex string key
  let key = format!("{}:{}:{}", slot_seconds_str, sequence_number_str, transaction_id_hex);
  
  key
}
```

I assume you mean by allocating `[u8; 80]` of zeros and then setting array indices.

util/movement-types/src/lib.rs
```diff
@@ -22,34 +22,44 @@ impl Id {
...
+	pub data: Vec<u8>,
+	pub sequence_number: u64,
+}
 
 impl From<Vec<u8>> for Transaction {
 	fn from(data: Vec<u8>) -> Self {
-		Self(data)
+		Self { data, sequence_number: 0 }
 	}
 }
```
mzabaluev, line 47: This infallible conversion is a bit too risque to begin with, and now it's not clear that the sequence number 0 is the result of an `.into()`. I think it's better to remove this `From` impl and use the explicit constructor everywhere.

util/movement-types/src/lib.rs
```diff
@@ -22,34 +22,44 @@ impl Id {
...
 	fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
 		write!(f, "{:?}", &self.0)
 	}
 }
 
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct Transaction(pub Vec<u8>);
+pub struct Transaction {
+	pub data: Vec<u8>,
+	pub sequence_number: u64,
```
mzabaluev, line 40: Public struct members are not future-proof.
Consider making them private and providing a getter API; not necessarily in this PR though.

mzabaluev, line 40: The sequence number field could use some documentation: what are the uniqueness and monotonicity requirements?

mzabaluev:
A side note on the database keys: these are stringized to an ASCII format, which is less efficient (but more developer-friendly) than binary keys.

util/movement-types/src/lib.rs
l-monninger: Okay

util/movement-types/src/lib.rs
l-monninger: I mean, it should be easy enough to add in this PR.

l-monninger: Those would be defined per unit. For just the sequencer, it's basically viewed as merely another means for ordering. `maptos` requires consecutiveness. But, I wasn't planning on this container itself holding those opinions.

musitdev:
About the sequence number. The ordering is done first on the Tx ts it mean that a Tx with ts 1 and seq 2 will be before the tx with ts 2 and seq 1 which break the seq number order. Ts is attributed by the mempool and seq number is attributed by the client. So a Seq number attributed before another can arrive after the other. Seq1 is attributed before seq 2 but can arrive after seq2 like the example.
The Aptos client attribute himself sequence number by adding 1 each time a Tx is created. It doesn't query the node to get the last one. It's only done when the client is created.

protocol-units/mempool/util/src/lib.rs
```diff
@@ -122,7 +122,14 @@ impl Ord for MempoolTransaction {
...
-		// If slot_seconds are equal, then compare by transaction
+
+		// If slot_seconds are equal, then compare by sequence number
+		match self.transaction.sequence_number.cmp(&other.transaction.sequence_number) {
+			Ordering::Equal => {}
+			non_equal => return non_equal,
+		}
+
+		// If sequence number is equal, then compare by transaction on the whole
 		self.transaction.cmp(&other.transaction)
```
mzabaluev, line 133: This theoretically can compare the sequence numbers again, but in a very corner case. This last resort comparison is OK as a backstop to achieve a stable order without accounting for any new field we could add to the `Transaction` struct in the future.

But you could achieve the same effect implicitly using just this derived `PartialOrd` impl, if you put the `sequence_number` field first in the struct definition.

l-monninger: Will change.

musitdev:
`I don't think there's a great way to avoid deserializing and the reserializing while preserving the API as we do currently.` Here the tx is desirialized reserialized only because of the seqnumber: https://github.com/movementlabsxyz/movement/blob/37b071ed8ba108d6e59dd18e0b76e1566d121454/networks/suzuka/suzuka-full-node/src/partial.rs#L105 . If you order Tx by their arrival orders, you don't need to deserialize and reserialize it before adding it to the mempool.
You reorder Tx if needed when the block is created.
And if you need to deserialize the json, you can use a more optimized format to save it.

protocol-units/mempool/util/src/lib.rs
mzabaluev: To think of it, it might be better to not violate layering here and move the sequence number comparison into (explicitly defined and commented) `impl PartialOrd for Transaction`.

l-monninger:
> About the sequence number. The ordering is done first on the Tx ts it mean that a Tx with ts 1 and seq 2 will be before the tx with ts 2 and seq 1 which break the seq number order. Ts is attributed by the mempool and seq number is attributed by the client. So a Seq number attributed before another can arrive after the other. Seq1 is attributed before seq 2 but can arrive after seq2 like the example.

Yes, but as long as they arrive to the sequencer in the same slot, they will be correctly reordered. The edge case here would be when you send two right on the end of a slot. However, the current implementation safeguards against this a bit by trying to flush the mempool every 500ms (waiting to accumulate transactions).

There would be undecidability here. You can't know when to stop including transactions that might have certain sequence numbers; likewise the client can always send a transaction with an earlier sequence number. The Aptos mempool itself can't avoid this either.

l-monninger:
> If you order Tx by their arrival orders, you don't need to deserialize and reserialize it before adding it to the mempool.

This does not deserialize and reserialize the same data. It serializes the Aptos transaction and then wraps it as is needed for the sequencer. It then serializes the wrapped data.

```rust
let serialized_aptos_transaction = serde_json::to_vec(&transaction)?;
let movement_transaction = movement_types::Transaction {
data : serialized_aptos_transaction,
sequence_number : transaction.sequence_number()
};
let serialized_transaction = serde_json::to_vec(&movement_transaction)?;
transactions.push(BlobWrite { data: serialized_transaction });
```

The arrival order here is not reliable and further reordering would take place in the sequencer. So, I can't just FIFO it. That's the sequence number is passed down to the sequencer itself.

l-monninger:
So, @mzabaluev @musitdev, should I change to `bcs` for serialization.

musitdev:
`Yes, but as long as they arrive to the sequencer in the same slot, they will be correctly reordered. `: I didn't see the ts was defined in second. So it decreases the risk, but it's still there, as you say. The 500ms doesn't change anything because they are not locked to the second, so there's a lot of change that the 500ms end in the middle of the second and not at the end.
I can take some time to run the test with this commit to see if when I start 100 scenarios it works like with Aptos.

musitdev:
`The arrival order here is not reliable and further reordering would take place in the sequencer. ` I think it can be interesting to see how Aptos order its Tx. They have solved all these issues.

musitdev:
`So, @mzabaluev @musitdev, should I change to bcs for serialization.` yes why not but not now. We shouldn't do any other modification than correcting the sequence number. I think we should open an issue to do some bench between different possibilities and decide.

musitdev:
`I mean tomato, tomato here. What defines the client being "created" is up for interpretation. If you look at the usage of the API in aptos-core, they are fetching and refetching the sequence number fairly often: https://github.com/search?q=repo%3Amovementlabsxyz%2Faptos-core%20get_sequence_number&type=code`
In the faucet, the get_sequence_number is not always used to update the sequence number, but more to see if the account exist:
If you look at the transfer::fund process for example: 
In fund, the seq number is not used for the Tx: https://github.com/movementlabsxyz/aptos-core/blob/d29fa616e07040fc01447bc2d30455b80973bb8f/crates/aptos-faucet/core/src/funder/transfer.rs#L251
The `execute_transaction` call don't use it: https://github.com/movementlabsxyz/aptos-core/blob/d29fa616e07040fc01447bc2d30455b80973bb8f/crates/aptos-faucet/core/src/funder/transfer.rs#L281
And in execute_transaction, they call `sign_with_transaction_builder`  that increase the client seq number by one manually: https://github.com/movementlabsxyz/aptos-core/blob/d29fa616e07040fc01447bc2d30455b80973bb8f/sdk/src/types.rs#L172 or here https://github.com/movementlabsxyz/aptos-core/blob/d29fa616e07040fc01447bc2d30455b80973bb8f/sdk/src/types.rs#L339 .
So the faucet client seq number is not synchronized with the node; It's what I see from my test, the faucet get desynchronized when the seq number error arrive and the node discard the Tx. After all fund request fail. So one error of seq number in tx execution (they are not managed on client side) can desynchronize the client and break it.

l-monninger:
@musitdev Would you mind approving?

musitdev (approved):
You should wait the @0xmovses test before merging.

protocol-units/mempool/util/src/lib.rs
mzabaluev: > Instead of `Ord`?

Oh yes, both. Normally you just proxy `Ord` to `PartialOrd` which in fact implements a total order.
Or just derive both, but put the `sequence_number` member first in the struct.

util/movement-types/src/lib.rs
```diff
@@ -40,12 +40,6 @@ pub struct Transaction {
 	pub sequence_number: u64,
 }
 
-impl From<Vec<u8>> for Transaction {
-	fn from(data: Vec<u8>) -> Self {
-		Self { data, sequence_number: 0 }
-	}
-}
-
```
mzabaluev, line 39: OK.

l-monninger:
Lol, how in the world is the behavior so different in CI.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -200,14 +200,13 @@ impl LightNodeService for LightNodeV1 {
...
-			.map(|blob| {
-				let transaction = Transaction::from(blob.data);
-				transaction
-			})
-			.collect();
-
+		let mut transactions = Vec::new();
+		for blob in blobs_for_submission {
+			let transaction : Transaction = serde_json::from_slice(&blob.data)
+				.map_err(|e| tonic::Status::internal(e.to_string()))?;
```
mzabaluev, line 206: This added layer of JSON formatting in what used to be a blob bothers me a little, too.
Can we work the sequence number into the gRPC message instead, and keep the transaction data as a separate `bytes` field?

l-monninger: That's what I was asking about the other day. We could do that. Given the hour, it may be better to do it in a separate PR after CI issues are sorted here. That way we can push this to update the devnet.

mzabaluev: Filed as #182.

PR #181, opened 2024-07-10
Fix up MCR integration

mzabaluev:
# Summary
- **Categories**: `networks`, `misc`.

Fixes in suzuka-full-node after MCR settlement integration
was reinstated.

# Testing

The existing e2e tests should pass (if they do in `main`).


l-monninger (approved):
Looks good to me! In case you are wondering about other uses of `Future::pending`, we do need to keep it for the formerly "setup" now "runner" binary in a forthcoming PR.

PR #189, opened 2024-07-11
Faucet Fix with Docker Containers

l-monninger:
# Summary
- **RFCs**: $\emptyset$
- **Categories**: `networks`

Restores Docker Containers viz. the faucet fix.

# Changelog

# Testing

1. e2e `native` and `docker-compose` tests.

# Outstanding issues


networks/suzuka/setup/src/local.rs
```diff
@@ -29,10 +29,14 @@ impl Local {
 		// Update the config with the new m1_da_light_node_config
 		config.m1_da_light_node = new_m1_da_light_node_config;
 
-		tracing::info!("Running mcr_settlement_setup");
+		/*tracing::info!("Running mcr_settlement_setup");
 		let mcr_settlement_config: mcr_settlement_config::Config = config.mcr.clone();
 		let (mcr_config, join_handle) = self.mcr_settlement_strategy.setup(&dot_movement, mcr_settlement_config).await?;
-		config.mcr = mcr_config;
+		config.mcr = mcr_config;*/
```
mzabaluev, line 35: Is the plan to merge this before #191?

networks/suzuka/suzuka-full-node/src/manager.rs
```diff
@@ -27,6 +29,27 @@ impl Manager<SuzukaPartialNode<Executor>> {
     }
 
     pub async fn try_run(&self) -> Result<(), anyhow::Error> {
+
+        let (stop_tx, mut stop_rx) = tokio::sync::watch::channel(());
+        tokio::spawn({
+            let mut sigterm = signal(SignalKind::terminate()).context("Can't register to SIGTERM.")?;
+            let mut sigint = signal(SignalKind::interrupt()).context("Can't register to SIGKILL.")?;
+            let mut sigquit = signal(SignalKind::quit()).context("Can't register to SIGKILL.")?;
```
mzabaluev, line 37: I could offer the same comments here as with #191, but let's address this in one place.

protocol-units/settlement/mcr/client/src/lib.rs
```diff
@@ -4,14 +4,16 @@ use tokio_stream::Stream;
...
 pub mod tests;
 
-#[cfg(feature = "mock")]
 pub mod mock;
 
-// #[cfg(feature = "eth")]
-// pub mod eth;
-#[cfg(feature = "eth")]
+#[cfg(feature = "mock")]
+pub use mock::*;
```
mzabaluev, line 10: This is also shared with #191, it would be better to stage these branches for incremental reviews.

networks/suzuka/setup/src/local.rs
l-monninger: Yes

networks/suzuka/suzuka-full-node/src/manager.rs
l-monninger: Will do.

protocol-units/settlement/mcr/client/src/lib.rs
l-monninger: Will do.

networks/suzuka/setup/src/local.rs
```diff
@@ -6,6 +6,7 @@ use dot_movement::DotMovement;
 
 #[derive(Debug, Clone, Default)]
 pub struct Local {
+	// this is just for the current version without settlement because this field is never read.
```
mzabaluev, line 9: Might just slap `#[allow(dead_code)]` on it and keep it private, so we can grep for these warnings overrides later.

mzabaluev (approved):
Good to go, rest of the small issues to be resolved in #191

PR #191, opened 2024-07-12
Restore Settlement

l-monninger:
# Summary
- **RFCs**: $\emptyset$
- **Categories**: `networks`

Restores settlement and async setup management (start all services--including setup--together). 

# Changelog

# Testing

# Outstanding issues


mzabaluev (commented):
Generally looks good, I have got some minor concerns with error messages and the like.

networks/suzuka/setup/src/main.rs
```diff
@@ -15,6 +19,27 @@ async fn main() -> Result<(), anyhow::Error> {
...
+					_ = sigterm.recv() => (),
+					_ = sigint.recv() => (),
+					_ = sigquit.recv() => (),
+				};
+				tracing::info!("Receive Terminate Signal");
+				if let Err(err) = stop_tx.send(()) {
+					tracing::warn!("Can't update stop watch channel because :{err}");
+					return Err::<(), anyhow::Error>(anyhow::anyhow!(err));
+				}
+			}
```
mzabaluev, line 39: This does its job and it's OK, just seems a bit overcomplicated for what it needs to do. A future that does this `select!` once should be enough, and you can await it directly rather than use a watch channel, I think.

mzabaluev (approved):
OK. The error messages on signal installation are still wrong, but we can always fix them later if it ever occurs at all.

l-monninger:
> OK. The error messages on signal installation are still wrong, but we can always fix them later if it ever occurs at all.
I'll address my bad.

PR #192, opened 2024-07-12
fix bridge package publishing issue: init_module takes only 1 signer

andygolay:
# Summary

Bridge move modules package was compiling and passing tests but not publishing. It was because `init_module` in `atomic_bridge::atomic_bridge_counterparty` can only take 1 arg. Module tests were not exposing the error for some reason. 

# Changelog

I modified `init_module` and related tests to only take one signer. This change makes it so `moveth_minter` must have the same address as the module publisher. If I understand correctly, this will be okay to do because the plan is to implement a resource account pattern where the module is published under the resource account and the resource account is the minter. I may be misunderstanding that, so please review and adjust accordingly, @0xmovses.

# Testing

Unit tests in module pass,  module publishes successfully.

One way to publish the module is change `Move.toml` as follows:

```
[addresses]
atomic_bridge = "<your-address>"
moveth = "<your-address>"
master_minter = "0xbab"
minter = "0xface"
pauser = "0xdafe"
denylister = "0xcade"

# [dev-addresses]
# moveth = "0xbeef"
# atomic_bridge = "0xfeef"
# master_minter = "0xbab"
# minter = "0xface"
# pauser = "0xdafe"
# denylister = "0xcade"
```

And publish to Aptos devnet. (As far as I know, Movement does not have a network up yet that includes `function_info` and `dispatchable_fungible_asset` modules in the framework.)

# Outstanding issues
I may be incorrect in assuming it's okay to force the moveth minter to be the same account as the module publisher. My solution is proposed as one way to make the module compile but may need some modification to make sense in the system design.

### DIFF
--- a/protocol-units/bridge/move-modules/sources/atomic_bridge.move
+++ b/protocol-units/bridge/move-modules/sources/atomic_bridge.move
@@ -49,21 +49,21 @@ module atomic_bridge::atomic_bridge_counterparty {
     struct BridgeTransferCancelledEvent has store, drop {
         bridge_transfer_id: vector<u8>,
     }
-
-    entry fun init_module(deployer: &signer, moveth_minter: &signer) {
+    
+    entry fun init_module(deployer: &signer) {
         let bridge_transfer_store = BridgeTransferStore {
             pending_transfers: smart_table::new(),
             completed_transfers: smart_table::new(),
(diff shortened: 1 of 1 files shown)


l-monninger (approved):
LGTM

PR #195, opened 2024-07-12
MOVETH module mint resource account refactor

andygolay:
# Summary

Refactored `moveth` module for resource account. Tests still need to be rewritten as they were from the old minting model.

# Changelog

- Module is published under resource account. 
- In `init_module` the `Roles`, `Management`, and `State` resources are moved to the resource account address.  
- Changed `Move.toml` so `atomic_bridge`, `moveth`, `master_minter`, and `resource_addr` take the resource account address, and `minter` takes the address of whoever published the module, for convenience.

So for example in this case to publish I'm running

```
aptos move create-resource-account-and-publish-package --seed 0123456789 --address-name moveth --profile default --named-addresses source_addr=0x99c117835788d6310c09ba4c88b3f81058b9b1684d74bc62f2bbbd7b06ed8d79
```

and to mint I'm running

```
aptos move run --function-id 0xc26f08a91754150c6c8ab9a79f51d1d23bf70828e8fd62b6ea49ec2a9bf95cfa::moveth::mint  --args address:0xb118a531aa5efb194f5e322aeee8e9d1e98c6220936d1cee90e1480911d51d64 u64:5
``` 

# Testing

I'll rewrite the tests later today. Right now they fail, and it appears that the `mint` function is failing at some point.

# Outstanding issues
It appears that the mint function is not working, despite not returning errors. Will rewrite the tests for the resource account model later today.

I could use feedback on the general logical design and suggestions as to why the mint function fails without errors.

### DIFF
--- a/protocol-units/bridge/move-modules/Move.toml
+++ b/protocol-units/bridge/move-modules/Move.toml
@@ -4,10 +4,11 @@ version = "1.0.0"
 authors = []
 
 [addresses]
-atomic_bridge = "0xd704dffbd878b8928d72c4a5db2eb326310bcb2f2291aa541f40b86853f90fa6"
-moveth = "0xd704dffbd878b8928d72c4a5db2eb326310bcb2f2291aa541f40b86853f90fa6"
-master_minter = "0xfeef"
-minter = "0xbab"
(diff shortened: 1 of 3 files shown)


andygolay:
Closing this in favor of https://github.com/movementlabsxyz/movement/pull/205 (got a fresh start and came up with a more elegant solution).

PR #198, opened 2024-07-14
Remove unused dependencies

c410-f3r:
Some were distributed for production but only used in testing environments


c410-f3r:
Closing due to the lack of feedback

PR #201, opened 2024-07-15
Sequence Number Panic

l-monninger:
# Summary
- **RFCs**: $\emptyset$
- **Categories**: `protocol-units`

Updates sequence number handling in mempool to not panic executor thread. 

# Changelog

1. Introduces `thiserror` enum for `transaction_pipe`.
2. Catches `MempoolStatus::NotAccepted` in `dof-executor`. 

# Testing

1. e2e tests. 

# Outstanding issues
1. Error types should be introduced elsewhere.

PR #202, opened 2024-07-15
add Tx pre execution when submited

musitdev:
# Summary

When executing the malicious Test, the Suzuka behavior was different from the Aptos node. Some Tx issues were not detected during submission and the client was waiting indefinitely because the execution error weren't seen.
This PR add Tx pre execution when the Tx is submitted.
From the test I've done, the node behavior is the same as the Aptos node.
For example: 
 * INVALID_AUTH_KEY and BAD_CHAIN_ID return during the Tx submission.
 * INSUFFICIENT_BALANCE and WRONG ARGUMENTS return during the Tx waiting.

# Changelog

# Testing

To run the Suzuka node use:
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just suzuka-full-node native build.setup.local --keep-tui"
```
Malicious Test can be executed using the branch `eng-442/malicious_load_soak_test`

# Outstanding issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7145,6 +7145,7 @@ dependencies = [
  "aptos-vm-genesis",
  "aptos-vm-logging",
  "aptos-vm-types",
+ "aptos-vm-validator",
  "async-channel 2.3.1",
  "async-trait",
  "bcs 0.1.4",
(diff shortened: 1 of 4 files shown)


l-monninger (approved):
We already discussed this. It looks correct to me, though we will want to make sure we resolve any merge conflicts to respect: https://github.com/movementlabsxyz/movement/pull/201

PR #205, opened 2024-07-16
Moveth mint resource account refactor

andygolay:
# Summary
Bridge moveth module can now be published under resource account. 

# Changelog

- Added admin role to allow adding minters
- Configured addresses in Move.toml as an example of how addresses could be set up for resource account
- Modified the `complete_bridge_transfer` function so anyone can call it. The logic of this should be checked against concrete bridge implementation.

# Testing

- Modified unit tests in module to use admin role for adding minter. All tests pass.
- Tested publishing module under resource account and minting using CLI. 

To reproduce publishing under resource account:

## 1. Initialize `movement`

```
movement init
```

Press Return at each step to accept default config settings.

## 2. Publish package under resource account. Modify `Move.toml` as you go.

```
aptos move create-resource-account-and-publish-package --address-name default --seed 1234
```

## 3. Interact with package

To reproduce minting (assuming source address `0xe4...` and resource account address `0x6a...`; must be called from a minter address):
```
aptos move run --function-id 0x6a05dca724278092fb96a8ff91e1b814092e4e2059e1769793cda8d48f199951::moveth::mint --args address:0xe4442499a19ba92b5901cb059bad46eb15c88ebf99bc4d18f7f17e8beea4a4b2 u64:5
```

# Outstanding issues
Design or address configuration may need to be tweaked for concrete implementation with the other bridge components. More thorough testing of various module functions from CLI could be conducted.

This is tested against Aptos testnet as Suzuka is under maintenance.


andygolay:
@0xmovses all unit tests now pass. The way I have `Move.toml` set up is 

```
[package]
name = "bridge-modules"
version = "1.0.0"
authors = []

[addresses]
atomic_bridge = "<resource-account-address>"
moveth = "<resource-account-address>"
master_minter = "<resource-account-address>"
minter = "<source-address>"
admin = "<source-address>"
pauser = "0xdafe"
denylister = "0xcade"

[dev-addresses]
# moveth = "0xbeef"
# atomic_bridge = "0xfeef"
# master_minter = "0xbab"
# minter = "0xface"
# admin = "0xad"
# pauser = "_"
# denylister = "_"
```

protocol-units/bridge/move-modules/Move.toml
```diff
@@ -4,21 +4,24 @@ version = "1.0.0"
...
-[dev-addresses]
-moveth = "0xbeef"
-atomic_bridge = "0xfeef"
-master_minter = "0xbab"
-minter = "0xface"
+atomic_bridge = "0x6a05dca724278092fb96a8ff91e1b814092e4e2059e1769793cda8d48f199951"
+moveth = "0x6a05dca724278092fb96a8ff91e1b814092e4e2059e1769793cda8d48f199951"
+master_minter = "0x6a05dca724278092fb96a8ff91e1b814092e4e2059e1769793cda8d48f199951"
+minter = "0xe4442499a19ba92b5901cb059bad46eb15c88ebf99bc4d18f7f17e8beea4a4b2"
+admin = "0xe4442499a19ba92b5901cb059bad46eb15c88ebf99bc4d18f7f17e8beea4a4b2"
```
0xmovses, line 11: Where do these addresses come from?

0xmovses (commented):
I `cd protocol-units/bridge/move-modules` then run
```
aptos move create-resource-account-and-publish-package --address-nam
e default --seed 1234
```
I get :
{
  "Error": "Unable to find config /Users/movses/.aptos/config.yaml, have you run `aptos init`?"
}

Is there a some step I'm missing for testing?

andygolay:
@0xmovses I removed the misleading `mint_to` address which I had put in as kind of a kludge to solve the  error I was getting when adding a minter. To fix the `EALREADY_MINTER` error I commented out the redundant `add_minter` call in `test_complete_bridge_transfer`. It seems to me that test of client minting some MovETH to themselves is not neccesary.

0xmovses:
I still get the error above, I think there is a step missing in the testing instructions https://github.com/movementlabsxyz/movement/pull/205#pullrequestreview-2179996421 let me know.

andygolay:
> I still get the error above, I think there is a step missing in the testing instructions [#205 (review)](https://github.com/movementlabsxyz/movement/pull/205#pullrequestreview-2179996421) let me know.

You ran `aptos init` and it is still giving you that same "Unable to find config" error?

0xmovses:
No I didn't run aptos init. Wasn't in the testing instructions, I'll try that. I think we should add some CI job that runs these tests, as we had a change in the MCR contracts that broke the settlment last week. Adding a job to CI that runs the tests would prevent this from happening.

0xmovses:
So here's what I did 
```
⋊> ~/m/m/p/b/move-modules on andygolay/moveth_mint_resource_account_refactor-2 ⨯ aptos init                                                                   15:45:44
Configuring for profile default
Choose network from [devnet, testnet, mainnet, local, custom | defaults to devnet]

No network given, using devnet...
Enter your private key as a hex literal (0x...) [Current: None | No input: Generate new key (or keep one if present)]

No key given, generating key...
Account 0x58.. doesn't exist, creating it and funding it with 100000000 Octas
Account 0x58.. funded successfully

---
Aptos CLI is now set up for account 0x58.. as profile default!  Run `aptos --help` for more information about commands
{
  "Result": "Success"
}
```

Then I ran the test command
`aptos move create-resource-account-and-publish-package --address-name default --seed 1234`

And I got this output 

```
Do you want to publish this package under the resource account's address 0x6c18b5c30971d72c20be09b366711e01d95a56e47257a8cf4061221513a961ae? [yes/no] >
y
package size 12042 bytes
{
  "Error": "Simulation failed with status: CONSTRAINT_NOT_SATISFIED\nExecution failed with message: metadata and code bundle mismatch: unregistered dependency: '6a05dca724278092fb96a8ff91e1b814092e4e2059e1769793cda8d48f199951::moveth'"
}
```

Should I manually updating my `Move.toml` with the `0x58.. ` account generated?

andygolay:
> 0x6c18b5c30971d72c20be09b366711e01d95a56e47257a8cf4061221513a961ae

Yeah exactly. Each time I test it from CLI, I'll create a different seed and then manually update the addresses in `Move.toml`.

So if `0x6c18b5c30971d72c20be09b366711e01d95a56e47257a8cf4061221513a961ae` is the resource account address for the seed you chose, then that can be the address for `atomic_bridge`, `moveth`, and `master_minter` as that is all under the resource account.

The `0x58...` address created via `aptos init` is for the source account. I set it up to be the address for `minter` and `admin`. `admin` might be optional but it creates the possibility of some conveniences for deployers so I left it in for now.

0xmovses:
Ok cool I'll try that, in the meantime I created this issue, so that we can test move modules in CI with automation. Something for later https://github.com/movementlabsxyz/movement/issues/211

protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -222,7 +227,8 @@ module atomic_bridge::atomic_bridge_counterparty {
        complete_bridge_transfer(
```
0xmovses, line 227: Question about this comment (which I previously wrote)
`// Client must be a moveth minter, otherwise this will fail`, this is no longer true now right?

Would be great if you could write another test, that explicitly calls `complete_bridge_transfer` from a non-moveth minter, to prove this all works

0xmovses (changes requested):

protocol-units/bridge/move-modules/sources/atomic_bridge.move
andygolay: Right, removed that comment.

Because `complete_bridge_transfer` includes this:

```
        // Make caller a minter of MovETH
        moveth::add_minter(master_minter, signer::address_of(caller));
```

that means the test `test_complete_transfer_assets` uses a non-moveth minter now.

I removed the line in `test_complete_transfer_assets` where the non-moveth minter is made a minter, before calling `complete_bridge_transfer` so now `test_complete_transfer_assets` can be 

The `client` in `test_complete_transfer_assets` becomes `caller` in `complete_bridge_transfer` and gets made into a minter pending reveling the hashlock preimage.

0xmovses:
:shipit:

PR #206, opened 2024-07-16
Faucet Fix w/ Docker Containers

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Includes changes from #201 

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/networks/suzuka/suzuka-full-node/src/manager.rs
+++ b/networks/suzuka/suzuka-full-node/src/manager.rs
@@ -7,6 +7,8 @@ use godfig::{
 };
 use suzuka_config::Config;
 use maptos_dof_execution::v1::Executor;
+use tokio::signal::unix::signal;
+use tokio::signal::unix::SignalKind;
 
 #[derive(Clone)]
 pub struct Manager<Dof>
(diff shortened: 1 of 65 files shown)


networks/suzuka/suzuka-full-node/src/manager.rs
```diff
@@ -27,6 +29,25 @@ impl Manager<SuzukaPartialNode<Executor>> {
     }
 
     pub async fn try_run(&self) -> Result<(), anyhow::Error> {
+
+        let (stop_tx, mut stop_rx) = tokio::sync::watch::channel(());
+        tokio::spawn({
+            let mut sigterm = signal(SignalKind::terminate()).context("Can't register to SIGTERM.")?;
+            let mut sigint = signal(SignalKind::interrupt()).context("Can't register to SIGKILL.")?;
```
mzabaluev, line 36:
```suggestion
            let mut sigint = signal(SignalKind::interrupt()).context("Can't register to SIGINT.")?;
```

networks/suzuka/suzuka-full-node/src/manager.rs
```diff
@@ -27,6 +29,25 @@ impl Manager<SuzukaPartialNode<Executor>> {
...
+            async move {
+                loop {
+                    tokio::select! {
+                        _ = sigterm.recv() => (),
+                        _ = sigint.recv() => (),
+                    };
```
mzabaluev, line 42: And again, this whole business could be replaced by a future doing a single `select!` because we're not interested in polling for more termination signals after receiving one.

protocol-units/settlement/mcr/client/src/lib.rs
```diff
@@ -4,14 +4,16 @@ use tokio_stream::Stream;
...
 pub mod tests;
 
-#[cfg(feature = "mock")]
 pub mod mock;
 
-// #[cfg(feature = "eth")]
-// pub mod eth;
-#[cfg(feature = "eth")]
+#[cfg(feature = "mock")]
+pub use mock::*;
```
mzabaluev, line 10: The mock can still be imported from top level with nothing in the name indicating that it's a mock: `use mcr_settlement_client::McrSettlementClient`. I think this is confusing and this glob re-export better be removed.

mzabaluev (approved):
The improvements look good, though some past remarks still apply.

networks/suzuka/suzuka-full-node/src/manager.rs
mzabaluev: Remember that `Signal::recv` is not special in the way low-level `signal` callbacks are: it's called in a normal application thread and already uses some synchronization magic under the hood that's triggered by the actual callbacks installed by Tokio.

l-monninger: Yes, that's what I assumed too. I just don't know how to explain why this works reliable and the setup you describe does not yet. So, let's save this for a future PR.

networks/suzuka/suzuka-full-node/src/manager.rs
l-monninger: This is resolved in https://github.com/movementlabsxyz/movement/pull/191. I'll leave it to that.

protocol-units/settlement/mcr/client/src/lib.rs
l-monninger: Okay, I'll make this change.

l-monninger: I think I'll put in the #191  though so that we can update the application.

PR #208, opened 2024-07-16
Opt executor test fixes, code cleanup

mzabaluev:
# Summary

In `test_pipe_mempool_with_malformed_transaction`, check the transaction submission status that the transaction is pre-validated.
This makes the test fail rather than hang up.

Also commit some drive-by code cleanups.

# Testing

`cargo check -p maptos-opt-executor`

# Outstanding issues

The test is still broken, we need a setup that passes the validator with a randomly generated transaction.


protocol-units/execution/opt-executor/src/executor/initialization.rs
```diff
@@ -167,7 +167,7 @@ impl Executor {
 		// replace the db path with a temporary directory
 		let value = tempfile::tempdir()?.into_path(); // todo: this works because it's at the top level, but won't be cleaned up automatically
 		maptos_config.chain.maptos_db_path.replace(value);
+		dbg!(&maptos_config.chain.maptos_db_path);
```
l-monninger, line 170: Remember to remove this, unless you have a good reason to keep it.

l-monninger (changes requested):

protocol-units/execution/opt-executor/src/executor/initialization.rs
mzabaluev: Done in the subsequent commit.

protocol-units/execution/opt-executor/src/executor/services.rs
```diff
@@ -35,49 +35,43 @@ impl Executor {
...
+		let (mut executor, _tempdir) = Executor::try_test_default(GENESIS_KEYPAIR.0.clone())?;
 		let server_executor = executor.clone();
 
 		let handle = tokio::spawn(async move {
 			server_executor.run_service().await?;
 			Ok(()) as Result<(), anyhow::Error>
 		});
 
-		let user_transaction = create_signed_transaction(0, executor.maptos_config.chain.maptos_chain_id.clone());
+		let user_transaction = create_signed_transaction(1, &executor.maptos_config);
```
l-monninger, line 74: Do we actually need `1` here?

l-monninger (approved):
LGTM

protocol-units/execution/opt-executor/src/executor/services.rs
mzabaluev: No, 0 works just as well. I have changed this helper function for consistency with the one in the sibling module `transaction_pipe`; perhaps a shared test util module would be better, but we may need different transaction data for different tests, too.

l-monninger: Shared is better ofc. But, don't worry about it. Just wanted to clarify conceptually.

PR #209, opened 2024-07-16
Integrate top level services and e2e test events

0xmovses:
# Summary
Adds Eth Event Logging and necessary logic for implementing the top level servie

# Changelog

Traits implemented for Eth Event Logging

# Testing
- No new tests have been added as the top level `BridgeService` that supports both Networks needs to be integrated. Follow up PR Incoming


andygolay:
@0xmovses this is not yet ready for review, right?

0xmovses:
@andygolay Correct. I've de-prioritised it for the time being while I get the contracts for movement -> eth bridge first, as we need that for the bridge service to function. Then will come back to this 👍

PR #217, opened 2024-07-17
Improvements to complete_bridge_transfer flow, additional test added

andygolay:
# Summary

Made it so minter role is removed after minting, so there doesn't accumulate an unneeded list of minters

# Changelog

* Added `remove_minter` function to `moveth` module
* Call `remove_minter` in `complete_bridge_transfer` after minting 
* Wrote test for someone already a minter calling `complete_bridge_transfer` (expected failure)

# Testing

All unit tests pass.

# Outstanding issues
Should be integration-tested with full concrete bridge implementation as we put it all together.

### DIFF
--- a/protocol-units/bridge/move-modules/sources/atomic_bridge.move
+++ b/protocol-units/bridge/move-modules/sources/atomic_bridge.move
@@ -184,15 +184,6 @@ module atomic_bridge::atomic_bridge_counterparty {
         let recipient = @0xface; 
         let asset = moveth::metadata();
 
-        // the master minter sets client to be a minter
-        // moveth::add_minter(master_minter, signer::address_of(client));
-
-        //client now mints themselves 100 moveth
-        // moveth::mint(client, signer::address_of(client), 100);
-        // assert!(primary_fungible_store::balance(signer::address_of(client), asset) == 100, 0);
-
-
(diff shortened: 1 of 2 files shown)


protocol-units/bridge/move-modules/sources/atomic_bridge.move
```diff
@@ -167,7 +170,82 @@ module atomic_bridge::atomic_bridge_counterparty {
...
+        let recipient = @0xface; 
+        let asset = moveth::metadata();
+
+        // the master minter sets client to be a minter
+        // moveth::add_minter(master_minter, signer::address_of(client));
+
+        //client now mints themselves 100 moveth
+        // moveth::mint(client, signer::address_of(client), 100);
+        // assert!(primary_fungible_store::balance(signer::address_of(client), asset) == 100, 0);
+
```
0xmovses, line 193: Let's remove these commented out lines above from 187 - 192.

0xmovses (changes requested):
Good call, one clean up note.

protocol-units/bridge/move-modules/sources/atomic_bridge.move
andygolay: Ah yep removed the unneeded comments from both tests.

0xmovses (approved):
:shipit:

PR #219, opened 2024-07-17
Update settlement test

musitdev:
# Summary

Add the MCR Settlement client test updated in PR #105 .

First step of the migration of PR #105 in main.

# Changelog

Add the settlement client test to current genesis ceremonial.
Change the sent Tx insufficient gas error management to add new case.
Activate the Test in the CI.

# Testing

To test, remove the `.movement` folder and run the command:
`CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just mcr-client native build.local.test --keep-tui"`

# Outstanding issues


protocol-units/settlement/mcr/config/src/common/testing.rs
```diff
@@ -38,13 +35,17 @@ env_short_default!(
...
+// env_or_none!(
+//     default_maybe_testing,
+//     Config,
+//     default_mcr_testing_admin_account_private_key,
+//     default_move_token_contract_address,
+//     default_movement_staking_contract_address
+// );
+
+pub fn maybe_testing() -> Option<Config> {
+    std::env::var("MAYBE_TESTING_MCR").ok().map(|_| Config::default())
```
mzabaluev, line 47: What's the meaning of `MAYBE_TESTING_`?

musitdev: I copy the pattern from the deploy config. It's the way @l-monninger define how to activate optional config part.

protocol-units/settlement/mcr/client/src/send_eth_transaction.rs
```diff
@@ -113,9 +113,12 @@ pub async fn send_transaction<
...
 				);
-				if transaction_receipt.gas_used == estimate_gas {
-					tracing::warn!("Send commitment Transaction  fail because of insufficient gas, receipt:{transaction_receipt:?} ");
-					estimate_gas += (estimate_gas * 10) / 100;
+				// Some valid Tx can abort cause of insufficient gas without consuming all its gas.
+				// Define a threshold a little less than estimated gas to detect them.
+				let tx_gas_consumption_threshold = estimate_gas - (estimate_gas * 10) / 100;
+				if transaction_receipt.gas_used >= tx_gas_consumption_threshold {
+					tracing::info!("Send commitment Transaction  fail because of insufficient gas, receipt:{transaction_receipt:?} ");
+					estimate_gas += (estimate_gas * 30) / 100;
```
mzabaluev, line 121: There is `ReceiptEnvelope::cumulative_gas_used`, I wonder if that gives the true amount.

protocol-units/settlement/mcr/setup/src/deploy/mod.rs
```diff
@@ -144,39 +125,40 @@ impl Deploy {
...
+		// 	if let Some(existing_testing_config) = config.testing.clone() {
+		// 		existing_testing_config.well_known_account_private_keys
+		// 	} else {
+		// 		let mut keys = Vec::new();
+		// 		for _ in 0..10 {
+		// 			let wallet = PrivateKeySigner::random();
+		// 			keys.push(wallet.to_bytes().to_string());
+		// 		}
+		// 		keys
+		// 	};
```
mzabaluev, line 145: Is this meant to be switched off temporarily?

protocol-units/settlement/mcr/setup/src/local/mod.rs
```diff
@@ -29,32 +29,37 @@ impl Local {
...
 		path.push("anvil.json");
 
-		let exists = tokio::fs::try_exists(&path).await.context("Failed to check if Anvil file exists")?;
+		let exists = tokio::fs::try_exists(&path)
+			.await
+			.context("Failed to check if Anvil file exists")?;
 
 		info!("Anvil path: {:?}", path);
-		info!{"Anvil exists: {:?}", exists};
+		info! {"Anvil exists: {:?}", exists};
```
mzabaluev, line 56: A nicer printout in one message, without debug cruft and unusual macro syntax:

```rust
		info!("Anvil path: `{}`, {}", path.display(), if exists { "exists" } else { "does not exist" });
```

protocol-units/settlement/mcr/setup/src/local/mod.rs
```diff
@@ -86,19 +91,23 @@ impl Local {
...
-			deploy.mcr_deployment_account_private_key = anvil_addresses.get(0).ok_or(anyhow!("Failed to get Anvil address"))?.private_key.clone();
+			deploy.mcr_deployment_account_private_key = anvil_addresses
+				.get(0)
+				.ok_or(anyhow!("Failed to get Anvil address"))?
+				.private_key
+				.clone();
 		}
 		if let Some(testing) = &mut config.testing {
+			// Remove the old one if the config was existing.
+			testing.well_known_account_private_keys.clear();
```
mzabaluev, line 105: Wouldn't it be surprising for someone who pre-adds some keys to the config, to see them not present for the test run?

protocol-units/settlement/mcr/client/src/send_eth_transaction.rs
musitdev: No clear what is it. It seems to be the cumulative gas use by the client, I suppose. Here, I try to estimate how much gas is needed to execute the Tx. With the submit commitment call, Eth doesn't estimate correctly the gas needed. That why I retry by increasing the gas until it pass.

protocol-units/settlement/mcr/setup/src/deploy/mod.rs
musitdev: I should be removed. I keep it because it's @l-monninger code and I change the behavior. I would like to discuss with him if it's ok.

protocol-units/settlement/mcr/setup/src/local/mod.rs
musitdev: In fact, I add the case when I've started several time process-compose without removing the .movement folder. If the config file already exists, without the clear, we add the new execution keys after the old one.

protocol-units/settlement/mcr/setup/src/local/mod.rs
musitdev: Change done.

protocol-units/settlement/mcr/setup/src/local/mod.rs
mzabaluev: This sounds messy to me. Should we at least have a flag so this behavior is explicitly enabled?

mzabaluev: Also, if we have a test that reuses the setup, this should be a special case API in setup that does not get entangled with the normal case of running once on a clean `.movement`?

mzabaluev: I think I'm just confused by this code. `config` is passed to this function, presumably read from some configuration on file in the test setup. The configuration is further modified by setup, then returned, to be used as the effective configuration by the test. That returned config should never be passed again to `Local::setup`, instead the original config source should be used.

mzabaluev (changes requested):

protocol-units/settlement/mcr/setup/src/local/mod.rs
mzabaluev: OK, this is arguing about minor things and practically, the list of private keys should always be in sync with `anvil.json`, which this achieves.

mzabaluev (approved):

PR #220, opened 2024-07-17
Update aptos-core, use 0 as the reward percentage

mzabaluev:
# Summary

After https://github.com/movementlabsxyz/aptos-core/pull/45, the `rewards_apy_percentage` field in the genesis configuration is used to calculate the rewards without Movement-specific hacks, but this means it has to be set to 0 for our purposes.

# Changelog

* In the Aptos genesis configuration, the `rewards_apy_percentage` parameter is set to 0 to represent the reality of not using rewards on Aptos blockchain level.

# Testing

Currently run tests should pass.

### DIFF
--- a/protocol-units/execution/opt-executor/src/executor/initialization.rs
+++ b/protocol-units/execution/opt-executor/src/executor/initialization.rs
@@ -58,7 +58,7 @@ impl Executor {
 				max_stake: 100_000_000_000_000,
 				recurring_lockup_duration_secs: EPOCH_DURATION_SECS * 2,
 				required_proposer_stake: 0,
-				rewards_apy_percentage: 10,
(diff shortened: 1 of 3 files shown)


protocol-units/execution/opt-executor/src/executor/initialization.rs
```diff
@@ -58,7 +58,7 @@ impl Executor {
 				max_stake: 100_000_000_000_000,
 				recurring_lockup_duration_secs: EPOCH_DURATION_SECS * 2,
 				required_proposer_stake: 0,
-				rewards_apy_percentage: 10,
+				rewards_apy_percentage: 0,
```
l-monninger, line 61: As advertised.

l-monninger (approved):

PR #228, opened 2024-07-18
correct spelling atomic_bridge.move

cratiu222:
Hello
I fixed some minor spelling issues.
Hope it helps.


andygolay:
Hi @cratiu222, thank you but "cancelling" is correct spelling in many geographic locales so don't think it needs a change in this case.

PR #230, opened 2024-07-18
Make cancellation of mempool request callbacks non-fatal

mzabaluev:
# Summary

When a mempool client request get passed with a oneshot callback,
the recipient of the callback may be dropped if the client
has cancelled the request. This condition should not be fatal
for the executor's transaction pipe.

# Testing

Will add a test to exercise the non-fatal handling specifically.

### DIFF
--- a/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
@@ -1,14 +1,15 @@
 use super::Executor;
-use aptos_mempool::SubmissionStatus;
 use aptos_mempool::{core_mempool::TimelineState, MempoolClientRequest};
 use aptos_sdk::types::mempool_status::{MempoolStatus, MempoolStatusCode};
 use aptos_types::transaction::SignedTransaction;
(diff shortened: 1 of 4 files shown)


protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -46,22 +47,16 @@ impl Executor {
...
-										callback.send(Ok(status)).map_err(
-											|e| anyhow::anyhow!("Error sending callback: {:?}", e)
-										)?;
-										continue;
-									}
-
-									// add to the mempool
-									{
 
+									let status = if let Some(vm_status) = tx_result.status() {
```
mzabaluev, line 51: Blame @0xmovses's tab settings in `rustfmt.toml` for this diff gore.

On the other hand, this code could be moved out to a helper function to be less nested.

0xmovses: True diff gore. We could try not using the rule `hard_tabs = true`.

musitdev (approved):
It seems to be ok to panic on `vm_validator.validate_transaction()` error. This type of error seems to be not recoverable.

PR #231, opened 2024-07-19
Bridge User Interface

0xPrimata:
# Summary
- **RFCs**:  $\emptyset$.

Bridge Service User facing Interface

# Changelog

adds to the protocol-units/bridge the interface frontend

# Testing

# Outstanding issues


0xmovses:
I don't think we should be putting user interfaces into `movement` core and I wouldn't consider a user inerface a protocol-unit. I think it would be nice to have a seperate repo for org build dapps and UI's @l-monninger, what do you think?

PR #232, opened 2024-07-19
Downgrade Docker Vars

l-monninger:
# Summary
- **RFCs**: $\emptyset$
- **Categories**:  `cicd`

Docker compose support for boolean environment variables is engine-specific, so strings are safer. 

# Changelog

1. Changes boolean `true` to string `"true"` for maybe actions. 

# Testing

1. e2e tests. 

# Outstanding issues


musitdev (approved):
Strange you get an issue because the boolean value is not tested when the env var is read, only its presence.

mzabaluev (approved):
Makes sense.

I have more quality of life concerns, though:

* Why have we taken to naming environment variables `MAYBE_*` and what are the semantics?
* It's a bit weird to use stringized booleans; as can be seen here, it's error-prone with YAML as well. The convention with boolean-ish flags in the environment is to treat unset as "no" and use numbers when the variable is set: 0 again means "no" and any other number means "yes". [See](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables) `RUST_BACKTRACE` for example.

l-monninger:
> Why have we taken to naming environment variables MAYBE_* and what are the semantics?
Two cases:
1. The `run_local` `MAYBE_*` is supposed to only be applied if the variables that are actually needed to `run_local` are correct.
2. The `MAYBE_RUN_DEPLOY` is supposed to respect existing state and not deploy if a deployed contract is already known.

l-monninger:
> It's a bit weird to use stringized booleans; as can be seen here, it's error-prone with YAML as well. The convention with boolean-ish flags in the environment is to treat unset as "no" and use numbers when the variable is set: 0 again means "no" and any other number means "yes". See RUST_BACKTRACE for example.

I agree. I was going to change it to numbers, but there are a few more changes that would need to take place. I will put it in a separate PR.

PR #234, opened 2024-07-19
Transfer Faucet

l-monninger:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
 [[package]]
 name = "abstract-domain-derive"
 version = "0.1.0"
-source = "git+https://github.com/movementlabsxyz/aptos-core?rev=b2f58eaeda1d3929a7b381afca78408731b71d77#b2f58eaeda1d3929a7b381afca78408731b71d77"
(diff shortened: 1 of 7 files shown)


l-monninger:
It's amazing how unergonomic and brittle this bit of software is. 

1. You cannot use the standard address derived from the provided private key to perform minting operations. You have to use `aptos_test_rood_address` and then sign with the private key that was used to sign the genesis. 
2. You can use the `MintFunder` struct. But, despite supposedly handling sequence numbers, you will get a sequence number error, likely first `SEQUENCE_NUMBER_TOO_OLD` on the full-node block execution which is not actually picked up in the faucet.  If, however, you increase this sequence number by one on the same transaction, you will get `SEQUENCE_NUMBER_TOO_NEW` . 
3. If you use `with_delegated_account` you may be able to get past the initial sequence number issues, I was able to once. But, then you will likely get a `TRANSACTION EXPIRED` Warn message and skip from the executor. However, appears from the output the WARN message that secs--rather than usecs--have somehow been sent by `MintFunder` internals for the expiration time. 

As a next step, you may wish to instead try calling the mint operations directly gain. But, instead of using `TransactionFactory`, use `TransactionBuilder`.

l-monninger:
Perhaps there is some kind of race condition here between the facuet and full-node that the `RunSimple` CLI command somehow manages. I tried waiting for longer before I started the faucet, but that didn't seem to make a difference.

PR #235, opened 2024-07-21
Sequencer Improvements

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Improvements targeted increased sequencer throughput and increased block size. 

# Changelog

1. Increases default block size (which was a dev value previously). 
2. Allows sequencer to stock building a block during inner loop if time-limit is exceeded. 
3. Optimizes sequencer transaction key composition. 

# Testing

1. e2e and unit testing for relevant parts. 

# Outstanding issues
1. Sequencer serialization can still be optimized.


benches/howzit/src/howzit.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub fn generate_exponential<R>(rng : &mut R) -> Self
+        where R : rand::Rng
+    {
+
+        // generate a random number between 0 and 2^PROBE_RANGE.end
+        let max = 2u32.pow(*Self::PROBE_RANGE.end());
+        let random = rng.gen_range(0, max);
+
+        // subtract the random number from 2^PROBE_RANGE.end to invert the distribution
+        let inverse = 2u32.pow(*Self::PROBE_RANGE.end()) - random;
```
mzabaluev, line 41: It would be clearer to use `max` here as well rather than copy the expression.

benches/howzit/src/howzit.rs
```diff
@@ -0,0 +1,212 @@
...
+        match self {
+            Probe::Probe1 => Ok(Identifier::new("probe_1")?),
+            Probe::Probe2 => Ok(Identifier::new("probe_2")?),
+            Probe::Probe3 => Ok(Identifier::new("probe_3")?),
+        }
+    }
+}
+pub struct Howzit {
+    pub howzit_package_path : PathBuf,
+    pub wallet : Arc<RwLock<LocalAccount>>,
```
mzabaluev, line 80: Does this really have to have interior mutability?

benches/howzit/src/howzit.rs
```diff
@@ -0,0 +1,212 @@
...
+                rest_client
+            ).with_auth_token(
+                faucet_auth_token.clone()
+            ),
+            faucet_auth_token
+        }
+    }
+
+    /// Builds and publishes the howzit package
+    pub async fn build_and_publish(&self) -> Result<(), anyhow::Error> {
```
mzabaluev, line 132: Why not just make the `&mut self` receiver so you don't have to deal with locking?

mzabaluev (approved):

benches/howzit/src/howzit.rs
l-monninger: `LocalAccount` ought to be borrowed mutably because it has to coordinate its sequencer number globally.

PR #243, opened 2024-07-22
Performance Tuning

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Tunes full node execution performance. 

# Changelog

1. Removes unnecessary synchronization primitives in `maptos-opt-executor`, `move-rocks`, `memseq`.
2. Optimizes `move-rocks` serialization. 
3. Uses `spawn_blocking` for CPU-heavy operations in `maptos-opt-executor`. 
4. Removes VM validation from mempool ingestion.
5. Removes sleeps, timeouts, and inefficient batching in transaction dataflow--generally pushing more of these concerns to the scheduler/tokio runtime. 

# Testing

1. e2e tests passing. 

# Outstanding issues
1. This is part of ongoing performance tuning of the `suzuka-full-node`.


benches/howzit/Cargo.toml
```diff
@@ -0,0 +1,36 @@
+[package]
+name = "howzit"
```
mzabaluev, line 2: Just how many diffs with howzit are we going to have, and which one will be the one to review and merge? Let's settle on one version and get it to main, so that any other PR can be rebased to only have the relevant changes.

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -32,78 +32,76 @@ impl Executor {
...
-									if sender.send(mempool_result).is_err() {
-										debug!("get_transaction_by_hash request has been canceled");
-									}
+		// Drop the receiver RwLock as soon as possible.
+		let next = {
+			let mut mempool_client_receiver = self.mempool_client_receiver.write().await;
+			mempool_client_receiver.next().await
+		};
+		
+		if let Some(request) = next  {
```
mzabaluev, line 41: When the stream is closed, assuming it's fused, `next` will be `None` and the `tick_transaction_pipe` call with return `Ok(())` indefinitely. Perhaps it's better to return an error in this case like in #244.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -34,21 +34,36 @@ impl RocksdbMempool {
...
+		key.push_str(&pad_to_32_chars(transaction.transaction.sequence_number));
+		key.push(':');
+	
 		// Assuming transaction.transaction.id() returns a hex string of length 32
 		let transaction_id_hex = transaction.transaction.id(); // This should be a String of hex characters
-
-		// Concatenate the two parts to form a 80-character hex string key
-		let key = format!("{}:{}:{}", slot_seconds_str, sequence_number_str, transaction_id_hex);
-
+		key.push_str(&transaction_id_hex.to_string().as_str());
```
mzabaluev, line 61: It may be actually faster to go through the formatting framework to compose the key than create the temporary strings. BTW, `"str".to_owned()` is faster than `"str".to_string()`, because it does not work through the formatting.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -34,21 +34,36 @@ impl RocksdbMempool {
...
-	pub fn construct_mempool_transaction_key(transaction: &MempoolTransaction) -> String {
-		// pad to 32 characters for slot_seconds
-		let slot_seconds_str = format!("{:032}", transaction.timestamp);
-
-		// pad to 32 characters for sequence number
-		let sequence_number_str = format!("{:032}", transaction.transaction.sequence_number);
 
+	pub fn construct_mempool_transaction_key(transaction: &MempoolTransaction) -> String {
+		// Pre-allocate a string with the required capacity
+		let mut key = String::with_capacity(32 + 1 + 32 + 1 + 32);
```
mzabaluev, line 40: So, this is an optimization because it pre-allocates the exact capacity...

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -34,21 +34,36 @@ impl RocksdbMempool {
...
+	
+		// Function to pad an integer to 32 characters with leading zeros
+		fn pad_to_32_chars(num: u64) -> String {
+			let mut s = num.to_string();
+			while s.len() < 32 {
+				s.insert(0, '0');
+			}
+			s
+		}
```
mzabaluev, line 49: ...but this is certainly worse than `format!("{num:032}")` because it repeatedly reallocates the temporary string and repeatedly inserts a character (a UTF-8 sequence, in fact) to the beginning with all of `O(N*2)` efficiency. Have you measured how this change improves performance?

protocol-units/mempool/move-rocks/src/lib.rs
l-monninger: Should we use or avoid formatting? I feel like both my implementation and this comment are a bit chimerical. Do you want me to bench?

protocol-units/mempool/move-rocks/src/lib.rs
l-monninger: What string are you saying is reallocated? What is the $n$ in your $O(n^2)$?

l-monninger: Oh, you're saying because the `s` is not allocated with the appropriate capacity.

l-monninger:
```rust
fn pad_to_32_chars(num: u64) -> String {

	let num_str = num.to_string();
    let num_len = num_str.len();

    // If the length is already 32 or more, return the string as is
    if num_len >= 32 {
        return num_str;
    }

    // Create a new String with a capacity of 32
    let mut padded_str = String::with_capacity(32);

    // Calculate the number of zeros needed
    let zeros_to_add = 32 - num_len;

    // Push the required number of zeros
    for _ in 0..zeros_to_add {
        padded_str.push('0');
    }

    // Push the original number string
    padded_str.push_str(&num_str);

    padded_str
}


#[test]
fn test_pad_to_32_chars_outperforms_format32() -> Result<(), anyhow::Error> {

// generate a vec of random u64s
let mut rng = rand::thread_rng();
let nums: Vec<u64> = (0..1000).map(|_| rng.gen()).collect();

// for each number time the two functions and store the results in a vec
let mut results = Vec::with_capacity(nums.len());
for num in nums.iter() {
	let start = std::time::Instant::now();
	let _ = pad_to_32_chars(*num);
	let pad_to_32_chars_time = start.elapsed();

	let start = std::time::Instant::now();
	let _ = format!("{:032}", num);
	let format32_time = start.elapsed();

	results.push((pad_to_32_chars_time, format32_time));
}

// check the mean time for each function
let pad_to_32_chars_mean = results.iter().map(|(a, _)| a).sum::<std::time::Duration>() / results.len() as u32;
let format32_mean = results.iter().map(|(_, b)| b).sum::<std::time::Duration>() / results.len() as u32;

// check that pad_to_32_chars is faster than format32
println!("pad_to_32_chars_mean: {:?}", pad_to_32_chars_mean);
println!("format32_mean: {:?}", format32_mean);
assert!(pad_to_32_chars_mean < format32_mean);

Ok(())

}

```

```
pad_to_32_chars_mean: 522ns
format32_mean: 245ns
```

Format is faster.

protocol-units/mempool/move-rocks/src/lib.rs
mzabaluev: I think it's best to return to the original. There is one trick that can let you both pre-allocate and format: call [`write_fmt`](file:///Users/mzabaluev/.rustup/toolchains/stable-aarch64-apple-darwin/share/doc/rust/html/std/fmt/trait.Write.html#method.write_fmt) on the mutable `key: String`

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
mzabaluev: Done in [9e10091](https://github.com/movementlabsxyz/movement/pull/243/commits/9e10091af6867347029b877c3582f2037e52bdb9).

protocol-units/mempool/move-rocks/src/lib.rs
mzabaluev: > There is one trick that can let you both pre-allocate and format: call [`write_fmt`](https://doc.rust-lang.org/std/fmt/trait.Write.html#method.write_fmt) on the mutable `key: String`.

Formatting in pre-allocated string done in [7772481](https://github.com/movementlabsxyz/movement/pull/243/commits/777248114d716e25f67f7ccf65bddd5d9c51682b)

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -34,27 +36,19 @@ impl RocksdbMempool {
...
-		key.push_str(&transaction_id_hex.to_string().as_str());
-	
+		// Write key components. The numbers are zero-padded to 32 characters.
+		key.write_fmt(format_args!(
+			"{:032}:{:032}:{}",
+			transaction.timestamp,
+			transaction.transaction.sequence_number,
+			transaction.transaction.id(),
+		))
+		.unwrap();
```
mzabaluev, line 49: Note that using a binary key instead of a stringized decimal-hex one would improve performance much more than these micro-optimizations.

l-monninger: Feel free to make this change.

networks/suzuka/suzuka-client/src/tests/mod.rs
```diff
@@ -182,76 +182,6 @@ async fn test_example_interaction() -> Result<(), anyhow::Error> {
 			.context("Failed to get Bob's account balance the second time")?
 	);
 
-	// malformed sequence number
-	println!("\n=== Malformed Sequence Number ===");
```
mzabaluev, line 186: If this test code is still useful, it could be moved to its own test case that is marked with `#[ignore]` if it has to be disabled at the moment.

networks/suzuka/suzuka-full-node/src/main.rs
```diff
@@ -8,6 +8,7 @@ use std::process::ExitCode;
 fn main() -> Result<ExitCode, anyhow::Error> {
 	let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build()?;
 
+	// console_subscriber::init();
```
mzabaluev, line 11: This could be composed with `tracing_subscriber` and conditionally enabled. The docs give an [example](https://docs.rs/console-subscriber/latest/console_subscriber/#adding-the-console-subscriber) to build on.

protocol-units/execution/opt-executor/Cargo.toml
```diff
@@ -69,7 +68,7 @@ aptos-indexer-grpc-table-info = { workspace = true }
 aptos-indexer = { workspace = true }
 aptos-protos = { workspace = true }
 aptos-logger = { workspace = true }
-
+tonic = { workspace = true }
```
mzabaluev, line 71: This dependency does not seem to be needed.

protocol-units/execution/opt-executor/src/executor/execution.rs
```diff
@@ -41,51 +41,50 @@ impl Executor {
 			}
 		};
 
+		let block_executor = self.block_executor.clone();
```
mzabaluev, line 44: How large is the state that is being cloned here?

protocol-units/execution/opt-executor/src/executor/execution.rs
```diff
@@ -41,51 +41,50 @@ impl Executor {
...
-			block_executor.committed_block_id()
-		};
+		let parent_block_id = block_executor.committed_block_id();
 
 		let mempool_transactions = block.transactions.clone().into_txns();
 
-		let state_compute = {
-			let block_executor = self.block_executor.write().await;
-			block_executor.execute_block(
+		let block_executor_clone = block_executor.clone();
```
mzabaluev, line 51: Twice?

Perhaps it's time to refactor this from the god `&self` object and methods interface that is really forcing either the locking or cloning? #63 for the discussion.

l-monninger: I think you're taking a look at an actor pattern implementation that will potentially resolve this. But, the `.clone()` here is fairly low overhead. It's an `Arc` that we're cloning.

protocol-units/execution/opt-executor/src/executor/execution.rs
l-monninger: It's an `Arc`.

protocol-units/execution/opt-executor/Cargo.toml
l-monninger: This was from trying to see if the result of `git bisect` was indeed valid.

l-monninger: It was not.

networks/suzuka/suzuka-full-node/src/main.rs
l-monninger: Yes, I needed to figure out how to add the `EnvFilter` though and deprioritized.

networks/suzuka/suzuka-client/src/tests/mod.rs
l-monninger: Without the `ValidationVM::validate` call in the mempool, it will not pass.

mzabaluev: We still have that code, albeit commented out. The ignored test would remind us to add it in.

mzabaluev (approved):
It's generally an improvement, we need to bring lots of this code into main, and I have some more changes on top, so let's go ahead with it.

PR #244, opened 2024-07-23
Make maptos-opt-executor tick better

mzabaluev:
# Summary

Eliminate unnecessary looping and timeouts in `Executor::tick_transaction_pipe`.

# Testing

CI tests shall pass.


mzabaluev:
#243 addresses the same issue and more. The only change missing there is returning an error when the input stream is closed, I commented on that.

PR #247, opened 2024-07-24
Add spans and timing into tracing logs

mzabaluev:
# Summary
- **Categories**: any of `protocol-units`, `networks`.

Instrument async processing flows with tracing spans to get information about latency of processing.
Use KV pairs in spans to identify entities such as transactions.

Currently based on #243 to have simpler, better behaving code to instrument.

# Changelog

* suzuka-full-node, m1-da-light-node: log tracing events with timing information suzuka-full-node if the `SUZUKA_TIMING` environment variable is set. The information logged is filtered accordingly to the directives given in the value of `SUZUKA_TIMING`, with the same syntax as `RUST_LOG`.

# Testing

Run suzuka-full-node in a local setup with e.g. the `test` overlay:

```
SUZUKA_TIMING=info just suzuka-full-node native build.setup.eth-local.celestia-local.test
```

Observe the events created in the `suzuka-full-node.timing.json` file.

### DIFF
--- a/networks/suzuka/suzuka-full-node/src/main.rs
+++ b/networks/suzuka/suzuka-full-node/src/main.rs
@@ -4,12 +4,12 @@ use suzuka_full_node::{manager::Manager, partial::SuzukaPartialNode};
 use std::env;
 use std::process::ExitCode;
 
-const TIMING_ENV_VAR: &str = "SUZUKA_TIMING";
-const TIMING_LOG_ENV_VAR: &str = "SUZUKA_TIMING_LOG";
-const DEFAULT_TIMING_LOG_FILE: &str = "suzuka-full-node.timing.json";
+const TIMING_LOG_ENV: &str = "SUZUKA_TIMING_LOG";
 
 fn main() -> Result<ExitCode, anyhow::Error> {
-	let _guard = init_tracing_subscriber();
+	let tracing_config =
+		movement_tracing::Config { timing_log_path: env::var_os(TIMING_LOG_ENV).map(Into::into) };
+	let _guard = movement_tracing::init_tracing_subscriber(tracing_config);
 
 	let runtime = tokio::runtime::Builder::new_multi_thread().enable_all().build()?;
 
@@ -31,51 +31,4 @@ async fn run_suzuka() -> Result<ExitCode, anyhow::Error> {
 	manager.try_run().await?;
 
 	Ok(ExitCode::SUCCESS)
-}
-
-fn init_tracing_subscriber() -> Option<tracing_appender::non_blocking::WorkerGuard> {
-	use std::{fs::File, path::PathBuf};
-	use tracing_subscriber::filter::{EnvFilter, LevelFilter};
-	use tracing_subscriber::fmt::format::FmtSpan;
-	use tracing_subscriber::prelude::*;
-
-	// TODO: compose console_subscriber as a layer
-	let env_filter = EnvFilter::builder()
-		.with_default_directive(LevelFilter::INFO.into())
-		.from_env_lossy();
-	let log_layer = tracing_subscriber::fmt::layer().with_filter(env_filter);
-
-	let (timing_layer, timing_writer_guard) =
-		if let Ok(timing_directives) = env::var(TIMING_ENV_VAR) {
-			let env_filter = EnvFilter::try_new(timing_directives);
-			let timing_log_path: PathBuf = env::var_os(TIMING_LOG_ENV_VAR)
-				.unwrap_or_else(|| DEFAULT_TIMING_LOG_FILE.into())
-				.into();
-			let timing_log_file = File::create(&timing_log_path);
-			match (env_filter, timing_log_file) {
-				(Ok(env_filter), Ok(file)) => {
-					let (writer, guard) = tracing_appender::non_blocking(file);
-					let layer = tracing_subscriber::fmt::layer()
-						.with_writer(writer)
-						.json()
-						.with_span_events(FmtSpan::CLOSE)
-						.with_filter(env_filter);
-					(Some(layer), Some(guard))
-				}
-				(Err(e), _) => {
-					eprintln!("invalid value of {TIMING_ENV_VAR}: {e}");
-					(None, None)
-				}
-				(_, Err(e)) => {
-					eprintln!("can't create `{}`: {}", timing_log_path.display(), e);
-					(None, None)
-				}
-			}
-		} else {
-			(None, None)
-		};
-
-	tracing_subscriber::registry().with(log_layer).with(timing_layer).init();
-
-	timing_writer_guard
 }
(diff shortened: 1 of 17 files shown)


protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -43,55 +42,15 @@ impl Executor {
...
-							.await
-							.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
-
-						// report status
-						let ms = MempoolStatus::new(MempoolStatusCode::Accepted);
-						(ms, None)
-					};
+					let span = info_span!(
+						"submit_transaction",
+						seq = transaction.sequence_number(),
```
l-monninger, line 47: Should have transaction hash.

l-monninger: Also, let's use `sequence_number`.

networks/suzuka/suzuka-full-node/src/main.rs
```diff
@@ -36,3 +32,15 @@ async fn run_suzuka() -> Result<ExitCode, anyhow::Error> {
 
 	Ok(ExitCode::SUCCESS)
 }
+
+fn init_tracing_subscriber() {
+	// TODO: compose console_subscriber as a layer
+	let env_filter = EnvFilter::builder()
+		.with_default_directive(LevelFilter::INFO.into())
+		.from_env_lossy();
+	let mut subscriber = tracing_subscriber::fmt().with_env_filter(env_filter);
```
l-monninger, line 41: I think I'd like to have an option to write these spans directly to `.dat` perhaps by analysis. `aptos-core` has a setup where they use an enum to identify the table that their performance based logs should go into I believe. Probably easier to just use an generic type. 

But, here, even if it's just making the data mining a little be easier, it would be nice to have these written directly to a file.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -169,43 +162,12 @@ where
 				}
 			};
 
-			let block: Block = serde_json::from_slice(&block_bytes)?;
+			let span = info_span!("execute_block", id = block_id);
```
l-monninger, line 165: This needs a companion in the `m1-da-light-node`.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -104,8 +100,8 @@ where
 					let serialized_aptos_transaction = serde_json::to_vec(&transaction)?;
```
l-monninger, line 100: We want to trace the transaction here as well, so that we know when it's come out of the mempool.

protocol-units/execution/opt-executor/src/executor/execution.rs
```diff
@@ -41,51 +41,50 @@ impl Executor {
...
+		let block_executor_clone = block_executor.clone();
+		tokio::task::spawn_blocking(move || {
+			block_executor_clone.commit_blocks(vec![block_id], ledger_info_with_sigs)
+		}).await??;
+
+		// commit mempool transactions in batches of size 16
+		for chunk in mempool_transactions.chunks(16) {
+			let mut core_mempool = self.core_mempool.write().await;
+			for transaction in chunk {
 				match transaction.clone().into_inner() {
```
l-monninger, line 83: We also want to trace here when the transaction is confirmed.

l-monninger (changes requested):
I think let's also include the transaction signer address wherever possible. This will allow me to analyze contention w.r.t. the number of transactions a single user can keep in flight.

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
mzabaluev: > Should have transaction hash.

I'm a bit grossed out by the API to get that out of `SignedTransaction`: you either call `.committed_hash()` directly in which case it gets also cached in the object, or you do `.clone().committed_hash()` like in other places in our code for some reason (to avoid the caching side-effect when logging statement is enabled vs. when it's not?). This looks a bit wasteful to do just for logging, but if there is no other way, I will put it there.

networks/suzuka/suzuka-full-node/src/main.rs
mzabaluev: I need to look if it's possible to have a custom subscriber for this.

protocol-units/execution/opt-executor/src/executor/execution.rs
mzabaluev: Do we need a span around the `commit_transaction` call, or is it expected to be quick?

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Is this async block of interest: https://github.com/movementlabsxyz/movement/blob/497ec0f388596efdf837b48a9aa71accebbc1e78/protocol-units/da/m1/light-node/src/v1/sequencer.rs#L60?

l-monninger: I would `span` the `wait_for_block` and the `submit_block`. Then, I would also at least log here: 

https://github.com/movementlabsxyz/movement/blob/497ec0f388596efdf837b48a9aa71accebbc1e78/protocol-units/da/m1/light-node/src/v1/passthrough.rs#L163

networks/suzuka/suzuka-full-node/src/main.rs
mzabaluev: Here's what I'm trying to implement quickly: a separate `fmt` subscriber with JSON output written to a file, logging only the closing events for any spans that match filtering directives. The directives will be configured from the value of the `SUZUKA_TIMING` environment variable.

JSON can then be massaged into any data format suitable for analysis.

musitdev: I've done something similar for the load and soak framework: https://github.com/movementlabsxyz/movement/blob/803354aa6c369885e0163b44001989f7729d6fa1/networks/suzuka/suzuka-client/src/load_soak_testing/mod.rs#L27
Perhaps it can help.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: If we want to have a span for waiting at the beginning in `tick_block_proposer`, what could be the identifying information? Can we anticipate the next height, for example?

mzabaluev: I have added an instrumentation span around the `LightNodeV1::get_blobs_at_height` method, which also logs a debug level event with each blob's ID. Hopefully this gives enough visibility from the passthrough flow.

mzabaluev:
gali

mzabaluev:
This has been merged into #258.

PR #248, opened 2024-07-24
change or remove tokio RwLock

musitdev:
# Summary

I change tokio `RwLock` to Std one when it's possible and remove one that is not necessary.

Changes in:
 * Executor::core_mempool
 * Memseq::parent_block
 * 
Remove RocksdbMempool::DB one.

From my test it doesn't change much the perf. I propose the PR because it was mostly done when we discuss it.
Decide if it should be applied or not.

# Changelog

# Testing

# Outstanding issues


protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -56,7 +56,7 @@ impl Executor {
 					let status = {
 						// add to the mempool
 						{
-							let mut core_mempool = self.core_mempool.write().await;
+							let mut core_mempool = self.core_mempool.write().unwrap();
```
l-monninger, line 59: We should not be using unwrap.

l-monninger (changes requested):
`cargo check` not passing for some tests.

1. The removal of the `RwLock` on the mempool is good. 
2. I don't think the change to an `std::RwLock` is necessary.

l-monninger:
It seems like maybe my comments didn't actually go through from VsCode. I think the removing the async `RwLock` is probably necessary. I think for `memseq` the improvements are great. But, I do think we should be `spawn_blocking` because `rocksdb` is blocking IO. So, if you can rewrite to `Arc<DB>` and do things like this:

```rust
let da_db = self.da_db.clone();
let height = tokio::task::spawn_blocking(move ||{
	let cf = da_db.cf_handle("synced_height").ok_or(anyhow::anyhow!("No synced_height column family"))?;
	let height = da_db.get_cf(&cf, "synced_height").map_err(|e| anyhow::anyhow!("Failed to get synced height: {:?}", e))?;
	let height = match height {
		Some(height) => serde_json::from_slice(&height).map_err(|e| anyhow::anyhow!("Failed to deserialize synced height: {:?}", e))?,
		None => 0
	};
	Ok::<u64, anyhow::Error>(height)
}).await??;
Ok(height)
```

That would be good IMO.

musitdev:
When I was building the hotstuff node, I had an issue with Tokio. When the load was increasing, the node become late, but there were no reason regarding the CPU. I solve the pb using `spawn_blocking` on all Rocksdb call.  Db access was creating micro delay in tokio scheduling and some channels that has no link with the DB weren't processed in time. 
On the other hand, yesterday when I try to spwan_blocking the Tx validation, I get worse performance.
So it depends a lot on the application. Generally I would preconize to spawn_blocking all bounded CPU call, but it can be counterproductive when there's too much task. It starts a lot of thread and the kernel spend a lot of time to process them. SUI has this type of issue when they started their testnet. That why in the end I will preconize to define an architecture where CPU processing and network process are dissociated. Even then you can do some CPU affinity, you can start one CPU/thread for the network and use all others CPU for the remaining.

PR #251, opened 2024-07-24
replaced serde_json with bcs in move-rocks

andygolay:
# Summary
Replace `serde_json` with `bcs` in `move-rocks`

# Changelog

Replaced in relevant functions (see diff)

# Testing

All five unit tests pass:

```
test test::test_rocksdb_mempool_basic_operations ... ok
test test::test_transaction_sequence_number_based_ordering ... ok
test test::test_rocksdb_transaction_operations ... ok
test test::test_slot_and_transaction_based_ordering ... ok
test test::test_transaction_slot_based_ordering ... ok
```

# Outstanding issues
- Please review to ensure that there are no side effects that might impact other protocol units.
- This should be benchmarked to know whether it indeed provides a performance boost.

PR #253, opened 2024-07-25
Replace serde_json with bcs in Suzuka partial.rs

### LINKED ISSUE #182: Remove ad-hoc JSON formatting in sequencer blobs
In #171, sequence numbers are introduced in a little kludgy way, by imposing JSON encoding on a `bytes` field of a protobuf message. Move these values to be first-class fields of the protobuf message instead.

_Originally posted by @mzabaluev in https://github.com/movementlabsxyz/movement/pull/171#discussion_r1672390762_

andygolay:
Closes #182

# Summary
Replaced `serde_json` with `bcs`

# Changelog

- Changed `serde_json::to_vec` to `bcs::to_bytes`

- Changed `serde_json::from_slice` to `bcs::from_bytes`

# Testing

This compiles but hasn't been tested. (See below for details.)

# Outstanding issues
I don't know that it's reasonable to assume transactions will come in in BCS format, even after making the mempool use BCS in [this PR](https://github.com/movementlabsxyz/movement/pull/251). There are some uses of `serde_json` in the DA/M1 protocol unit which seem like they may need to be changed first before this is a good change to implement. 

I wasn't sure how to test this E2E. I tried `just suzuka-full-node native build.setup.test.local` which returned:

```
FTL Failed to read process-compose/suzuka-full-node/process-compose.local.yml error="open process-compose/suzuka-full-node/process-compose.local.yml: no such file or directory"
```

Any advice on how to test this is appreciated.

### DIFF
--- a/networks/suzuka/suzuka-full-node/src/partial.rs
+++ b/networks/suzuka/suzuka-full-node/src/partial.rs
@@ -1,4 +1,5 @@
 use crate::SuzukaFullNode;
+use bcs;
 use m1_da_light_node_client::{
 	blob_response, BatchWriteRequest, BlobWrite, LightNodeServiceClient,
 	StreamReadFromHeightRequest,
@@ -7,9 +8,7 @@ use maptos_dof_execution::{
 	v1::Executor, DynOptFinExecutor, ExecutableBlock, ExecutableTransactions, HashValue,
 	SignatureVerifiedTransaction, SignedTransaction, Transaction,
 };
-use mcr_settlement_client::{
- McrSettlementClient, McrSettlementClientOperations,
-};
+use mcr_settlement_client::{McrSettlementClient, McrSettlementClientOperations};
 use mcr_settlement_manager::CommitmentEventStream;
 use mcr_settlement_manager::{McrSettlementManager, McrSettlementManagerOperations};
 use movement_rest::MovementRest;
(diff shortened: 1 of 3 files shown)


networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -99,13 +98,13 @@ where
...
+					let serialized_aptos_transaction = bcs::to_bytes(&transaction)?;
 					debug!("Serialized transaction: {:?}", serialized_aptos_transaction);
 					let movement_transaction = movement_types::Transaction {
-						data : serialized_aptos_transaction,
-						sequence_number : transaction.sequence_number()
+						data: serialized_aptos_transaction,
+						sequence_number: transaction.sequence_number(),
 					};
-					let serialized_transaction = serde_json::to_vec(&movement_transaction)?;
+					let serialized_transaction = bcs::to_bytes(&movement_transaction)?;
```
mzabaluev, line 107: Good.
I wonder if there is a non-accidental reason why we need to apply two levels of framing here. @l-monninger is the inner transaction's encoding meant to be opaque in the wire representation of `movement_types::Transaction`?

andygolay: Opened for review

l-monninger: > Good. I wonder if there is a non-accidental reason why we need to apply two levels of framing here. @l-monninger is the inner transaction's encoding meant to be opaque in the wire representation of `movement_types::Transaction`?

Yes, it's meant to be opaque as the sequencer is supposed to be as general as possible and `movement_types` was not supposed to contain specific protos. But, perhaps we should change it as the types have been mirroring one another.

l-monninger:
#299 redoes this.

PR #254, opened 2024-07-25
Update: replace serde_json with bcs in move-rocks

andygolay:
# Summary
This an updated version of https://github.com/movementlabsxyz/movement/pull/251

which was opened around the same time of a merge which affected the `move-rocks` lib.

I opted to create a new branch (old one is deleted).

# Changelog

Updated to be consistent with main

# Testing

All unit tests pass

# Outstanding issues
Benchmarking could be good to to ensure it indeed provides a performance boost.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8443,11 +8443,11 @@ name = "move-rocks"
 version = "0.3.0"
 dependencies = [
  "anyhow",
+ "bcs 0.1.4",
  "mempool-util",
  "movement-types",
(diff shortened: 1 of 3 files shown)


l-monninger (approved):
Because this only affects the mempool, I think we are good merge.

musitdev (approved):

mzabaluev (approved):
Looks good to me.
Migration concern: do we empty this storage on startup of the process?

PR #255, opened 2024-07-25
Load Shedding

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

## Objective
Implement load shedding for transaction flow.

## Transaction Load Milestones
load-shedding

In the current implement only `v1`, pardon the uninspiring naming, is considered. I think ideally, we would transform this such that we have an `enum TransactionLoadMilestone`, that we increment and decrement on relevant structs.

On the full-node, implement `v1` is fairly trivial, as we can report status directly to the client via the `MempoolRequest` API. I believe you may also be able to use the `commit_transaction` and `reject_transaction` API in a similar manner. But I have not yet confirmed. 

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/execution/dof/src/v1.rs
+++ b/protocol-units/execution/dof/src/v1.rs
@@ -115,6 +115,13 @@ impl DynOptFinExecutor for Executor {
 	async fn rollover_genesis_block(&self) -> Result<(), anyhow::Error> {
 		self.executor.rollover_genesis_now().await
 	}
+
+	fn decrement_transactions_in_flight(&self, count : u64) {
+		self.executor.transactions_in_flight.fetch_sub(
+			count,
+			std::sync::atomic::Ordering::Relaxed, // relaxed because this is just for load shedding, now need for strict ordering.
+		);
+	}
 }
 
 #[cfg(test)]
--- a/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
@@ -1,5 +1,5 @@
 use super::Executor;
-use aptos_logger::info;
+use aptos_logger::{info, warn};
 use aptos_mempool::{core_mempool::TimelineState, MempoolClientRequest};
 use aptos_sdk::types::mempool_status::{MempoolStatus, MempoolStatusCode};
 use aptos_types::transaction::SignedTransaction;
@@ -81,7 +81,9 @@ impl Executor {
 							.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
 							self.transactions_in_flight.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
 						},
-						_ => {}
+						_ => {
+							warn!("Transaction not accepted: {:?}", status);
+						}
 					}
 
 					callback.send(Ok((status.clone(), None))).map_err(
--- a/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
@@ -1,15 +1,11 @@
 use super::Executor;
+use aptos_logger::info;
 use aptos_mempool::{core_mempool::TimelineState, MempoolClientRequest};
 use aptos_sdk::types::mempool_status::{MempoolStatus, MempoolStatusCode};
 use aptos_types::transaction::SignedTransaction;
-use aptos_vm_validator::vm_validator::TransactionValidation;
-use aptos_vm_validator::vm_validator::VMValidator;
-
 use futures::StreamExt;
 use thiserror::Error;
 use tracing::debug;
-
-use std::sync::Arc;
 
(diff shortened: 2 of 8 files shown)


l-monninger:
@mzabaluev It would be good to move these parameters out into the `SuzukaConfig` before going too much further.

protocol-units/execution/dof/src/v1.rs
```diff
@@ -115,6 +115,13 @@ impl DynOptFinExecutor for Executor {
 	async fn rollover_genesis_block(&self) -> Result<(), anyhow::Error> {
 		self.executor.rollover_genesis_now().await
 	}
+
+	fn decrement_transactions_in_flight(&self, count : u64) {
+		self.executor.transactions_in_flight.fetch_sub(
+			count,
+			std::sync::atomic::Ordering::Relaxed, // relaxed because this is just for load shedding, now need for strict ordering.
```
mzabaluev, line 122: The value itself will be always incremented and decremented atomically, so `Relaxed` is fine.
The stronger guarantees are only needed when you need to establish ordering with regard to other atomic operations.

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -43,68 +39,63 @@ impl Executor {
...
-					// 	// If the verification failed, return the error status.
-					// 	let ms = MempoolStatus::new(MempoolStatusCode::VmError);
-					// 	(ms, Some(vm_status))
-					// } else {
+
+					// Shed load.
+					// Low-ball the load shedding for now with 4096 transactions allowed in flight.
+					// For now, we are going to consider a transaction in flight until it exits the mempool and is sent to the DA as is indicated by WriteBatch.
+					if self.transactions_in_flight.load(std::sync::atomic::Ordering::Relaxed) > 2^12 {
+						info!("Transaction ins flight: {:?}, shedding load", self.transactions_in_flight.load(std::sync::atomic::Ordering::Relaxed));
```
mzabaluev, line 47: You already have the value from the comparison above, no need to load the second time. This can potentially even be a different value.

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -43,68 +39,63 @@ impl Executor {
...
+						let status = MempoolStatus::new(MempoolStatusCode::MempoolIsFull);
+						callback.send(Ok((status.clone(), None))).map_err(
```
mzabaluev, line 49: The clone is redundant:
```suggestion
						callback.send(Ok((status, None))).map_err(
```

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -43,68 +39,63 @@ impl Executor {
...
-								true,
-							);
-
-							match status.code {
-								MempoolStatusCode::Accepted => {
-									debug!("Transaction accepted: {:?}", transaction);
-								}
-								_ => {
-									debug!("Transaction not accepted: {:?}", status);
-									Err(TransactionPipeError::TransactionNotAccepted(status))?;
```
mzabaluev, line 80: This case has been turned to doing nothing, what's the reason for this?
Also, why not send the rejected status on the callback?

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -43,68 +39,63 @@ impl Executor {
...
+					match &status.code {
+						MempoolStatusCode::Accepted => {
+							// send along to the receiver
+							transaction_channel
 							.send(transaction)
 							.await
 							.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
+							self.transactions_in_flight.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
+						},
+						_ => {}
```
mzabaluev, line 83: Silently doing nothing here.

If this case is never supposed to happen, perhaps at least put out a `warn!()` tracing event in case it does?

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
l-monninger: Good point.

l-monninger:
@mzabaluev Is the above any better?

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
mzabaluev: Looks like it's still there, am I missing something?

This is easy to clean up later when we get around to fixing all the clippy lints, though.

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
mzabaluev: Any comment on this? Is it a legitimate case to silently ignore?

mzabaluev (approved):
Good to merge as the implementation of the "v1" milestone.

PR #256, opened 2024-07-25
Adds Garbage Collection on Interval for Mempool

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Adds mempool garbage collection on hard-coded 60 second interval. This should help with mempool load. 

# Changelog

1. Add `gc()` methods to `maptos-opt-executor`.
2. Call on interval in `dof-executor` background task. 

# Testing

1. e2e tested. 

# Outstanding issues
1. Should be added to the config perhaps. 
2. Mempool task spawning still deserves optimization.

### DIFF
--- a/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
@@ -28,6 +28,13 @@ impl From<anyhow::Error> for TransactionPipeError {
 }
 
 impl Executor {
+
+	pub async fn gc_mempool(&self) -> Result<(), TransactionPipeError> {
+		let mut core_mempool = self.core_mempool.write().await;
+		core_mempool.gc();
+		Ok(())
+	}
+
 	/// Pipes a batch of transactions from the mempool to the transaction channel.
 	/// todo: it may be wise to move the batching logic up a level to the consuming structs.
 	pub async fn tick_transaction_pipe(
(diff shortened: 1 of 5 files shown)


protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -28,6 +28,13 @@ impl From<anyhow::Error> for TransactionPipeError {
 }
 
 impl Executor {
+
+	pub async fn gc_mempool(&self) -> Result<(), TransactionPipeError> {
+		let mut core_mempool = self.core_mempool.write().await;
+		core_mempool.gc();
```
musitdev, line 34: Gc can be dangerous because it locks all the mempool. Perhaps we should add a log to instrument how long it takes to have an idea and identify too long lock.

mzabaluev: I can add a span to the tracing instrumentation later on.

mzabaluev: Even better, add a span now and we'll add target information later.

mzabaluev (approved):

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
l-monninger: @musitdev To make this compatible with the lockless mempool, I can just track the last time `gc` was run in the mempool loop.

PR #261, opened 2024-07-26
Atomic bridge initiator Move module

andygolay:
# Summary
Atomic bridge initiator Move module

# Changelog

Adds `atomic_bridge_initiator` module, with functions:

- `init_module` (creates `BridgeTransferStore` and `BridgeConfig` resources and moves them to deployer)
- `initiate_bridge_transfer` (initiator transfers some amount of MovETH to atomic bridge, new bridge transfer is created, bridge transfer store is updated)
- `complete_bridge_transfer` (equal amount of MovETH is burned, transfer state is updated)
- `refund_bridge_transfer` (equal amount of MovETH is transferred back to initiator, transfer state is updated) 

as well as a helper function `get_bridge_transfer_index`.

Constants for error codes and bridge transfer states are included. 

This module is designed for consistency with existing bridge modules and contracts.

# Testing

To run unit tests:
1. Navigate to `protocol-units/bridge/move-modules`
2. Run `aptos init` (Press Enter or input yes at all prompts.)
3. Update relevant addresses in `Move.toml` with your address generated in the `aptos init` step. Currently Move.toml uses `0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45` as an example address:

```
[addresses]
atomic_bridge = "0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45"
moveth = "0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45"
master_minter = "0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45"
minter = "0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45"
admin = "0x4143f27fd7c76ed9e191968ebfe7f2a46ec188f815776678a8a4b80ccf4ffa45"
```
4. Run `aptos move test`

Note: tests with the `#[expected_failure]` attribute pass if the test aborts. To verify that the test aborts in the specific expected manner, comment out the `#[expected_failure]` line, re-run `aptos move test` and check that when the test fails the expected error code is thrown.

The following unit tests are included:
- `test_initialize`
- `test_initiate_bridge_transfer`
- `test_initiate_bridge_transfer_no_moveth` (expected failure)
- `test_complete_bridge_transfer`
- `test_complete_bridge_transfer_wrong_preimage` (expected failure)
- `test_refund_bridge_transfer` (marked with expected failure in unit tests for now due to issue https://github.com/movementlabsxyz/movement/issues/272)
- `test_refund_completed_transfer` (expected failure)

All tests pass. The logic should be reviewed to ensure tests are aligned with bridge requirements.

# Outstanding issues

- It would be good for all modules and contracts to align on naming conventions and what addresses are needed; for example, is the `atomic_bridge` going to be the same address as the `master_minter` in practice? If so, those can potentially be combined, along with `creator` and other names for the same address. 
- Tests are repetitive with a lot of redundant boilerplate; I could potentially reuse `test_initiate_bridge_transfer` in the other tests but unless anyone thinks that refactor is important, I think it's fine to leave as is for now.
- `Move.toml` addresses should be configured for deploying under a resource account.


0xmovses:
In the testing steps could you leave instructions for how `Move.toml` needs to be edited in order to get the tests to pass? I think it's something along the lines of. 
1. run `aptos init`
2. copy address generated there. 
3. replace that address for values ..
etc. 

This will be helpful for other reviewers too. 

I think we will need a script to do this for us, so that eventually we can run all our move tests in CI. But that's for later.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+    use aptos_framework::timestamp;
+    use aptos_framework::primary_fungible_store;
+    use aptos_framework::dispatchable_fungible_asset;
+    use aptos_std::aptos_hash;
+    use std::signer;
+    use std::vector;
+    use std::bcs;
+    use moveth::moveth;
+
+    // Constants to represent the state of a bridge transfer
```
0xmovses, line 13: I think we can remove this coment

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+    const INITIALIZED: u8 = 0;
+    const COMPLETED: u8 = 1;
+    const REFUNDED: u8 = 2;
+
+    // Constants for error handling
```
0xmovses, line 18: Remove this comment as its self explanatory.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+
+    const EINSUFFICIENT_AMOUNT: u64 = 0;
+    const EINSUFFICIENT_BALANCE: u64 = 1;
+    const EDOES_NOT_EXIST: u64 = 2;
+    const EWRONG_PREIMAGE: u64 = 3;
+    const ENOT_INITIALIZED: u64 = 4;
+    
+
+    // A struct to hold the details of a bridge transfer
+    struct BridgeTransfer has key, store {
```
0xmovses, line 28:
```suggestion
    struct BridgeTransferDetails has key, store {
```
And then you can remove the comment, the code becomes more self documenting.

andygolay:
> In the testing steps could you leave instructions for how `Move.toml` needs to be edited in order to get the tests to pass? I think it's something along the lines of.
> 
> 1. run `aptos init`
> 2. copy address generated there.
> 3. replace that address for values ..
>    etc.
> 
> This will be helpful for other reviewers too.
> 
> I think we will need a script to do this for us, so that eventually we can run all our move tests in CI. But that's for later.

Yep, updated testing steps.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+
+        vector::push_back(&mut store.transfers, bridge_transfer);
+
+        event::emit_event(&mut store.bridge_transfer_initiated_events, BridgeTransferInitiatedEvent {
+            bridge_transfer_id: bridge_transfer_id,
+            originator: addr,
+            recipient: recipient,
+            amount: amount,
+            hash_lock: hash_lock,
+            time_lock: timestamp::now_seconds() + time_lock,
```
0xmovses, line 134: we need to update this to use `blockheight` as opposed to timestamp. On solidity side we use blockheight, that's the agreed constant for the bridge's concept of "time" and sync. 

In Initiator .sol:
`timeLock: block.number + timeLock,`

After updating this you'll have to update the test to to kick off blockheights within the test. Let me know if you need a pointer where to find, I saw some code for this in aptos-framework

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+        master_minter: &signer
+    ) acquires BridgeTransferStore, BridgeConfig {
+        let config_address = borrow_global<BridgeConfig>(@atomic_bridge).bridge_module_deployer;
+        let store = borrow_global_mut<BridgeTransferStore>(config_address);
+        let idx = get_bridge_transfer_index(&store.transfers, &bridge_transfer_id);
+        let bridge_transfer = vector::borrow_mut(&mut store.transfers, idx);
+
+        assert!(bridge_transfer.state == INITIALIZED, 1);
+        assert!(aptos_hash::keccak256(bcs::to_bytes(&pre_image)) == bridge_transfer.hash_lock, 2);
+        assert!(timestamp::now_seconds() <= bridge_transfer.time_lock, 3);
```
0xmovses, line 153: update assertions with blockheight after that's been updated.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,527 @@
...
+
+    public fun refund_bridge_transfer(
+        account: &signer, 
+        bridge_transfer_id: vector<u8>,
+        atomic_bridge: &signer
+    ) acquires BridgeTransferStore, BridgeConfig {
+        let config_address = borrow_global<BridgeConfig>(@atomic_bridge).bridge_module_deployer;
+        let store = borrow_global_mut<BridgeTransferStore>(config_address);
+        let idx = get_bridge_transfer_index(&store.transfers, &bridge_transfer_id);
+        let bridge_transfer = vector::borrow_mut(&mut store.transfers, idx);
```
0xmovses, line 175: Is this the only way to lookup the bridge_transfer, via its index?

0xmovses (changes requested):
Looking good! I think your suggestion about removing the repeating boilerplate logic is a good one. Let's add that in as it will reduce bloat here. There's a lot of repeating logic in the tests that obfuscates the core thing that's being tested

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: So then `time_lock` will represent a number of blocks rather than a number of seconds, correct?

0xmovses: yep, that's right.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: There are various `inline` helper functions in `[aptos_framework::vector](https://github.com/aptos-labs/aptos-core/blob/main/aptos-move/framework/move-stdlib/sources/vector.move)` but they all ultimately borrow with ether `vector::borrow_mut` or `vector::borrow` using the index, so I don't see how it would improve performance to use those functions rather than a pattern like this. 

If this boilerplate is too bloated I could potentially refactor it into a helper function in this module?

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: Okay I'll be afk for about an hour and then will get cracking on this. If you do have a link handy to what you saw in the framework, I'd be interested, thanks.

Once you finish this review pass I'll implement everything and re-submit for review.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: For this change, I think it's implied that the details are in the struct fields. So removing the comment is fine but we don't want to change the name of the struct to append `Details` at the end of the name. 

In the `atomic_bridge_counterparty` Move module it's named `BridgeTransfer`

Likewise it's named `BridgeTransfer` in [AtomicBridgeInitiator.sol](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/contracts/src/AtomicBridgeInitator.sol) 

For consistency to keep the name of the transfer struct the same across contracts and modules seems to me that `BridgeTransfer` is good to stay as it is.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
0xmovses: Up to you, if we need idx every time fine to leave it as it is .

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
0xmovses: Ah yes true, in the rust service its `BridgeTransferDetails`, all good lets leave it as it is.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,519 @@
...
+            vector::append(&mut combined_bytes, bcs::to_bytes(&i));
+            
+            let id = aptos_hash::keccak256(combined_bytes);
+            if (id == *bridge_transfer_id) {
+                break
+            };
+            i = i + 1;
+        };
+        i - 1
+    }
```
cde8, line 209: This seems really inefficient, I hope there is a better way to do this. Does the movelang has something like HashMap?

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
cde8: Guys, there must be another way, because this vector can become huge. We are not removing entries after they are completed, as we seem to want to keep them in state. This vector will accumulate to thousands of records. This would cost too much gas to process, especially since hashing is done on each visit. It's not even an option to keep it like this.

cde8 (changes requested):
We must revisit our strategy for managing bridge transfers, as the current approach involves maintaining comprehensive records, thereby creating a substantial data vector. It is unreasonable to expect end users to incur high gas fees for processing this entire vector. We need to identify an alternative method to handle this efficiently.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
0xmovses: There is a map yes, `table`and `smart_table`, also as long as entries are removed after completion, this should be fine. As indexing historic data will be done via events off chain. Same as on eth side.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: Okay, I'll find a better way.

andygolay: I implemented smart_table to replace vector for the list of transfers, to eliminate the need for searching by index and so we could get rid of the extra helper function I wrote. 

Still need to handle removing transfers once complete, will do that next.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
andygolay: Good point. I implemented `smart_table` which should be more efficient, and removed this `get_bridge_transfer_index` function.

andygolay:
> We must revisit our strategy for managing bridge transfers, as the current approach involves maintaining comprehensive records, thereby creating a substantial data vector. It is unreasonable to expect end users to incur high gas fees for processing this entire vector. We need to identify an alternative method to handle this efficiently.

I changed from vectors to smart tables for lists of transfers, which should be more efficient.

I also added the `drop` ability to the `BridgeTransfer` struct and made it so after refunding or completing a bridge transfer, the transfer is removed from the smart table.

So, there could still be congestion if tens of thousands of transfers are simultaneously initiated. It would be good to tests the practical upper bounds for number of initiated transfers.

Because events are emitted whenever transfers are initiated, completed or refunded, those events can potentially be used as a historical record.

andygolay:
> Looking good! I think your suggestion about removing the repeating boilerplate logic is a good one. Let's add that in as it will reduce bloat here. There's a lot of repeating logic in the tests that obfuscates the core thing that's being tested

I've added an `init_test` function to eliminate much of the repetitive upfront boilerplate in tests. I don't seem to be able to compose the test functions by, say passing the output of one in as the input to another, so this might be as streamlined as I can make the tests given the multi-step nature of the tests that refund or complete bridge transfers.

cde8 (approved):
Good work, Andy. It looks good to me as far as I can assess.

cde8 (changes requested):
The only thing are the questions on the master minter.

andygolay:
I believe I have made all requested changes @cde8 @0xmovses

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,491 @@
...
+            time_lock: block::get_current_block_height() + time_lock,
+        });
+
+        bridge_transfer_id
+    }
+
+    public fun complete_bridge_transfer(
+        account: &signer,
+        bridge_transfer_id: vector<u8>,
+        pre_image: vector<u8>,
```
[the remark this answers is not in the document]
andygolay: Sorry for missing this comment @cde8, I believe having separate names for master_minter and atomic_bridge addresses may be redundant / could have the same name but I'm not 100% sure. 

If we publish the package under a resource account then we could unify the moveth, master_minter, and atomic_bridge addresses under the same (resource account) address and give them all the same name. Does that seem like a good plan @0xmovses ? Or do we need the any of those roles to be distinct from the bridge address for decentralization or other reasons?

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -0,0 +1,491 @@
...
+        let combined_bytes = vector::empty<u8>();
+        vector::append(&mut combined_bytes, bcs::to_bytes(&addr));
+        vector::append(&mut combined_bytes, recipient);
+        vector::append(&mut combined_bytes, hash_lock);
+        vector::append(&mut combined_bytes, bcs::to_bytes(&store.nonce));
+
+        let bridge_transfer_id = aptos_std::aptos_hash::keccak256(combined_bytes);
+
+        let bridge_transfer = BridgeTransfer {
+            amount: amount,
```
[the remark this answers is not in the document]
andygolay: Changed.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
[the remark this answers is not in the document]
0xmovses: They can be under the same name, yep. And use the resource account is owned by the atomic bridge, so it has the ability to mint / burn moveth.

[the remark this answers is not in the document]
andygolay: @0xmovses @cde8 I've replace  moveth_minter with with atomic_bridge throughout the module and ensured tests still pass.

Should be fine with atomic bridge as resource account but let's double check when E2E testing as it requires configuring the addresses correctly.

andygolay:
> The only thing are the questions on the master minter.

I eliminated master minter in favor of only using the atomic_bridge address. Yes, anyone can call `complete_bridge_transfer` and make it execute successfully provided the right pre-image is provided. 

After it's verified that the pre-image is correct, then the caller is made a minter. Now that you mention it, though, I'm not sure that's necessary. Right now we have in `complete_bridge_transfer`:

```
        moveth::add_minter(atomic_bridge, signer::address_of(account));
        moveth::burn(account, @atomic_bridge, bridge_transfer.amount);
        moveth::remove_minter(atomic_bridge, signer::address_of(account));
```

But why not just have the atomic bridge as a minter do the burning:

```
         moveth::burn(atomic_bridge, @atomic_bridge, bridge_transfer.amount);
```

I just updated it to do this instead as it seems more efficient to me.

As long as the caller passes in the correct pre-image value, it seems to me that it's fine for the atomic bridge to do the minting. If the atomic bridge is published under a resource account, then there will be no key to expose; callers will simply pay transaction fees and the atomic bridge will mint or burn as needed provided that the the function parameters and assertions are satisfied.

0xmovses (approved):
:shipit:

PR #266, opened 2024-07-29
Improvements to Celestia Block Submission w.r.t Block Building 

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Moves the block builder and submitter into separate task loops. 

# Changelog

1. Task loop for building blocks.
2. Task loop for submitting blocks.
3. Also, I created a bad `XOR` with the load shedding PR which would limit our transactions in flight to 12. This is replaced with the correct `.pow` here. 

# Testing

1. e2e tests.

# Outstanding issues
1. This is only valid because, currently, we allow blocks to be dropped at this stage. In the future, our requirements should change to prevent this. In which case, there will need to be handlers that try to resize and resubmit the block.


protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -45,10 +44,17 @@ impl Executor {
...
-					info!("Transactions in flight: {:?}", in_flight);
-					if in_flight > 2^16 {
-						info!("Shedding load.");
+					let in_flight = self.transactions_in_flight.load(std::sync::atomic::Ordering::SeqCst);
+					info!(
+						target: "movement_timing",
+						in_flight = %in_flight,
+						"transactions_in_flight"
+					);
+					if in_flight > 2u64.pow(16) {
```
musitdev, line 53: Does the compiler optimize `2u64.pow(16) `to its value. If not we can define a const, so the calculus is done one time.

musitdev (approved):

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
mzabaluev: Yes, `pow` is a const fn so it's computed at compile time. Being an old programmer I'd use `1 << 16` though 😁

protocol-units/execution/opt-executor/src/executor/transaction_pipe.rs
```diff
@@ -45,10 +44,17 @@ impl Executor {
...
+						info!(
+							target: "movement_timing",
+							"shedding_load"
+						);
```
mzabaluev, line 59: I don't know if we should be using the timing target for individual events that much. Is this data helpful in the timing log?

PR #268, opened 2024-07-29
Performance Updates w/ Compression, Batching Improvements, and Concurrency Alterations

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `misc`.

Includes changes from #277 and #296 and #299.

# Changelog

# Testing

# Outstanding issues


networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -187,7 +198,7 @@ where
 
 			// check if the block has already been executed
 			if self.has_executed_block(block_id.clone()).await? {
-				warn!("Block already executed: {:?}. It will be skipped", block_id);
+				warn!("Block already executed: {:#?}. It will be skipped", block_id);
```
mzabaluev, line 201: Fancy debug formatting for the block ID?

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -75,23 +80,79 @@ impl LightNodeV1 {
...
-	async fn tick_publish_blobs(&self, receiver : &mut Receiver<Block>) -> Result<(), anyhow::Error> {
-		
-		let half_building_time = self.memseq.building_time_ms();
-		let start = std::time::Instant::now();
-		let mut blocks = Vec::new();
+	/// Attempts to submit a batch of blobs according to the chunk size
+	async fn submit_and_resize(&self, block_blobs: &mut Vec<celestia_types::Blob>, chunk_size : usize) -> Result<usize, anyhow::Error> {
+	
+
+		// iter in chunks of 4
```
mzabaluev, line 95: This comment does not seem to be true, the chunk size is variable?

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -75,23 +80,79 @@ impl LightNodeV1 {
...
+		while !block_blobs.is_empty() {
+			let mut chunk: Vec<celestia_types::Blob> = Vec::with_capacity(chunk_size);
+			
+			for _ in 0..chunk_size {
+				if let Some(blob) = block_blobs.pop() {
+					chunk.push(blob);
+				} else {
+					break;
+				}
+			}
```
mzabaluev, line 105: Could be more concise and preserve the order of blobs:
```suggestion
			let chunk = block_blobs.drain(..min(chunk_size, block_blobs.len())).collect::<Vec<_>>();
```

If the order is not important, `drain` can take from the end, it will work a bit faster than popping one by one.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -75,23 +80,79 @@ impl LightNodeV1 {
...
+
+	/// Submits block blobs in chunks, beginning with a chunk size of 4
+	async fn submit_block_blobs(&self, block_blobs: &mut Vec<celestia_types::Blob>) -> Result<(), anyhow::Error> {
+		if block_blobs.len() > 0 {
 
+			let mut chunk_size = 4;
+			loop {
+
+				let new_chunk_size = self.submit_and_resize(block_blobs, chunk_size).await?;			
+				if new_chunk_size == chunk_size {
```
mzabaluev, line 136: This exit condition assumes particulars from the behavior of `submit_and_resize` method. I'd rather have the enclosing loop as `while block_blobs.len() != 0`, then you can also eliminate the top-level check.

mzabaluev (changes requested):

protocol-units/da/m1/util/src/config/mod.rs
```diff
@@ -175,6 +175,15 @@ impl Config {
 			),
 		}
 	}
+
+	pub fn try_block_building_parameters(&self) -> Result<(u32, u64), anyhow::Error> {
```
mzabaluev, line 179: The return type should be a struct with descriptive field names.
What's the purpose of making this function fallible?

benches/howzit/src/bin/howzit-bench.rs
```diff
@@ -38,9 +38,9 @@ pub async fn main() -> Result<(), anyhow::Error> {
 	howzit.build_and_publish().await?;
 
 	// fund the accounts in an orderly manner
-	let n = 32;
-	let l = 3000;
-	let k = 64;
+	let n = std::env::var("N").unwrap_or("64".to_string()).parse::<usize>()?;
+	let l = std::env::var("L").unwrap_or("3000".to_string()).parse::<u64>()?;
+	let k = std::env::var("K").unwrap_or("64".to_string()).parse::<u64>()?;
```
mzabaluev, line 43: Ehh... Can we have some distinctive prefix on these environment variable names?

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -91,69 +90,98 @@ where
...
+	}
+
 	pub async fn write_transactions_to_da(&self) -> Result<(), anyhow::Error> {
 		loop {
-			self.tick_write_transactions_to_da().await?;
+			// run send transaction batch writes and submit transaction batch writes concurrently
+			match self.send_transaction_batch_writes().await {
+				Ok(_) => {}
+				Err(e) => {
+					error!("Failed to send transaction batch writes: {:?}", e);
```
mzabaluev, line 179: Do we want to continue looping at this point, or is it better to return the error?
If the error is not recoverable, this would result in endless spam in the logs.

l-monninger: It's not unrecoverable for all failure cases. It could be unrecoverable if the `m1-da-light-node` service goes down.

benches/howzit/src/bin/howzit-bench.rs
mzabaluev: `HOWZIT_` is fine.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Is there information on the error that we can use to decide if it's recoverable?
We can try to figure the hard way, of course. But I'd rather whitelist the known benign cases than let the node run broken for indefinite amount time.

l-monninger: I think we would need to start replacing a bunch of `anyhow::Error`s with typed errors. It almost certainly falls out of scope here. Do you want me to just allow the crash?

PR #271, opened 2024-07-30
Integrate block reversion

### LINKED ISSUE #115: Integrate block reversion

mzabaluev:
Closes #115 

# Summary
- **Depends on**: https://github.com/movementlabsxyz/aptos-core/pull/52

Handle `BlockCommitmentEvent::Rejected` from the settlement manager
by reverting to the previous block height if it's smaller than the currently committed height.

# Changelog

* Block rejection events from the MCR settlement contract are handled by reverting the committed optimistic execution state.

# Testing

Currently lacking.

# Outstanding Issues

Need to figure out how to trigger block reversion in e2e test setup, possibly with MCR clients simulating rejection on behalf of 2/3 of validators.


networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -306,7 +307,17 @@ where
 			}
 			BlockCommitmentEvent::Rejected { height, reason } => {
 				debug!("Commitment rejected: {:?} {:?}", height, reason);
-				// TODO: block reversion
+				let current_head_height = executor.get_block_head_height().await?;
+				if height > current_head_height {
+					// Nothing to revert
+				} else {
+					match executor.revert_block_head(height - 1) {
```
l-monninger, line 314: If this is the correct usage, might it be better to name this `revert_block_head_to`?

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -306,7 +307,17 @@ where
...
+						Ok(_) => {}
+						Err(e) => {
+							error!("Failed to revert to block height: {:?}", e);
```
l-monninger, line 317: This might actually be a case where we want to crash. If the node needs to revert but cannot, it will be costly for the operator and the network for it to continue operation. 

It's outside of the current scope, but... 

In cases like these, I think the most graceful thing to do would be to have the node optionally try and delegate before it goes offline. That way there's not a risk to the network of losing active stake and the node operator can continue to earn some rewards.

l-monninger (changes requested):

l-monninger:
Do you want any help trying to figure out how to test this? I could develop a fraud node for you?

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Right, the quick solution is to propagate the error so that the node terminates.

> In cases like these, I think the most graceful thing to do would be to have the node optionally try and delegate before it goes offline.

I will need some pointers on how to do that. But I would treat an error here as an unrecoverable implementation bug.

mzabaluev:
> Do you want any help trying to figure out how to test this? I could develop a fraud node for you?

I don't think we need a full fraud node for this; just simulating settlement failure by >=2/3 of MCR clients in e2e should trigger the reversion flow. We don't care if the transactions in reverted blocks were actually correct.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Made failures fatal in [350e002](https://github.com/movementlabsxyz/movement/pull/271/commits/350e00245c56917afd5cc0623e7b36dadb2e73b2)

musitdev (approved):
It seems ok for me.

l-monninger:
So, I think this is good besides not having e2e tests. In the governance pamphlet I'm working on, one proposal suggests that we handle this kind of testing absence with feature flags.

l-monninger (approved):

Issue #272, opened 2024-07-30
Increment block height in testing may need E2E test workaround


andygolay:
**Describe the bug**
There may not be a way to simulate incrementing block height in Aptos Move tests.

A possible workaround is E2E testing (using a client and a live blockchain) rather than unit testing for:

- `refund_bridge_transfer` in the initiator module and
- `abort_bridge_transfer` in the counterparty module.

**To Reproduce**

See https://github.com/aptos-labs/aptos-developer-discussions/discussions/351

In the `atomic_bridge_initiator` Move module, the test `test_refund_bridge_transfer` is currently annotated as `#[expected_failure]`

**Expected behavior**
`test_refund_bridge_transfer` (initiator-side) and unit tests for `abort_bridge_transfer` (counterparty-side) should be able to pass unit tests but we don't know yet whether it's possible to simulate incrementing the block height in unit tests.

andygolay:
Closed as we are no longer using block height.

PR #274, opened 2024-07-30
Replace timestamp with block height in atomic_bridge_counterparty module

andygolay:
# Summary

Updated `atomic_bridge_counterparty` module to use block height instead of timestamp for time lock.

# Changelog

Substituted 
- `block::get_current_block_height()` for `timestamp::now_seconds()` and 
- `genesis::setup()` for `timestamp::set_time_has_started_for_testing(aptos_framework)`

# Testing

All unit tests pass.

# Outstanding issues

E2E tests should be done for `abort_bridge_transfer` (see issue https://github.com/movementlabsxyz/movement/issues/272 regarding challenges simulating incrementing block height in unit tests).


0xmovses:
Thanks for the effort here, but as discussed, we will be using timestamp going forward.

PR #277, opened 2024-07-31
Heuristics for Celestia

l-monninger:
# Summary
- **RFCs**: $\emptyset$

See discussion: #278 

# Changelog

# Testing

# Outstanding issues


mzabaluev:
Can you provide a brief description on what this new infrastructure does? Doc comments on the main types and abstraction traits would be nice, too.

l-monninger:
> Can you provide a brief description on what this new infrastructure does? Doc comments on the main types and abstraction traits would be nice, too.

This should still be a draft, but: https://github.com/movementlabsxyz/movement/discussions/278

l-monninger:
@mzabaluev there should be a fair number of doc comments on the `grouping_heuristic` module as well.

util/movement-types/src/algs/grouping_heuristic/skip.rs
```diff
@@ -0,0 +1,121 @@
...
+        Box::new(SkipFor::new(count, heuristic))
+    }
+
+    pub fn evaluate(&mut self) {
+        self.skip_if.set_condition(self.counter < self.for_count);
+    }
+
+    pub fn increment_counter(&mut self) {
+        self.counter += 1;
+    }
```
mzabaluev, line 106: These look like internal helpers, maybe they don't need to be public?

This is a more general comment about code style: if you give everything public visibility by default, we'll end up with poorly maintainable API that would be difficult to cut down later.

util/movement-types/src/algs/grouping_heuristic/splitting.rs
```diff
@@ -0,0 +1,116 @@
...
+        // Calculate the chunk size based on the factor
+        self.into_iter()
+        .chunks(chunk_size)
+        .into_iter()
+        .map(|chunk| chunk.collect())
+        .collect() 
+    }
+}
+
+pub mod block {
```
mzabaluev, line 84: This is an empty module now (trait impls are effectively bound to the crate rather than a module), why is it in the public namespace?

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -88,64 +97,77 @@ impl LightNodeV1 {
...
+		let start_distribution = GroupingOutcome::new_apply_distribution(blocks);
+		let block_group_results = heuristic.run_async_sequential_with_metadata(
+			start_distribution,
+			|index, grouping, mut flag| async move {
 
-			let mut chunk_size = 4;
-			loop {
+				if index == 0 {
+					flag = false;
+				}
```
mzabaluev, line 143: This seems like maybe `run_async_sequential_with_metadata` needs to pass the initial metadata value, like `Iterator::fold`. Passing the index around is not strictly necessary in the general case because you can pass an `FnMut` closure that can have synchronous code for the loop iteration, like this:

```rust
move |grouping, mut flag| {
    // logic about the index, maybe copy it to a capture var for the async move closure

    let fut = async move {
        // ...
    };

    index += 1;
    fut
}
```

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -190,30 +212,16 @@ impl LightNodeV1 {
...
+		if blocks.is_empty() {
+			return Ok(());
 		}
-
+		let ids = blocks.iter().map(|b| b.id()).collect::<Vec<_>>();
 
 		// submit the blobs, resizing as needed
 		for block_id in &ids {
 			info!(target: "movement_timing", %block_id, "submitting_block_batch");
 		}
```
mzabaluev, line 223: Why allocate a vector just for logging when you can log while iterating over blocks themselves?
```suggestion
		// submit the blobs, resizing as needed
		for block_id in blocks.iter().map(|b| b.id()) {
			info!(target: "movement_timing", %block_id, "submitting_block_batch");
		}
```

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -88,64 +97,77 @@ impl LightNodeV1 {
...
+	}
 
-		Ok(chunk_size)
+	pub async fn submit_with_heuristic(&self, blocks : Vec<Block>) -> Result<(), anyhow::Error> {
 
-	}
+		let mut heuristic : GroupingHeuristicStack<Block> = GroupingHeuristicStack::new(vec![
+			DropSuccess::boxed(),
+			ToApply::boxed(),
+			SkipFor::boxed(1, Splitting::boxed(2)),
```
mzabaluev, line 132: Since the `SkipFor` combinator is generic, the inner strategy does not need to be dynamically dispatched?
```suggestion
			SkipFor::boxed(1, Splitting::new(2)),
```

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+
+impl <T> Debug for ElementalFailure<T>
+where T: Debug {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ElementalFailure::Instrumental(t) => write!(f, "Instrumental({:?})", t),
+            ElementalFailure::Terminal(t) => write!(f, "Terminal({:?})", t)
+        }
+    }
+}
```
mzabaluev, line 26: This is a poorly re-implemented `#[derive(Debug)]` that does not use `f.debug_tuple(...)` helpers. Just use the derive?

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+impl <T> Debug for ElementalOutcome<T>
+where T: Debug {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            ElementalOutcome::Apply(t) => write!(f, "Apply({:?})", t),
+            ElementalOutcome::Success => write!(f, "Success"),
+            ElementalOutcome::Failure(failure) => write!(f, "Failure({:?})", failure)
+        }
+    }
+}
```
mzabaluev, line 94: Same here, derive is no code and does it better with some edge cases.

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+impl <T> Eq for GroupingOutcome<T> where T: Eq {}
+
+impl <T> Debug for GroupingOutcome<T>
+where T: Debug {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "GroupingOutcome({:?})", self.0)
+    }
+}
```
mzabaluev, line 220: And one more that can well be a derive.

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+
+/// The outcomes for a particular group in a grouping heuristic.
+pub struct GroupingOutcome<T>(pub Vec<ElementalOutcome<T>>);
+
+impl <T> PartialEq for GroupingOutcome<T> where T: PartialEq {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.iter().zip(other.0.iter()).all(|(a, b)| a == b)
+    }
+}
+impl <T> Eq for GroupingOutcome<T> where T: Eq {}
```
mzabaluev, line 213: Just derive `Eq`.

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+}
+
+/// The outcomes for a particular group in a grouping heuristic.
+pub struct GroupingOutcome<T>(pub Vec<ElementalOutcome<T>>);
+
+impl <T> PartialEq for GroupingOutcome<T> where T: PartialEq {
+    fn eq(&self, other: &Self) -> bool {
+        self.0.iter().zip(other.0.iter()).all(|(a, b)| a == b)
+    }
+}
```
mzabaluev, line 212: The impl you'd get with `#[derive(PartialEq)]` does not only get same effect with no code, it's faster because it delegates to slice comparison which is optimized and specialized for bitwise-comparable values, which you can't do outside of the standard library.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -245,7 +245,7 @@ where
 		block_timestamp: u64,
 	) -> anyhow::Result<BlockCommitment> {
 
-		let block: Block = serde_json::from_slice(&block_bytes)?;
+		let block: Block = bcs::from_bytes(&block_bytes)?;
```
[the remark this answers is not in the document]
l-monninger: Yes

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        self.0.iter().all(|outcome| outcome.is_done())
+    }
+
+    /// Converts to inner.
+    pub fn into_inner(self) -> Vec<ElementalOutcome<T>> {
+        self.0
+    }
+
+    /// Into original
+    pub fn into_original(self) -> Vec<T> {
```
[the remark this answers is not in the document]
l-monninger: It produces a `Vec` of the original type instead of a `GroupingOutcome`--essentially unwrapping the inner values.

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        original
+    }
+
+}
+
+impl <T> From<Vec<ElementalOutcome<T>>> for GroupingOutcome<T> {
+    fn from(outcome: Vec<ElementalOutcome<T>>) -> Self {
+        Self {
+            0: outcome
+        }
```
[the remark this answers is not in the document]
l-monninger: Will accept the suggestion.

util/movement-types/src/algs/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,458 @@
...
+        Ok(distribution)
+    }
+
+    /// Runs the grouping heuristic synchronously.
+    pub async fn run(
+        &mut self, 
+        distribution: Vec<GroupingOutcome<T>>,
+        func: impl Fn(GroupingOutcome<T>) -> Result<GroupingOutcome<T>, anyhow::Error> + Send + Sync
+    ) -> Result<Vec<GroupingOutcome<T>>, anyhow::Error> {
```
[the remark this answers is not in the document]
l-monninger: This closure should be synchronous. `Send + Sync` are unnecessary.

util/movement-types/src/algs/grouping_heuristic/skip.rs
```diff
@@ -0,0 +1,121 @@
...
+
+    pub fn new(heuristic: Box<dyn GroupingHeuristic<T>>) -> Self {
+        Skip(heuristic)
+    }
+
+    pub fn boxed(heuristic: Box<dyn GroupingHeuristic<T>>) -> Box<Self> {
+        Box::new(Skip(heuristic))
+    }
+
+    pub fn skip(&mut self, distribution: Vec<GroupingOutcome<T>>) -> Result<Vec<GroupingOutcome<T>>, anyhow::Error> {
```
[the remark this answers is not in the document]
l-monninger: It doesn't need to be mutable. That was just me lazily copying function signatures around. 

This is intended to make a heuristic skip. I'm just using a compositional style.

util/movement-types/src/algs/grouping_heuristic/skip.rs
l-monninger: Here, these should be public because somebody may want to extend the `SkipFor` heuristic wrapper and adjust the way it increments and evaluates, by calling these methods.

util/movement-types/src/algs/grouping_heuristic/splitting.rs
l-monninger: Vestigial. I used to have imports from there.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
l-monninger: I was thinking about the same thing. Really, the metadata value performs a more general version of what the index provides, i.e., tracking and updating non-grouping data. I'll drop the index.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
l-monninger: This is for timing the entire heuristic.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
l-monninger: Good call.

l-monninger: Ah, actually, I think this starts forcing me down a path where I need to use associated types. Do you want me to do such a big rewrite?

l-monninger:
@mzabaluev 

Besides addressing your comments, a couple things I want to add...
1. I was thinking this should be moved out into its own crate.
2. This needs a `Catch` combinator, as right now if the `FirstFit` were to error, we would just stop attempting to send blobs. This is valid behavior generally, but `FirstFit` errors when a single item exceeds capacity by itself--an outcome which can be addressed by the `Splitting` heuristic.

Do those sound reasonable to you?

protocol-units/da/m1/light-node/src/v1/sequencer.rs
l-monninger: Actually, I changed my mind back. `index` does provide information that would be hard to compute otherwise. For example, `index == 0` indicates the start of a new distribution being applied and would be commonly used. It probably even makes sense to provide the distribution length to this callback in the future.

util/movement-algs/src/grouping_heuristic/splitting.rs
```diff
@@ -0,0 +1,121 @@
...
+			// unpack the transactions
+			let Block { metadata, transactions, parent, id: _ } = self;
+
+			// split the vector of transactions
+			let split_transactions = transactions.split(factor);
+
+			// create a new block for each split transaction
+			let mut blocks = Vec::new();
+			for split in split_transactions {
+				let parent = Block::new(metadata.clone(), parent.clone(), split);
```
mzabaluev, line 114: Why is this variable named `parent`?
Is it OK to just replicate the block metadata and parent field for each of the split blocks? Currently, metadata is a unit type, but if it gets more complicated, what should cloning actually mean?

util/movement-types/Cargo.toml
```diff
@@ -19,7 +19,10 @@ serde_with = { workspace = true }
 # derivative = { workspace = true }
 anyhow = { workspace = true }
 blake3 = { workspace = true }
-# hex = { workspace = true }
+itertools = { workspace = true }
+
+[dev-dependencies]
+tokio = { workspace = true }
```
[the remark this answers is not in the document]
mzabaluev: Leftovers from earlier work it seems, these deps are used in movement-algs.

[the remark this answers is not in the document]
l-monninger: Will remove.

util/movement-algs/src/grouping_heuristic/mod.rs
```diff
@@ -0,0 +1,378 @@
...
+		for outcome in self.0 {
+			match outcome {
+				ElementalOutcome::Apply(t) => original.push(t),
+				ElementalOutcome::Success => (),
+				ElementalOutcome::Failure(failure) => match failure {
+					ElementalFailure::Instrumental(t) => original.push(t),
+					ElementalFailure::Terminal(t) => original.push(t),
+				},
+			}
+		}
```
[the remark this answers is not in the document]
l-monninger: Will accept.

util/movement-algs/src/grouping_heuristic/splitting.rs
l-monninger: > Why is this variable named parent?
We are assuming a Block DAG. The parent of the block becomes the part of the split immediately proceeding. For the the 0th part of the split, the parent block is the original parent block.

We don't actually rely on the parent attribute at all at this point. In the future, we may have to consider how we want to preserve relationships between blocks. 

> but if it gets more complicated, what should cloning actually mean?
We kind of don't know until it gets more complicated.

util/movement-algs/src/grouping_heuristic/chunking.rs
```diff
@@ -0,0 +1,72 @@
...
+		// flatten the distribution
+		let mut distribution = distribution
+			.into_iter()
+			.flat_map(|outcome| outcome.into_inner())
+			.collect::<Vec<_>>();
+
+		// chunk the distribution
+		let mut chunks = Vec::new();
+		while !distribution.is_empty() {
+			let chunk = distribution.drain(0..self.size.min(distribution.len())).collect();
```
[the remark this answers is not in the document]
l-monninger: It needs to take ownership and this was before I added `itertools`.  Also, this is quadratic because of the resize of the vector no? Should I just use `VecDeque`, so that I can pop off the front?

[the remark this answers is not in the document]
l-monninger: Does `itertools` take care of this internally...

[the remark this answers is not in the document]
l-monninger: Seems like their `GroupBy` does.

util/movement-algs/src/grouping_heuristic/skip.rs
```diff
@@ -0,0 +1,104 @@
...
+
+	pub fn apply(
+		&mut self,
+		distribution: Vec<GroupingOutcome<T>>,
+	) -> Result<Vec<GroupingOutcome<T>>, anyhow::Error> {
+		self.0.distribute(distribution)
+	}
+}
+
+impl<T> GroupingHeuristic<T> for Skip<T> {
```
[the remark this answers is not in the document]
l-monninger: I'd tried this approach before, but struggled a bit with the unconstrained type `T`. Let me wrangle it.

util/movement-algs/src/grouping_heuristic/drop_success.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl DropSuccess {
+    
+    pub fn new() -> Self {
+        DropSuccess
+    }
+
+    pub fn boxed() -> Box<Self> {
+        Box::new(DropSuccess)
+    }
```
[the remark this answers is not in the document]
l-monninger: It found it easier for developmental purposes, but let me try to wangle.

util/movement-algs/src/grouping_heuristic/skip.rs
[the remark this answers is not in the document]
l-monninger: Yeah, I don't think there's an easy and good way to jiggle these types. The stack setup to me either dictates a dynamic dispatch or an enum, because the members are supposed to be different structs.

mzabaluev (approved):
Looks good with the changes.

PR #294, opened 2024-08-02
Catch All Heuristic

l-monninger:
# Summary

Early attempt at implementing catch-all heuristic.

The hard part here is it requires cloning with the current trait bounds. Even though this behavior is common, I think it may be better to either:

1. Implement a single recover point in a `GroupingHeuristic` stack. So that the cloning only occurs once.
```rust
CatchAll::boxed(vec![
    ToApply::boxed(),
    FirstFit::boxed()
]);
```
2. Even though this is a common/generic operation, just leave the catching to individual variations of distributions. 

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/util/movement-algs/src/grouping_heuristic/catch.rs
+++ b/util/movement-algs/src/grouping_heuristic/catch.rs
@@ -1,0 +1,43 @@
+use crate::grouping_heuristic::{GroupingHeuristic, GroupingOutcome};
+
+/// Heuristic combinator that converts all outcomes to failures if the inner heuristic fails.
+/// This relies on cloning the inner heuristic, so it is advised that this be used with types that can be cloned cheaply.
+pub struct CatchAllToFailure<T>(pub Box<dyn GroupingHeuristic<T>>)
+where
+	T: Clone;
+
+impl<T> CatchAllToFailure<T>
+where
+	T: Clone,
+{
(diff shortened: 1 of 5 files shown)


util/movement-algs/src/grouping_heuristic/catch.rs
```diff
@@ -0,0 +1,43 @@
...
+}
+
+impl<T> GroupingHeuristic<T> for CatchAllToFailure<T>
+where
+	T: Clone,
+{
+	fn distribute(
+		&mut self,
+		distribution: Vec<GroupingOutcome<T>>,
+	) -> Result<Vec<GroupingOutcome<T>>, anyhow::Error> {
```
mzabaluev, line 29: I've got an idea to eliminate the need for cloning here: instead of `anyhow::Error`, make a `DistributeError` type that will give back the passed distribution:

```rust
pub struct DistributeError<T> {
    distribution: Vec<GroupingOutcome<T>>,
    // The actual error, delegated to for Display and Error impls
    cause: anyhow::Error,
}

impl<T> DistributeError<T> {
    /// Recovers the distribution that went into the `distribute` call.
    pub fn recover(self) -> Vec<GroupingOutcome<T>> {
        self.distribution
    }
}
```

PR #296, opened 2024-08-02
Deterministic Timestamp Adjustments for VM Failure Cases

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Executes a block with retries. Deterministically incrementing the timestamp to allow for the DA to send blocks through with narrow gaps in timestamps. 

# Changelog

# Testing

# Outstanding issues


l-monninger:
Yes, this works to prevent crashes. I'm going to see if I can solve the actual issue with the timestamps though...

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -238,13 +246,34 @@ where
 		Ok(())
 	}
 
+	/// Retries executing a block several times.
+	/// This can be valid behavior if the block timestamps are too tightly clustered for the full node execution.
+	/// However, this has to be deterministic, otherwise nodes will not be able to agree on the block commitment.
+	async fn execute_block_with_retries(
```
mzabaluev, line 252: If the erstwhile `execute_block` is no longer called, remove it and rename this to `execute_block`.
Otherwise this produces dead code that's only going to slow down maintenance.

l-monninger: Wdym? This calls `execute_block`?

mzabaluev: Oh, my bad.

mzabaluev (approved):
LGTM

PR #297, opened 2024-08-02
0xmovses/bridge integration tests

0xmovses:
# Summary
Adds new package `bridge-integration-tests`

# Changelog
- Adds integration tests for `EthClient` contract call handlers. 

# Testing
Checks / bridge-eth-movement-client has been added to CI.

# Outstanding issues
- Note that `test_client_should_successfully_get_bridge_transfer_id` and `test_client_should_successfully_complete_transfer` have been ignored. They can be ran after https://github.com/movementlabsxyz/movement/pull/209 is merged in, as we need the event logging to capture the bridge transfer ids. 

- Also note that the `TestHarnes` cannot yet be ran in parallel. Rust be default runs test in paralell, so for now `--test-threads=1` has to be passed to the command. The underlying rpc_provider on the and the EthClient is `Clone` and when cloned shared the underlying connection.


andygolay:
@0xmovses when I navigate to `integration-tests` and run `cargo test` I get an "Address already in use" error leading to the failure of most tests. 

Here's the full output in case it's helpful - is some additional configuration needed in advance before running the tests?

```
running 5 tests
Error: Address already in use (os error 48)

Location:
    /Users/runner/work/foundry/foundry/crates/anvil/src/cmd.rs:278:33
Error: Address already in use (os error 48)

Location:
    /Users/runner/work/foundry/foundry/crates/anvil/src/cmd.rs:278:33
Error: Address already in use (os error 48)

Location:
    /Users/runner/work/foundry/foundry/crates/anvil/src/cmd.rs:278:33
Error: Address already in use (os error 48)

Location:
    /Users/runner/work/foundry/foundry/crates/anvil/src/cmd.rs:278:33
test test_client_should_successfully_get_bridge_transfer_id ... FAILED
test test_client_should_build_and_fetch_accounts ... FAILED
test test_client_should_successfully_complete_transfer ... FAILED
test test_client_should_successfully_call_initiate_transfer ... FAILED
test test_client_should_deploy_contract ... FAILED

failures:

---- test_client_should_successfully_get_bridge_transfer_id stdout ----
Anvil running at `http://localhost:8545`
thread 'test_client_should_successfully_get_bridge_transfer_id' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:164:10:
Failed to initiate bridge transfer: GenericError("Initiator contract address not set")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- test_client_should_build_and_fetch_accounts stdout ----
thread 'test_client_should_build_and_fetch_accounts' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:27:58:
called `Result::unwrap()` on an `Err` value: Timeout

---- test_client_should_successfully_complete_transfer stdout ----
thread 'test_client_should_successfully_complete_transfer' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:180:58:
called `Result::unwrap()` on an `Err` value: Timeout

---- test_client_should_successfully_call_initiate_transfer stdout ----
thread 'test_client_should_successfully_call_initiate_transfer' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:84:58:
called `Result::unwrap()` on an `Err` value: Timeout

---- test_client_should_deploy_contract stdout ----
thread 'test_client_should_deploy_contract' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:62:58:
called `Result::unwrap()` on an `Err` value: Timeout


failures:
    test_client_should_build_and_fetch_accounts
    test_client_should_deploy_contract
    test_client_should_successfully_call_initiate_transfer
    test_client_should_successfully_complete_transfer
    test_client_should_successfully_get_bridge_transfer_id

test result: FAILED. 0 passed; 5 failed; 0 ignored; 0 measured; 0 filtered out; finished in 10.00s

error: test failed, to rerun pass `--test eth_movement`
```

0xmovses:
@andygolay `cargo test` will not work. Run the full command in the description, the tests are not multi-threaded `cargo test --test eth_movement -- --nocapture --test-threads=1` 

If you still get the error 
`lsof -i :8545 `
`kill <PID>`

I'm still working on this PR so have put it into draft for now. Will ping when ready for review

protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -91,17 +105,44 @@ impl EthClient {
...
+		self.counterparty_contract = Some(contract);
+	}
+
+	pub async fn initialize_initiator_contract(
+		&self,
+		weth: EthAddress,
+		owner: EthAddress,
+	) -> Result<(), anyhow::Error> {
+		let contract =
+			AtomicBridgeInitiator::new(self.initiator_contract_address()?, &self.rpc_provider);
```
0xPrimata, line 137: Seems like you are initializing the implementation contract here. It should impede you from using it as a functional contract but there might be something you are trying to call that isn't included in the implementation contract, leading to a fail to estimate gas error

0xmovses: Yeah I tried before storing the contract instance in the Client. But I get the same error. This doesn't create a new contract, but instead creates a wrapped around the onchain instance. So state should persist. This is also the same pattern in MCR settlement.

0xmovses: This call does internally call weth, but that the WETH interface is imported inside `AtomicBridgeInitiator`, and so its calls are encoded into `AtomicBridgeInitiator` abi `bytecode` field, so it should be fine.

0xmovses: I've just tested this again and updated it to use the method I created that fetches the previously initialised contract instance:
```rust
let contract = self.initiator_contract().expect("Initiator contract not set");
		let recipient_bytes: [u8; 32] =
			recipient_address.0.try_into().expect("Recipient address must be 32 bytes");
		let call = contract
			.initiateBridgeTransfer(
				U256::from(0), // For now a 0 WETH amount
				FixedBytes(recipient_bytes),
				FixedBytes(hash_lock.0),
				U256::from(time_lock.0),
			)
```
So fetching the same initialised contract. It fails with the same error.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -0,0 +1,155 @@
...
+		assert_eq!(account, expected);
+	}
+}
+
+#[tokio::test]
+async fn test_client_should_deploy_initiator_contract() {
+	let mut harness: TestHarness = TestHarness::new_only_eth().await;
+	let anvil = Anvil::new().port(harness.rpc_port()).spawn();
+
+	// Set a funded signer
```
andygolay, line 49: It seems to me like the `// Set a funder signer` comments throughout this file aren't needed as the next line is fairly self-descriptive.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -0,0 +1,155 @@
...
+#[ignore] // To be tested after this is merged in https://github.com/movementlabsxyz/movement/pull/209
+async fn test_client_should_successfully_complete_transfer() {
+	let mut harness: TestHarness = TestHarness::new_only_eth().await;
+	let anvil = Anvil::new().port(harness.rpc_port()).spawn();
+
+	// set funded signer
+	let signer_address = harness.set_eth_signer(anvil.keys()[0].clone());
+	harness.deploy_init_contracts().await;
+
+	//some data to set for the recipient.
```
andygolay, line 133: This comment doesn't seem necessary as the next couple lines are self-describing and the comment doesn't add much in the way of clarification.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -0,0 +1,155 @@
...
+			InitiatorAddress(EthAddress(signer_address)),
+			RecipientAddress(recipient),
+			HashLock(hash_lock),
+			TimeLock(100),
+			Amount(1000), // Eth
+		)
+		.await
+		.expect("Failed to initiate bridge transfer");
+
+	//TODO: Here call get details with the captured event
```
andygolay, line 120: It it okay to have this left as TODO? I see the comment about merging PR #209 first so all good if that's the case.

andygolay (approved):
Logic of integration tests makes sense, but I think the TODOs need to be done to really be sure. Pending merging PR #209 and finishing fleshing out tests, this is looking good. 

One consideration: These all appear to be happy path tests. To make the test suite more comprehensive, do you think we should include sad path tests? This could be for next sprint.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
0xmovses: Yes, let's leave them in for now. They will be gone quite soon.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
0xmovses: Will remove

0xmovses:
@andygolay Good point, could you create an issue about sad paths 😢  for the integration tests? We can add some in if time allows, i'd say it's slightly lower pri for now

PR #299, opened 2024-08-03
Block Compression and Channel-based Transaction Batching

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Adds block compression and Channeld-based Transaction Batching. 

# Changelog

# Testing

# Outstanding issues
1. The heuristic may rerun compression logic--which can be hefty and in the heuristic case is no currently `spawn_blocking`.


networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -147,30 +150,61 @@ where
...
 
 		Ok(())
 	}
 
+	pub async fn send_transaction_batch_writes(
+		&self,
+		sender: Sender<BatchWriteRequest>,
+	) -> Result<(), anyhow::Error> {
+		loop {
+			self.next_transaction_batch_write(sender.clone()).await?;
```
mzabaluev, line 172: Useless clone:
```suggestion
			self.next_transaction_batch_write(sender).await?;
```

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -147,30 +150,61 @@ where
...
+				light_node_client.batch_write(batch_write).await?;
+				Ok::<(), anyhow::Error>(())
+			});
+		}
+	}
+
 	pub async fn write_transactions_to_da(&self) -> Result<(), anyhow::Error> {
 		loop {
-			self.tick_write_transactions_to_da().await?;
+			let (batch_sender, batch_receiver) = async_channel::unbounded();
```
mzabaluev, line 193: Try replacing `async_channel` with `tokio::sync::mpsc`. However here, it's dubious we need a channel at all yet, see below.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -147,30 +150,61 @@ where
...
+
+			// run send transaction batch writes and submit transaction batch writes concurrently
+			match tokio::try_join!(
+				self.send_transaction_batch_writes(batch_sender),
+				self.submit_transaction_batch_writes(batch_receiver)
+			) {
```
mzabaluev, line 199: So these two futures communicate over the (unnecessarily over-synchronized) channel, but are driven preemptively as one task by `try_join!`. I think this is just useless overhead now.

If you spawn these two as tasks, you may get real benefits on the multithreaded executor. But this would require state for each of them to be taken out of `self`, so it might be a taller order.

networks/suzuka/suzuka-full-node/src/partial.rs
```diff
@@ -215,12 +249,18 @@ where
 				anyhow::bail!("Invalid DA height: {:?}", da_height);
 			}
 
+			// decompress the block bytes
+			let block = tokio::task::spawn_blocking(move || {
+				let decompressed_block_bytes = zstd::decode_all(&block_bytes[..])?;
```
mzabaluev, line 254: It seems I was mistaken about the format freeze and we get to change the blob's inner format without resetting the chain. Will this really work seamlessly?

protocol-units/da/m1/light-node/src/v1/sequencer.rs
```diff
@@ -97,39 +94,46 @@ impl LightNodeV1 {
...
-		}
-		for block in blocks {
-			info!(target: "movement_timing", block_id = %block.id(), "submitting_block");
+			info!(target: "movement_timing", block_id = %block.block.id(), "inner_submitting_block");
 		}
+		// get references to celestia blobs in the wrapped blocks
+		let block_blobs = blocks
+			.iter()
+			.map(|wrapped_block| &wrapped_block.blob)
+			.cloned() // hopefully, the compiler optimizes this out
```
mzabaluev, line 105: My hunch is that the compiler can't optimize this, it needs the original `Vec` for the logging iteration below.

Maybe this is the case, converse to what I previously suggested, for collecting the IDs in a temporary `Vec` with a non-desctructive pass, and then moving the blob fields into their own `Vec` with `into_iter`. `itertools` may have an adapter for this.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: The benefit here is that the batches keep building without waiting for the submission of the `BatchWriteRequest` to occur. This is not intended for optimization viz the runtime, but more logical optimization w.r.t. the the throughput of the `m1-da-light-node` service.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: Like I said last week, I was looking into the ramifications and this is indeed fine if we prune the DA. It means that our partners need a trust start state, but they were going to need one anyone.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: Also, I don't think there's anything particularly CPU bound here. So, what would be the benefit? It's basically waiting on a channel--which the futures as coroutines should be good at--and sending requests to `m1-da-light-node` which coroutines should also be good at.

To my knowledge, spawn can be better for more CPU-intensive tasks because it may spawn an additional green thread for the task. However, shouldn't the benefit be negligible here?

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: The value is moved in the previous iteration of the loop. Do you want me to return the sender?

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Yes, the futures polled concurrently are still beneficial, but this can be done without a channel (which is now just a buffer with synchronization overhead), e.g. with `select!`.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
mzabaluev: A plain old loop with `push`ing two vectors would be even simpler.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: Yeah, should be `mpsc` not `mpmc`.

networks/suzuka/suzuka-full-node/src/partial.rs
l-monninger: I'm not seeing how `select!` replaces anything the current design. I think I would effectively need to `select!`, `join_all!`, or `join_each_concurrent` on a dynamic join set. I usually opt for a channel and a worker over trying something exotic for that kind of pattern

However, if you don't want the channel, I can just spawn and not await the submission task from the batching loop.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
l-monninger: Let me take a look at something here...

l-monninger: Yeah, my initial belief is that there wasn't really a way around this because we the client ultimately relies on taking a slice of owned `Blobs`. But, under the hood, the client must be serializing those blobs, which for most libs can be done with a reference. So, I was hoping the compiler could pick up on this. But, that doesn't seem correct. 

The id logging is not the issue I was contending with here; that's always pretty easy to move out. It's actually the heuristic and retries which require I provide a new distribution upon failure. The only ways I see to do this--because the third-party client takes and consumes the `Blobs`--is to clone or instead pass down a reference and ultimately clone. The latter results in less cloning because--as in this implementation--I don't have to clone the whole `WrappedBlock` only the `Blob`. So, I opted for that. 

I also tried playing around with the `Deref` trait on the `WrappedBlock`. But, I didn't get anything I liked.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: Sorry, I missed the loop when I looked at it yesterday.
But since `next_transaction_batch_write` is a helper method and borrows `self` anyway, it could borrow `sender` as well? These are all micro-optimizations unlikely to have major impact, anyway.

networks/suzuka/suzuka-full-node/src/partial.rs
mzabaluev: I thought of having a loop with `select!` in the body where one `Stream` object would yield batches and another future consume the available batch.

But if it's easier to do over a channel now, let it be.

protocol-units/da/m1/light-node/src/v1/sequencer.rs
mzabaluev: Ah I see, this is already with heuristics and blocks need to remain past the `submit_blocks` call in the failure case. On the other hand, the Celestia client takes a slice of blobs and we can't change that. In this case, fair to copy. Unless we can split blobs from IDs and pass them as parallel collections, though this will be ugly.

mzabaluev (approved):

Issue #211, opened 2024-07-16
Create tooling with scripts and/or process-compose for having Move tests run in CI


0xmovses:
Checkout this comment https://github.com/movementlabsxyz/movement/pull/205#issuecomment-2231144839

We want to be able to run automated tests on our move modules in CI. This will ensure updates wont break any dependant code. As of now we have to manualy update the `Move.toml` between running tests. 

I think there would be a way to dynamically update the Move.toml with the correct values, or maybe there is a nifty `aptos` CLI command. 

- Add the new tooling and a new job into CI to run the tests added in this PR https://github.com/movementlabsxyz/movement/pull/205

0xmovses:
Done in https://github.com/movementlabsxyz/movement/pull/304

PR #304, opened 2024-08-05
Move tests run in CI

andygolay:
# Summary
Solution for [Issue #211](https://github.com/movementlabsxyz/movement/issues/211) to make Move tests run automatically in CI

# Changelog

* Add Move test workflow to `./github/workflows/checks.yml`
* Add `./github/scripts/update_move_toml.sh` script to run `aptos init` and update `Move.toml` with address generated
* Update `./protocol-units/bridge/move-modules/Move.toml` to use our fork of `aptos-core`

# Testing

* All CI checks pass. 
* When I first ran the workflow successfully, it exposed a Move test failure which appears to have been due to using Aptos Labs' `aptos-core` GitHub repo as a dep, rather than our fork, hence updating `Move.toml` as part of this PR.
 
# Outstanding issues

* Next iteration will replace Aptos CLI with Movement CLI. I'd like to first create more convenient installation options for Movement CLI such as Homebrew and the Python script used in this PR to install Movement CLI, and then switch. Showing that this workflow passes with Aptos CLI provides a baseline level of validation so I think it makes sense as a small PR.

### DIFF
--- a/.github/scripts/update_move_toml.sh
+++ b/.github/scripts/update_move_toml.sh
@@ -2,9 +2,8 @@
 
 MOVE_TOML_PATH="protocol-units/bridge/move-modules/Move.toml"
 
-INIT_OUTPUT=$(aptos init)
+INIT_OUTPUT=$(aptos init 2>&1)
 
-echo "Aptos init output:"
 echo "$INIT_OUTPUT"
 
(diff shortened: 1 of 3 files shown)


0xmovses:
:shipit:

PR #334, opened 2024-08-11
Update types.rs

HeroLeft:
To be more flexible

# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/shared/src/types.rs
+++ b/protocol-units/bridge/shared/src/types.rs
@@ -36,6 +36,12 @@ where
 	}
 }
 
+impl Into<BridgeTransferId> for u64 {
+    fn into(self) -> BridgeTransferId {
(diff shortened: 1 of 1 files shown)


mzabaluev (commented):
What are the use cases for these changes?
Do we really want arbitrary integers and arbitrary bags of bytes to be convertible to these types?

protocol-units/bridge/shared/src/types.rs
```diff
@@ -36,6 +36,12 @@ where
 	}
 }
 
+impl Into<BridgeTransferId> for u64 {
```
mzabaluev, line 39: The idiomatic way in Rust is to implement `From` rather than `Into`. The latter is then available via the blanket implementation.

mzabaluev (changes requested):

Issue #338, opened 2024-08-12
Restrict abort calls to owner on Bridge sol contracts and move modules


0xmovses:
See the comment that came up in review. 
https://github.com/movementlabsxyz/movement/pull/215#discussion_r1713679237

Points: 1

0xmovses:
Merged in #342

PR #342, opened 2024-08-12
Primata/338 only owner

0xPrimata:
# Summary
- **RFCs**: https://github.com/movementlabsxyz/movement/issues/338
- **Categories**:`protocol-units`

add nonce comment, add onlyOwner, add ownership to movetoken, fix tests

# Changelog

# Testing

tests are fixed

# Outstanding issues


0xmovses:
There are changes here to MOVE.token but I think that's outside the scope of this PR. In the Issue the change should just be for Initiator and Counterparty contracts on both Eth and .move modules, maybe the Move word was confusing!

0xmovses:
Could you also add both a solidity unit test where a non-owner tries to call `abortBridgeTransfer`? And also the equivalent unit test on the .move Initiator module. Thanks!

0xPrimata:
Could you rereview @0xmovses ?

0xmovses:
Is this ready for review?

0xmovses (changes requested):
This looks good, but could you remove all the changes for `MCR.sol` and `MovementStaking.t.sol` I think they are just formatting changes. It will cause reviewers of the feature branch to do a double take if we leave it in.

0xPrimata:
all resolved

0xmovses:
Nice, thank you.

PR #366, opened 2024-08-15
Remove suzuka and dot-movement dependencies

0xmovses:
# Summary
The presence of `suzuka-config` as a dependency of `movement-bridge` _and_ `integration-tests` was causing rust-analyzer to not function correctly for these crates. Why, I'm not sure. But when removing it, we get all the compiler warnings and errors expected for `movement-bridge`, and `integration-tests`. 

This weird analyzer behaviour was confirmed by @andygolay, so was not a local-machine issue. 

# Changelog
- Remove `suzuka-config`, as a dependency
- Apply refactors as a result of the removal.

# Testing
- Tests should pass
- Go into the `integration-tests/tests/eth_movement.rs`, compiler warnings should now be seen in your IDE 🥳 

# Outstanding issues
@andygolay you'll now see all the various compiler warnings and clippy warnings that were not showing before on your new tests. If you could apply the fixes for `eth_movement.rs` in a separate PR, such as removing unused imports that would be awesome.


andygolay (approved):
LGTM

Issue #373, opened 2024-08-16
Add Experimental Delegation to MCR Contract for Conceptual Review


l-monninger:
**Is your feature request related to a problem? Please describe.**
Per discussion #371, the time has come to implement MCR. This request is for an initial experimental implementation to faciliate further discussion. 

**Describe the solution you'd like**
1. SHOULD implement the multi-indirection reference in #371. 
2. MUST update the stake method in `MovementStaking` to accept a parameter `delegate` which will specify the delegate behind which we stake. 
3. MUST update the unstake method in `MovementStaking` to accepter a parameter `delegate` which will specify the delegate from which we unstake. 
4. MUST update the stake weights computed to respect the delegate values.

l-monninger:
Delegation is only relevant to decentralized launch.

PR #378, opened 2024-08-19
Delegation (Experimental)

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Issues**: #373

# Changelog
1. Adds methods for staking and unstaking with delegation.
2. Adjusts stake weight computation accordingly. 

# Testing
1. Additional `testBasicDelegation` unit test covers staking and unstaking with delegation. 
2. All previous MCR unit tests unmodified and continue to pass. 

# Outstanding issues
1. Naming could be clearer.
3. Plenty of gas optimizations available.
4. Should delegatees be able to restrict 
5. The term "delegatees" is used instead of delegates to match OpenZeppelin verbiage. "Delegate" is generally more common however. 
6. A more complete delegation sim is probably advisable.


protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -70,69 +70,57 @@ contract MovementStaking is
...
-                    domain,
-                    getCurrentEpoch(domain),
-                    custodian,
-                    attester,
-                    attesterStake
-                );
+                // for every delegatee of the attester
+                for (
+                    uint256 k = 0;
+                    k < delegatorsByAttesterByDomain[domain][attester].length();
```
mzabaluev, line 76: I don't know how good the optimizer is, but perhaps getting the `delegatorsByAttesterByDomain[domain][attester]` reference for the duration of the loop would save some gas.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -141,9 +129,25 @@ contract MovementStaking is
...
-        epochStakesByDomain[domain][epoch][custodian][attester] -= amount;
+        epochStakesByDomain[domain][epoch][custodian][attester][
+            delegator
+        ] -= amount;
+
+        if (
+            epochStakesByDomain[domain][epoch][custodian][attester][
+                delegator
+            ] == 0
+        ) {
```
mzabaluev, line 143: Here as well: the value has just been decremented, could we hold on to the reference and get the value instead of spending gas on 5 nested lookups another time?

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
l-monninger: Yeah, I think it will actually save some gas.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
l-monninger: Yep, good call.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -63,23 +63,26 @@ contract MovementStaking is
...
                 // for every delegatee of the attester
-                for (
-                    uint256 k = 0;
-                    k < delegatorsByAttesterByDomain[domain][attester].length();
-                    k++
-                ) {
-                    // todo: can this be replaced with _rollOverAttester?
-                    address delegatee = delegatorsByAttesterByDomain[domain][
+                EnumberableSet.AddressSet
+                    storage delegatees = delegatorsByAttesterByDomain[domain][
```
mzabaluev, line 80: Looks much better, aesthentically and I assume also gas-wise. I'm not an expert on Solidity, though.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -132,15 +135,15 @@ contract MovementStaking is
...
         uint256 amount
     ) internal {
+        uint256 stake = epochStakesByDomain[domain][epoch][custodian][attester][
+            delegator
+        ];
+        stake -= amount; // okay to let this error if it underflows
         epochStakesByDomain[domain][epoch][custodian][attester][
             delegator
-        ] -= amount;
+        ] = stake;
```
mzabaluev, line 144: You could still use a storage reference here? This change only replaces one 5-step lookup for another.

protocol-units/settlement/mcr/contracts/test/staking/MovementStaking.t.sol
```diff
@@ -429,4 +435,125 @@ contract MovementStakingTest is Test {
         );
         staking.reward(attesters, amounts, custodians);
     }
+
+    function testBasicDelegation() public {
+        MOVEToken moveToken = new MOVEToken();
+        moveToken.initialize();
+
+        MovementStaking staking = new MovementStaking();
+        staking.initialize(moveToken);
```
0xPrimata, line 444: I would advise against building tests that initalize the implementation contract and not a proxy contract. This might lead to clash in functions or impredictable behavior. There has been recently a hack on deltaprime that was caused by this.
I'd be glad to redesign these tests.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -141,9 +132,25 @@ contract MovementStaking is
...
         address attester,
+        address delegator,
         uint256 amount
     ) internal {
-        epochStakesByDomain[domain][epoch][custodian][attester] -= amount;
+        uint256 stake = epochStakesByDomain[domain][epoch][custodian][attester][
+            delegator
+        ];
+        stake -= amount; // okay to let this error if it underflows
+        epochStakesByDomain[domain][epoch][custodian][attester][
```
0xPrimata, line 142: funds get stuck if it underflows, better handle it with "if amount > _stake, _stake = 0"

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -63,76 +49,45 @@ contract MovementStaking is
...
+            address attester = attesters.at(i);
+            // for every delegatee of the attester
+            EnumerableSet.AddressSet storage delegatees = delegatorsByAttesterByDomain[domain][attester];
+            for (uint256 j = 0; j < custodians.length(); j++) {
+                address custodian = custodians.at(j);
+
+                for (uint256 k = 0; k < delegatees.length(); k++) {
+                    // todo: can this be replaced with _rollOverAttester?
+                    address delegatee = delegatees.at(k);
+
```
0xPrimata, line 65: to use _rollOverAttester, it needs to start using nextEpoch param instead of currentEpoch. This way 0 can be passed as the param. This requires a modification of the rest of the code, but definitely seems okay.
Just rolling with the current implementation basicTest passes, it requires a test to verify if the correct epoch has been settled.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
```diff
@@ -63,76 +49,45 @@ contract MovementStaking is
...
+                    // get the genesis stake for the attester
+                    uint256 attesterStake = getStakeAtEpoch(domain, 0, custodian, delegatee, attester);
+
+                    // roll over the genesis stake to the current epoch
+                    _addStake(domain, getCurrentEpoch(domain), custodian, delegatee, attester, attesterStake);
+                }
             }
```
0xPrimata, line 72: If _rollOverAttester is to be used, it might be optimized to take storage params instead

protocol-units/settlement/mcr/contracts/test/staking/MovementStaking.t.sol
l-monninger: That's probably best for a separate PR as literally all of the tests use this approach.

protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol
l-monninger: This is outdated, but I was doing this to save the extra stake lookups later on. You do also in theory save one instruction in the bytecode here.

protocol-units/settlement/mcr/contracts/test/staking/MovementStaking.t.sol
l-monninger: Can you provide some links here. It would seem to me you'd have to be doing something exotic with your use of the proxy pattern to actually have behavior vary in ways that tests like these wouldn't assert.

l-monninger: Or, in the least, you would not need to check all of the logic and instead access controls.

0xPrimata: https://medium.com/@Delta_Prime/deltaprime-post-mortem-report-752bd60a25e6
I think it's a bit further than what I expected but I think that anything that gets us closer to the actual implementation might prevent us from running into a critical bug in production.

PR #391, opened 2024-08-19
Primata/338 cli2bridge

Primata:
# Summary

Includes internal API for clients to make movement calls to the bridge contract initiator and counterparty

# Changelog

# Testing

# Outstanding issues


andygolay (commented):
What's the plan for testing this?

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -61,8 +64,9 @@ impl Config {
 			rpc_url: Some("http://localhost:8080".parse().unwrap()),
 			ws_url: Some("ws://localhost:8080".parse().unwrap()),
 			chain_id: 4.to_string(),
-			signer_private_key: Arc::new(RwLock::new(LocalAccount::generate(&mut rng))),
+			signer_private_key: LocalAccount::generate(&mut rng).private_key().to_string(),
 			initiator_contract: None,
```
0xPrimata, line 68: Still struggling to understand Arc here, if I want to modify the signer_private_key instance, should I istantiate a new Arc::new(RwLock::new(LocalAccount::from(parse the key here?))) wouldn't the instantiation across the app break since we are changing an Arc on the fly and every existing Arc would outlive the modifiaction?

0xPrimata: @0xmovses

0xmovses: `Config` doesn't need to be `Clone.` But `MovementClient` needs to be `Clone` so all its inner types need to be `Clone`, inside the `Arc` is a `RwLock`, this means that many threads can read it at once but only one thread can write to it. 

The only truth that must hold up is that `MovementClient` must be `Clone`. 

I would probably not use an Arc type on the config but then you will need to construct one when building the `MovementClient` signer field. Also you will need to make sure any modifications do not introduce any regression into the tests.

0xmovses:
Thanks for this, could you include a PR Description and test code inside otherwise its hard to know the code is working as intended. https://github.com/movementlabsxyz/movement/blob/primata/338-cli2bridge/protocol-units/bridge/cli/tests/integration_tests.rs

0xmovses:
You need to merge in the feature branch to get all the cargo fmt changes into your branch, otherwise it makes this quite cumbersome to review!

PR #399, opened 2024-08-20
Integrate top level bridge service

0xmovses:
# Summary
Integrates top level bridge service

# Changelog
- Adds all the code necessary to support the `bridge-service` binary which uses both the `EthClient` and `MovementClient`
- Some updates to types and renaming of fields to improve semantics. 

# Testing
1. `cargo build -p bridge-service`
2. `cargo test --test eth_movement -- --nocapture --test-threads=1`
3. `cargo test --test movement_eth -- --nocapture --test-threads=1 `

# Outstanding issues
- 
- Event State Machine yet to be implemented, incoming in @musitdev's PR.


protocol-units/bridge/chains/ethereum/src/lib.rs
```diff
@@ -36,79 +38,85 @@ pub enum Transaction<A, H> {
...
-	H: From<HashLockPreImage>,
-{
-	pub fn new(mut rng: R, name: impl Into<String>) -> Self {
+impl EthereumChain {
+	pub async fn new(name: impl Into<String>, rpc_url: &str) -> Self {
 		let accounts = HashMap::new();
 		let events = Vec::new();
-		let (event_sender, event_receiver) = mpsc::unbounded();
+		let (_, event_receiver_1) = mpsc::unbounded();
+		let (_, event_receiver_2) = mpsc::unbounded();
```
musitdev, line 66: If you drop the sender, the listener won't ever get a message because you can't send them.
I've corrected it in the PR #624. I remove this part and correct in the client side. See previous message

protocol-units/bridge/chains/movement/src/event_monitoring.rs
```diff
@@ -0,0 +1,336 @@
...
+	type Address = MovementAddress;
+	type Hash = MovementHash;
+}
+
+impl MovementInitiatorMonitoring<MovementAddress, MovementHash> {
+	pub async fn build(
+		_rest_url: &str,
+		_listener: UnboundedReceiver<MovementChainEvent<MovementAddress, MovementHash>>,
+	) -> Result<Self, anyhow::Error> {
+		todo!()
```
musitdev, line 38: I've implemented this part in PR #624.

protocol-units/bridge/chains/ethereum/src/event_monitoring.rs
```diff
@@ -0,0 +1,482 @@
...
+			.event("BridgeTransferInitiated(bytes32,address,bytes32,uint256)")
+			.event("BridgeTransferCompleted(bytes32,bytes32)")
+			.from_block(BlockNumberOrTag::Latest);
+
+		let sub = ws.subscribe_logs(&filter).await?;
+		let mut sub_stream = sub.into_stream();
+
+		// Spawn a task to forward events to the listener channel
+		let (sender, _) =
+			tokio::sync::mpsc::unbounded_channel::<EthChainEvent<EthAddress, EthHash>>();
```
musitdev, line 61: If the listener is dropped, sent message won't be read. There's an issue with the listener in parameter and this send. They aren't connected, but the code supposes as they are. I've corrected in the PR #624.

musitdev (approved):
I comment one issue I found developing the PR #624. As I remember, I've corrected others. I think it's not important to merge them. It depends on the PR #624 result. If we merge it, we don't need this part anymore. So it's better to merge this one and correct depending on the other PR result.

protocol-units/bridge/chains/movement/src/event_monitoring.rs
0xmovses: Yep, that's why it's a todo!() here

protocol-units/bridge/chains/ethereum/src/event_monitoring.rs
0xmovses: Yes a noticed this, as separate listeners, if you've corrected in 624, this can go in like this and then 624 will follow up when ready.

andygolay:
@0xmovses after re-adding the `cicd:bridge` label I got this on `Checks / solidity-bridge-tests`:

```
Compiler run failed:
Error (6160): Wrong argument count for function call: 5 arguments given but expected 6.
  --> src/AtomicBridgeCounterparty.sol:74:14:
   |
74 |         emit BridgeTransferLocked(bridgeTransferId, recipient, amount, hashLock, counterpartyTimeLockDuration);
   |              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

Error: 
Compilation failed
Error: Process completed with exit code 1.
```

PR #429, opened 2024-08-22
Add shared bridge lib tests to CI

0xmovses:
# Summary
Noticed these tests were not in our CI. At some point they failed and we did not catch it. It's important that this shared lib tests are always ran as the entire bridge service implementation is based off it. 

# Changelog
- Adds the tests for `bridge/shared/*` to our CI

# Testing
- Runs in CI

# Outstanding Issues
None


andygolay (approved):
LGTM

Issue #405, opened 2024-08-21
Get Bridge Details implementation for Move side of the bridge


Primata:
**Is your feature request related to a problem? Please describe.**
get_bridge_details() is a required method for the bridge service and for general composability. It has to be implemented for the Move modules.

**Describe the solution you'd like**
add get_bridge_details(bridgeId), returns a a bridge attempt data.

PR #434, opened 2024-08-22
Primata/405 move get bridge transfer details

0xPrimata:
# Summary
- **RFCs**: https://github.com/movementlabsxyz/movement/issues/405
- **Categories**:  `protocol-units`

Handles move get bridge details by calling view function bridge_transfers(). It takes a bridge_transfer_id variable.

We might want to pay attention since it copy ability had to be added to struct BridgeTransfers

Because bridge_transfer requires a single SmartTable, counterparty had to be refactored.

# Changelog

fix init_modules functions
fix tests
complete, refund or cancel bridge transfers do not delete transfers anymore
refactor
bridge_transfers view function to retrieve bridge_transfer details by specifying a transfer id

# Testing

Tests are fine. 

# Outstanding issues

We might want to look to test thoroughly Counterparty modules as they are not completely covered.


protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
```diff
@@ -72,6 +72,25 @@ module atomic_bridge::atomic_bridge_counterparty {
...
+        let config_address = borrow_global<BridgeConfig>(@atomic_bridge).bridge_module_deployer;
+        let store = borrow_global<BridgeTransferStore>(config_address);
+        if (!aptos_std::smart_table::contains(&store.transfers, bridge_transfer_id)){
+            BridgeTransfer {
+                amount: 0,
+                originator: @atomic_bridge,
+                recipient: vector::empty<u8>(),
+                hash_lock: vector::empty<u8>(),
+                time_lock: 0,
+                state: 0,
```
0xmovses, line 87: If there is no `BridgeTransfer` for its key `bridge_transfer_id`, we would want to return an error here, instead of creating a zeroed `BridgeTransfer`.

0xmovses (changes requested):
One question on building the zeroed `BridgeTransfer`. Also some conflicts here that need to be resolved.

andyjsbell:
Do we want to add `copy` to this type?  Probably a `tuple` would be a better design here.

protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
0xPrimata: if changing to error.
```
#[view]
    public fun bridge_transfers(bridge_transfer_id : vector<u8>) : BridgeTransfer acquires BridgeTransferStore, BridgeConfig {
        let config_address = borrow_global<BridgeConfig>(@atomic_bridge).bridge_module_deployer;
        let store = borrow_global<BridgeTransferStore>(config_address);
        if (!aptos_std::smart_table::contains(&store.transfers, bridge_transfer_id)){
            assert!(false, 1);
        } else {
            let bridge_transfer = aptos_std::smart_table::borrow(&store.transfers, bridge_transfer_id);
            *bridge_transfer
        }
    }
```
    This errors with an error that indicates that I should return a BridgeTransfer
```
    transfers: SmartTable<vector<u8>, BridgeTransfer>,
   │                                             -------------- Found: '(atomic_bridge=0xC3BB8488AB1A5815A9D543D7E41B0E0DF46A7396F89B22821F07A4362F75DDC5)::atomic_bridge_initiator::BridgeTransfer'. It is not compatible with the other type.
   ·  
87 │ ╭         if (!aptos_std::smart_table::contains(&store.transfers, bridge_transfer_id)){
88 │ │             assert!(false, 1);
   │ │                              - Found: '()'. It is not compatible with the other type.
89 │ │         } else {
90 │ │             let bridge_transfer = aptos_std::smart_table::borrow(&store.transfers, bridge_transfer_id);
91 │ │             *bridge_transfer
92 │ │         }
   │ ╰─────────^ Incompatible branches
   ```

0xPrimata:
@andyjsbell do you mean returning a tuple? Certainly, just wondering how to do that exactly. Any suggestions on how to remove copy are welcome.

protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
andygolay: I've updated it to abort (see this commit: https://github.com/movementlabsxyz/movement/pull/434/commits/d90348c008adf232096186850f5d2f99aa6ff6b0)

andygolay:
> @andyjsbell do you mean returning a tuple? Certainly, just wondering how to do that exactly. Any suggestions on how to remove copy are welcome.

I updated it to return a tuple (see this commit: https://github.com/movementlabsxyz/movement/pull/434/commits/d90348c008adf232096186850f5d2f99aa6ff6b0). This allows removing the copy and drop ability, which don't really seem like abilities we want for `BridgeTransfer`. 

(I changed the name of `bridge_transfers` to `get_bridge_transfer_from_id` to make it describe what the function does.

andygolay (approved):
LGTM, I made some commits to address review comments by @0xmovses and @andyjsbell and to make CI pass.

PR #308, opened 2024-08-06
Refactor background tasks and services out of execution units

mzabaluev:
# Summary
- **Categories**: `protocol-units`, `networks`.

Start on https://github.com/movementlabsxyz/movement/issues/63 by refactoring maptos-opt-executor, maptos-fin-view, and their integration in maptos-dof-execution and suzuka-full-node.

# Testing

Locally:
`cargo test -p maptos-opt-executor`
`cargo test -p maptos-fin-view`

Existing integration tests should pass in CI after the changes.

PR #439, opened 2024-08-23
Restore Optimizations and Configuration Removed in Background Task Refactor

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

1. Moves `BlobWriteRequest` submission back into an unawaited task, allowing the next blob batch to be constructed. 
2. Restores configuration for block retries.

# Changelog

# Testing
1. e2e testing. 

# Outstanding issues
1. I'm not 100% I caught everything. But, I think this is all of consequence that got dropped.


musitdev (approved):

mzabaluev (approved):
Sorry I have clobbered these changes while updating #308.

PR #442, opened 2024-08-26
fix(maptos-opt-executor): lower load shedding limit

mzabaluev:
# Summary

Lower the load shedding limit in the transaction ingress pipe task
to 12k.

# Changelog

* Maximum number of transactions in flight set to 12000.


l-monninger:
@mzabaluev I would prefer we move this into the config. It makes it so the SRE team can specify based on what they see.

l-monninger (changes requested):
Move to config unless you have a good reason not to do so.

mzabaluev:
> I would prefer we move this into the config. It makes it so the SRE team can specify based on what they see.

@l-monninger Which section of the execution config should this belong to? `chain::Config` does not feel right.

l-monninger (approved):

Issue #314, opened 2024-08-08
Update bridge shared`Amount` type to handle both WETH and ETH values


0xmovses:
## Summary
The `AtomicBridgeInitiator.sol` has the ability to handle both WETH and ETH deposits.  

The bridge has a shared type `Amount`:
```rust
#[derive(Deref, DerefMut, Debug, Clone, Copy, PartialEq, Eq)]
pub struct Amount(pub u64);
```
We need to update it so that its inner type is generic over `V`, where V stands for value
pub struct Amount<A>(pub V);

We want it to be generic, because on the movement side this will just be a u64, whereas on the eth side, we want an enumerable. 

Ethereum `V` enum. 
```rust
enum EthValue {
Weth(u64)
Eth(u64)
}
```
The client can then handle and process both types. Depending on what variant of the enum is used, the calls will have to be updated. If Eth the value is passed in using the `value` call (sends native eth).
```rust 
.initiateBridgeTransfer(
				U256::from(0), 
				FixedBytes(recipient_bytes),
				FixedBytes(hash_lock.0),
				U256::from(time_lock.0),
			)
			.value(U256::from(val));
```
If WETH
```rust
.initiateBridgeTransfer(
				U256::from(val), 
				FixedBytes(recipient_bytes),
				FixedBytes(hash_lock.0),
				U256::from(time_lock.0),
			)
```
Do a match on the enum, then construct the appropriate call. 

Also because there is a new generic type A, that will have to be set in the trait implementation:
```rust
#[async_trait::async_trait]
impl BridgeContractInitiator for EthClient {
	type Address = EthAddress;
	type Hash = EthHash;
	type Value = EthValue; // the enum 
```
Also for `BridgeContractCounterparty`.

## Requirements
The above updates should be made, and existing tests should pass. Updates will need to be made to the tests. 

Points: 2

0xmovses:
No we shouldn't update the params of the traits. You can add a third variant `WethAndEth((u64, u64))` then for each variant construct the appropriate alloy call. 

You'll need to construct a different alloy call for each case, I think using a match to deal with each case would be the most idiomatic. But happy for you to do it some other way if you prefer, we shouldn't update the trait fn sigs.

0xmovses:
Good point raised though, for the third case.

0xmovses:
In the `WethAndEth((u64, u64))` you will have to de-structure the `val`,
```rust
match EthValue {
   _= > other cases
[...]

PR #215, opened 2024-07-17
ENG 546 atomic bridge

0xmovses:
# Summary
Atomic Bridge Feature Branch :

See the full [RFC40](https://github.com/movementlabsxyz/rfcs/tree/main/0040-atomic-bridge)

Tasks to be completed before review and merge
- [x] https://github.com/movementlabsxyz/movement/issues/250
- [x] https://github.com/movementlabsxyz/movement/issues/213
- [x] https://github.com/movementlabsxyz/movement/pull/225
- [x] https://github.com/movementlabsxyz/movement/issues/221#issue-2413657061
- [x] https://github.com/movementlabsxyz/movement/issues/222
- [x] https://github.com/movementlabsxyz/movement/issues/224#issue-2413712253
- [x] https://github.com/movementlabsxyz/movement/issues/223

# Changelog
- Adds service crates `/protocol-units/chains/*`
- Adds Abstract bridge logic
- Adds CI checks for tests
- Adds solidity contracts with unit tests (both directions)
- Adds move modules with unit tests (both directions)
- Adds `EthClient` integration tests

# Testing
- To test the fully mocked implementation of the shared bridge lib: `cargo test -p bridge-shared`
-  Integration tests (ongoing)`cargo test --test eth_movement -- --nocapture --test-threads=1 `
- All testing code has been added to CI. 
- All the code in this feature branch has been merged via PRs and reviews. 

# Outstanding issues
NB the shared bridge logic unit test is breaking due to some updated implementation. This is is low-pri and will shortly be fixed. It should not block the merge as integration tests pass. 

Further work is yet to be implemented. But we want to be merging directly into main and merge in this fairly long standing feature branch in now. 

The work yet to be completed : 
- https://github.com/movementlabsxyz/movement/issues/331
- https://github.com/movementlabsxyz/movement/issues/315
- https://github.com/movementlabsxyz/movement/issues/314
- https://github.com/movementlabsxyz/movement/issues/313
- https://github.com/movementlabsxyz/movement/issues/312
- https://github.com/movementlabsxyz/movement/issues/311
- https://github.com/movementlabsxyz/movement/issues/222
- https://github.com/movementlabsxyz/movement/issues/318

PR #455, opened 2024-08-27
weth and eth handling on ethereum

0xPrimata:
# Summary
- **RFCs**: [315](https://github.com/movementlabsxyz/movement/issues/314)

Adds Enum handling of different assets

# Changelog

- Adds AssetType enum which describes the asset being bridged, future proofing addition of new assets
- Adds methods to handle variants EthAndWeth and Moveth.
- adds eth client and test harness methods to handle WETH
- adds tests to handle both eth and weth

# Testing

Passing tests

# Outstanding issues

Attention to Moveth


protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -223,18 +250,44 @@ impl BridgeContractInitiator for EthClient {
 		time_lock: TimeLock,
 		amount: Amount, // the ETH amount
 	) -> BridgeContractInitiatorResult<()> {
+		if amount.weth() > 0 {
+			let deposit_weth_signer = self.get_signer_address();
+			println!("deposit_weth_signer: {:?}", deposit_weth_signer);
```
0xmovses, line 255: let's remove all occurrences of `println!`

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -223,18 +250,44 @@ impl BridgeContractInitiator for EthClient {
...
+			let approve_call: alloy::contract::CallBuilder<_, &_, _> = contract
+				.approve(self.initiator_contract_address()?, U256::from(amount.weth()));
+			let WETH9::balanceOfReturn { _0: balance } = contract
+				.balanceOf(deposit_weth_signer)
+				.call()
+				.await
+				.expect("Failed to get balance");
+			let signer = self.get_signer_address();
+			println!("caller: {}", signer);
+			println!("balance: {}", balance);
```
0xmovses, line 271: Remove these `println!`

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -223,18 +250,44 @@ impl BridgeContractInitiator for EthClient {
...
+
+			send_transaction(approve_call, &send_transaction_rules(), RETRIES, GAS_LIMIT)
+				.await
+				.expect("Failed to approve");
+		}
```
0xmovses, line 276: All this logic inside `if amount.weth() > 0 ` I would put inside the integration test. It's not the responsibility of this function to do WETH calls, its just for calling the bridge contracts. 

In real world if a user hasn't called approve on a WETH contract we shouldn't be handling that for them off-chain in a hidden way. 

Put all this logic inside the `eth_movement.rs` integration test, perhaps in a utility function so that you can reuse it. It could even be a method on the `TestHarness`, as its technically a testing utility method.

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -223,18 +250,44 @@ impl BridgeContractInitiator for EthClient {
...
 		let call = contract
 			.initiateBridgeTransfer(
-				U256::from(0), // For now a 0 WETH amount
+				U256::from(amount.weth()),
 				FixedBytes(recipient_bytes),
 				FixedBytes(hash_lock.0),
 				U256::from(time_lock.0),
 			)
-			.value(U256::from(amount.0));
+			.value(U256::from(amount.eth())).from(_initiator_address.0.0);
```
0xmovses, line 288: We should remove the underscore from `_initiator_address` as its now being used.

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -312,7 +365,7 @@ impl BridgeContractInitiator for EthClient {
 			hash_lock: HashLock(eth_details.hash_lock),
 			//@TODO unit test these wrapping to check for any nasty side effects.
 			time_lock: TimeLock(eth_details.time_lock.wrapping_to::<u64>()),
-			amount: Amount(eth_details.amount.wrapping_to::<u64>()),
+			amount: Amount(AssetType::EthAndWeth((0,eth_details.amount.wrapping_to::<u64>()))),
```
0xmovses, line 368: Wouldn't the asset always be WETH on `BridgeTransferDetails`? As its what's stored on the contract? No Eth is every stored in the contract.

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -342,7 +395,7 @@ impl BridgeContractCounterparty for EthClient {
 			FixedBytes(hash_lock.0),
 			U256::from(time_lock.0),
 			recipient.0 .0,
-			U256::from(amount.0),
+			U256::try_from(amount.0).unwrap(),
```
0xmovses, line 398: We should not use `unwrap()` in production code. Instead you want to return a `BridgeContractInitiatorResult`

```
U256::try_from(amount.0).map_err(|e| {
        // Handle the error appropriately, possibly converting it to your custom error type
        BridgeContractInitiatorError::ConversionError
    })
```

And you would need to add `ConversionError` variant to `BridgeContractInitiatorError` Because:
`pub type BridgeContractInitiatorResult<T> = Result<T, BridgeContractInitiatorError>;`

https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -402,7 +403,7 @@ impl BridgeContractCounterparty for MovementClient {
 			to_bcs_bytes(&hash_lock.0).unwrap(),
 			to_bcs_bytes(&time_lock.0).unwrap(),
 			to_bcs_bytes(&recipient.0).unwrap(),
-			to_bcs_bytes(&amount.0).unwrap(),
+			to_bcs_bytes(&amount.weth()).unwrap(),
```
0xmovses, line 406: The `Amount` type on Movement shouldn't be `weth()` but `Moveth`, this is why originally I was wanting Amount to have a generic param, so that on Movement side it could just be `u64`, we'll have to solve this just with the enum approach now. 

Probably a method `.moveth()`, even if the inner logic of the method is the same, the semantics are wrong.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -224,12 +225,89 @@ async fn test_eth_client_should_successfully_call_initiate_transfer() {
...
+		.initiate_bridge_transfer(
+			InitiatorAddress(EthAddress(signer_address)),
+			RecipientAddress(recipient),
+			HashLock(hash_lock),
+			TimeLock(100),
+			Amount(AssetType::EthAndWeth((1, 1))),
+		)
+		.await
+		.expect("Failed to initiate bridge transfer");
+}
```
0xmovses, line 282: We're missing the case `test_eth_client_should_successfully_call_initiate_transfer_only_eth`, let's add it.

protocol-units/bridge/shared/src/types.rs
```diff
@@ -135,13 +138,76 @@ impl From<Uint<256, 4>> for TimeLock {
 }
 
 #[derive(Deref, DerefMut, Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Amount(pub u64);
+pub struct Amount(pub AssetType);
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
```
0xmovses, line 142: Let's add doc comments for AssetType and its fields
```suggestion
/// The type of Asset being used
#[derive(Clone, Debug, PartialEq, Eq, Copy)]
```

protocol-units/bridge/shared/src/types.rs
```diff
@@ -135,13 +138,76 @@ impl From<Uint<256, 4>> for TimeLock {
...
+pub enum AssetType {
+	EthAndWeth((u64, u64)),
```
0xmovses, line 144:
```suggestion
       /// Where the first tuple value is `Eth` and the second tuple value is `Weth`
	EthAndWeth((u64, u64)),
```

0xmovses (changes requested):
Thanks for this! Few changes needed.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
0xPrimata: it's above, it's not named _only_eth

protocol-units/bridge/chains/ethereum/src/client.rs
0xPrimata: yes, that's why on BridgeTransferDetails we only use the second param for EthAndWeth

protocol-units/bridge/chains/movement/src/lib.rs
0xPrimata: Think that accessing .0 here would be fine, correct?

0xmovses: I mean, yes it would work, its just the semantics / method name I'm referring to. It would be confusing for the reader as they would think its actually weth being used here.

andygolay:
Mind filling out the PR description? For anything other than very small PRs we should be using the template provided.

0xmovses:
Yep lets add some documentation into the PR description, then this is good to merge 🚀

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -342,7 +401,9 @@ impl BridgeContractCounterparty for EthClient {
 			FixedBytes(hash_lock.0),
 			U256::from(time_lock.0),
 			recipient.0 .0,
-			U256::from(amount.0),
+			U256::try_from(amount.0).map_err(|e| {   
+				BridgeContractInitiatorError::ConversionError  
+			}).expect("Failed to convert amount to U256"),
```
0xmovses, line 406: If the `expect` is executed, the service will panic causing a crash. https://doc.rust-lang.org/std/result/enum.Result.html#method.expect

Some of these expect calls were left in when initially writing the service, let's remove them now. Instead you could to this, where you have implemented the trait `TryFrom` set`type Error = BridgeContractCounterpartyError`

Because `lock_bridge_transfer_assets` returns `BridgeContractCounterpartyResult<()>` and this is 
`pub type BridgeContractCounterpartyResult<T> = Result<T, BridgeContractCounterpartyError>;`

You should then be able to do `U256::try_from(amount.0)?`

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -402,7 +403,7 @@ impl BridgeContractCounterparty for MovementClient {
 			to_bcs_bytes(&hash_lock.0).unwrap(),
 			to_bcs_bytes(&time_lock.0).unwrap(),
 			to_bcs_bytes(&recipient.0).unwrap(),
-			to_bcs_bytes(&amount.0).unwrap(),
+			to_bcs_bytes(&amount.moveth()).unwrap(),
```
0xmovses, line 406: 👍

protocol-units/bridge/shared/src/types.rs
```diff
@@ -135,13 +138,84 @@ impl From<Uint<256, 4>> for TimeLock {
...
+
+impl AddAssign for AssetType {
+	fn add_assign(&mut self, other: Self) {
+		match (self, other) {
+			(AssetType::Moveth(ref mut a), AssetType::Moveth(b)) => *a += b,
+			(AssetType::EthAndWeth((ref mut a, ref mut b)), AssetType::EthAndWeth((c, d))) => {
+				*a += c;
+				*b += d;
+			}
+			_ => panic!("Cannot add different AssetTypes"),
```
0xmovses, line 182: Have this return an Result type instead of panicing as this would cause the program to crash. The Result type should be the same of the calling context then when you call `add_assign` you can append it with `?` to bubble up the error.

protocol-units/bridge/shared/src/types.rs
```diff
@@ -135,13 +138,84 @@ impl From<Uint<256, 4>> for TimeLock {
 }
 
 #[derive(Deref, DerefMut, Debug, Clone, Copy, PartialEq, Eq)]
-pub struct Amount(pub u64);
+pub struct Amount(pub AssetType);
+/// The type of Asset being used
+#[derive(Clone, Debug, PartialEq, Eq, Copy)]
+pub enum AssetType {
+	/// Where the first tuple value is `Eth` and the second tuple value is `Weth`  
+	EthAndWeth((u64, u64)),
```
0xmovses, line 146: I know we've gone over this a few times, but I still don't understand why we can't have:
```rust
pub enum AssetType {
   Eth(u64)
   Weth(u64)
   EthAndWeth((u64, u64))
   Moveth(u64)
}
```

Otherwise we have code like 
`Amount(AssetType::EthAndWeth((1000,0))),` 

This reads weird because its not Eth and Weth, it's just Eth. Do you see what I mean?

0xmovses (changes requested):
Looking good! Some room for a bit more improvement.

protocol-units/bridge/shared/src/types.rs
0xPrimata: Yep, but then if we are working with the bridge service we have to handle unexpected types. Notice that we cannot use more than one type. Also, this streamlines eth and weth handling for any call to the eth side. What if we expect eth only but get weth too? That weth is lost in value. I just feel like and eth and weth variants is over engineering. We can call ethandweth something else though

0xmovses: In rust any unexpected type at runtime would have to be `dyn` which its not here. These are all concrete types known at compile time. 

Any new `AssetType` could just be added as an enum variant, for eg `Btc`, `Sol`, `Usdc` etc. We can add as many as we want. So it becomes code as documentation. This would setup a nice pattern for configuration and extension. 

In your example, if we expect only eth but get weth too that would match on the `EthAndWeth` arm, and we handle that case.  I don't see the problem and see it only as a worthwhile improvement that improves readability, extensibility (more asset types) and semantics.   

Anything coming "out" from ETH initiator is always WETH.

0xPrimata: But then if we are handling BTC, SOL, USDC, etc, they would be new enum types, for sure, but they are not being handled as BTC and WBTC, they'll be just a single type. If we are handling BTC and WBTC from a chain that supports native and wrapped btc (Rootstock for example), We add a variant BtcAndWbtc((u64,u64)). This way it's easy to be explicit that this implementation, on this side of the bridge, handled two assets for the sake of compatibility with the chain.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -458,12 +459,12 @@ impl BridgeContractCounterparty for MovementClient {
 	) -> BridgeContractCounterpartyResult<()> {
 
 		let args = vec![
-			utils::serialize_vec(&initiator.0)?,
-			utils::serialize_vec(&bridge_transfer_id.0[..])?,
-			utils::serialize_vec(&hash_lock.0[..])?,
-			utils::serialize_u64(&time_lock.0)?,
-			utils::serialize_vec(&recipient.0.0.to_vec())?,
-			utils::serialize_u64(&amount.0)?,
+			to_bcs_bytes(&initiator.0).unwrap(),
```
0xmovses, line 462: I think you resolved the conflict incorrectly, use the new utils:: methods created that bubble up the error with ? Don't unwrap. 

These were added by @andygolay in https://github.com/movementlabsxyz/movement/pull/215/commits/3cde18ff89de6330d620b3a0de7676632ea1caa4

0xmovses (changes requested):
Wrong change accepted on the merge conflict. To avoid this all three of us @0xPrimata @andygolay and me should all be reviewing each others PR. 

Unless the PR is time-sensitive let's create a soft-rule that we want to approvals form other bridge Team members before approval. This will help create more awareness about eachothers changes and why they were made etc.

protocol-units/bridge/chains/movement/src/lib.rs
0xPrimata: conflict was around line 406, not this one, let me take a look

0xmovses (approved):
:shipit:

PR #472, opened 2024-08-29
refactor: match explicit variants for AssetType

0xmovses:
# Summary
Any new asset types would be added, BTC, USDC, SOL etc:

# Changelog
- Add two further variants to the `AssetType` enum, construct necessary calls for each arm. 

# Testing
- CI must pass


0xmovses:
For demonstration purposes only

Issue #465, opened 2024-08-29
Unit test calling `wrapping_to::<u64>` on `Uint<256, 4>`


0xmovses:
## Summary 

`ruint::from::Uint`provides the method `wrapping to`

```rust
impl<const BITS: usize, const LIMBS: usize> Uint<BITS, LIMBS>
pub fn wrapping_to<T>(&self) -> T
where
    Self: UintTryTo<T>,
# Examples
assert_eq!(300_U12.wrapping_to::<i8>(), 44_i8);
assert_eq!(255_U32.wrapping_to::<i8>(), -1_i8);
assert_eq!(0x1337cafec0d3_U256.wrapping_to::<U32>(), 0xcafec0d3_U32);
```

In our code we are calling it on two values `eth_details.timelock` and `eth_details.amount`:

https://github.com/movementlabsxyz/movement/blob/3cde18ff89de6330d620b3a0de7676632ea1caa4/protocol-units/bridge/chains/ethereum/src/client.rs#L314

## Requirements

Write unit tests inside `protocol-units/bridge/chains/ethereum/src/client.rs` to test that calling `wrapping_to::<u64>()` on  type `Uint<256, 4>` returns the expected results.

PR #479, opened 2024-08-29
Primata/465 unit test wrapping to amount and timelock

0xPrimata:
# Summary
- **RFCs**: [Issue 465](https://github.com/movementlabsxyz/movement/issues/465).
- **Categories**:`protocol-units`.

Adds a test that ingests eth_details and checks if wrapped to value is as expected.

# Changelog

- adds test_wrapping_to_on_eth_details
- adds fuzz_test_wrapping_to_on_eth_details

# Testing

singular and fuzz

# Outstanding issues
deeper integration with the stack maybe

### DIFF
--- a/protocol-units/bridge/chains/ethereum/src/client.rs
+++ b/protocol-units/bridge/chains/ethereum/src/client.rs
@@ -474,6 +474,7 @@ impl BridgeContractCounterparty for EthClient {
 	}
 }
 
+#[cfg(feature = "wrapping_to_test")]
 fn test_wrapping_to(a: &U256, b: u64) {
     assert_eq!(a.wrapping_to::<u64>(), b);
 }
--- a/protocol-units/bridge/chains/ethereum/src/client.rs
+++ b/protocol-units/bridge/chains/ethereum/src/client.rs
@@ -22,6 +22,7 @@ use bridge_shared::types::{
 use serde_with::serde_as;
 use std::fmt::{self, Debug};
 use url::Url;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::types::{
 	AlloyProvider, AtomicBridgeCounterparty, AtomicBridgeInitiator, CounterpartyContract, WETH9Contract, WETH9,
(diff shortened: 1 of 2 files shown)


0xmovses:
Can you drop the command in the description to run the unit test? Just good documentation. We don't need to add this unit test into our CI.

andygolay (commented):
Looks good. One question: Are the tests in `client.rs` in the CLI workflow?

I don't see them in https://github.com/movementlabsxyz/movement/blob/eng-546-atomic-bridge/.github/workflows/checks-all.yml

If they aren't in CI, perhaps they should be added in as part of this PR?

protocol-units/bridge/chains/ethereum/src/client.rs
```diff
@@ -472,3 +473,46 @@ impl BridgeContractCounterparty for EthClient {
 		}))
 	}
 }
+
+fn test_wrapping_to(a: &U256, b: u64) {
+    assert_eq!(a.wrapping_to::<u64>(), b);
+}
```
0xmovses, line 479: We don't need this as a util fn outside of `mod tests {}` you can pop it inside the test module.

0xmovses (changes requested):

protocol-units/bridge/chains/ethereum/src/client.rs
0xPrimata: Does it break the code if it's outside? why not expose it so that you could use it when building some tests by yourself?

0xmovses: We can do that but we would want to feature flag it. Any code for testing shouldn't be built in the release binary, because it adds bloat and could open up security attack vector.

0xmovses: But no it doesn't break anything.

andygolay:
@0xPrimata could you add the test command to run to test this, into the PR description.

0xmovses (approved):

Issue #380, opened 2024-08-19
Feature flag testing bridge code 


0xmovses:
## Summary 

We have several methods in the bridge crates that are intended for testing only. For example in `MovementClient::new()` this actually genning a random address and this should be renamed to `MovementClient::new_for_test`

This should be feature flagged as such 
```rust
#[cfg(test)]
pub fn new_for_test() {}
```
## Requirements
- Feature flag the code, rename the method and improve the semantics everywhere where we have code intended for testing purposes only inside `bridge/chains/ethereum/*` and `/bridge/chains/movement/*`
- All checks should pass

0xPrimata:
Here you mentioned changing new to new_for_test but new_for_test already exists. Not sure how to handle that case

andygolay:
There's already a function `MovementClient::new_for_test`. I added it to provide a way to spin up a local Movement network, which is not needed on the Eth side because Eth has Anvil.

`MovementClient::new()` might not be fully fleshed out, for example there are dummy values in the current implementation. But there needs to be a production, non-test function for generating a new `MovementClient`, right? Isn't that the purpose of `MovementClient::new`? 

`EthClient::new` is called in production (non-test) in [the CLI](https://github.com/movementlabsxyz/movement/blob/eng-546-atomic-bridge/protocol-units/bridge/cli/src/eth_to_moveth.rs)

From what I can tell, the use for `MovementClient::new` hasn't been written into the CLI yet. It will need to be, and at that point we'll need `MovementClient::new` for production use. 

It appears to me that the way @0xPrimata is addressing this issue, by annotating `MovementClient::new_for_test` and other test functions as shown [here](https://github.com/movementlabsxyz/movement/pull/495/files) is correct.

PR #495, opened 2024-08-30
feat: feature flag tests

0xPrimata:
# Summary
- **RFCs**: [issue 380](https://github.com/movementlabsxyz/movement/issues/380)

Adding feature flags across tests

# Changelog

# Testing

# Outstanding issues


0xPrimata:
Cargo check is running fine locally.

0xmovses (changes requested):
Also anything relating to weth calls is just for testing. The Relayer wont make calls to Weth contract on behalf of the user. All of those should be feature flagged. For eg `deposit_weth_and_approve`
`weth_contract_address`
`weth_contract`

The field itself too 

```rust
#[allow(dead_code)]
#[derive(Clone)]
pub struct EthClient {
	rpc_provider: AlloyProvider,
	rpc_port: u16,
	ws_provider: Option<RootProvider<PubSubFrontend>>,
	initiator_contract: Option<InitiatorContract>,
	counterparty_contract: Option<CounterpartyContract>,
	#[cfg(test)]
	weth_contract: Option<WETH9Contract>,
	config: Config,
}

There may be a few other things, but this reduce size of the release bin.

File: protocol-units/bridge/integration-tests/src/lib.rs [hunk 1]
```diff
@@ -31,11 +31,13 @@ use ethereum_bridge::{
 use movement_bridge::{Config as MovementConfig};
 use std::sync::{Arc, RwLock};
 
+#[cfg(test)]
```
0xmovses, line 34: You don't need to feature flag anything inside the `/integration-tests` lib as this who binary will be conditionally built. 

I think we have to use a custom feature we could call it bridge-tests
in bridge-integration-tests Cargo.toml you do
```
[package]
name = "bridge-integration-tests"
version = "0.1.0"
[features]
bridge-tests = []
```
then in the root workspace Cargo.toml  you do
```
[features]
bridge-tests = ["bridge-integration-tests"]
```
and
```
[dev-dependencies]
bridge-integration-tests = { path = "./path/to/package", features = ["bridge-tests"] }
```

Something like this, It's a way to create your own testing binaries that get conditionally built. Without having to conform to the rather limiting workspace `/tests/` rule.

0xmovses:
And to add, you only want to feature flag any production code that is used for testing. I.e anything not inside the `integration-tests` binary, as the `integration-tests` bin will be conditionally built using the pattern outlined above.

0xPrimata:
instead of taking that approach, I had `tests/eth_movement` feature flagged as test.
I wanted to also flag other dependencies, like TestHarness under `src/lib.rs` but it breaks imports in tests.
Same thing happens with flagging weth stack, but the biggest issue is that it breaks functions being used elsewhere like `deploy_weth_contract` under TestHarness.

PR #502, opened 2024-09-02
Ordering Fairness

### LINKED ISSUE #500: Impose order on transactions going into a block and filter out duplicates
To help mitigate DoS issues with transaction replay discussed in #416, re-impose a deterministic ordering on transactions that will form a block and filter out identical transactions. Changes to make to this end:

1. Make all fields of the `Transaction` type in movement-types private to preserve the invariant: the `id` must be computed as the hash of the other members. Provide the public accessor methods named the same as the fields. The values of this type are then immutable.
2. Implement `Ord` (and `PartialOrd`) on `Transaction` so that the `sequence_number` fields are compared first. If not equal, their order decides the ordering of the transactions. Otherwise, it is decided by the lexicographical ordering between the `id` fields.
3. Change the collection of transactions inside the `Block` type to `BTreeSet`. When a block is formed, transactions will be ordered by construction and duplicate transactions eliminated.

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**:  `protocol-units`

Resolves #500

# Changelog

# Testing
1. `cargo test -p memseq`
2. `cargo test -p move-rocks`

# Outstanding issues
1. Add tests back to CI.


util/movement-types/src/lib.rs
```diff
@@ -1,11 +1,21 @@
...
 use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+use std::collections::BTreeSet;
 
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct Id(pub [u8; 32]);
+pub struct Id([u8; 32]);
 
 impl Id {
+	pub fn new(data: [u8; 32]) -> Self {
```
mzabaluev, line 11: If `Id` is only supposed to be created in movement-types, this could be further restricted to `pub(crate)`.

util/movement-types/src/lib.rs
```diff
@@ -1,11 +1,21 @@
...
+		Self(data)
+	}
+
+	pub fn inner(&self) -> &[u8; 32] {
```
mzabaluev, line 15: `as_bytes`?

util/movement-types/src/lib.rs
```diff
@@ -50,25 +60,52 @@ impl Transaction {
 		Self { data, sequence_number, id }
 	}
 
-	pub fn id(&self) -> Id {
-		self.id.clone()
+	pub fn id(&self) -> &Id {
+		&self.id
+	}
+
+	pub fn data(&self) -> &Vec<u8> {
```
mzabaluev, line 67:
```suggestion
	pub fn data(&self) -> &[u8] {
```

There is never a need to return a reference to an immutable `Vec`.

util/movement-types/src/lib.rs
```diff
@@ -50,25 +60,52 @@ impl Transaction {
...
+impl Ord for Transaction {
+	fn cmp(&self, other: &Self) -> Ordering {
+		// First, compare by sequence_number
+		match self.sequence_number().cmp(&other.sequence_number()) {
+			Ordering::Equal => {}
+			non_equal => return non_equal,
+		}
+
+		// If sequence number is equal, then compare by transaction on the whole
+		self.id().cmp(other.id())
```
mzabaluev, line 89: Looks good.

util/movement-types/src/lib.rs
```diff
@@ -118,27 +159,55 @@ impl Block {
...
+
+	pub fn id(&self) -> &Id {
+		&self.id
+	}
+
+	pub fn parent(&self) -> &Vec<u8> {
+		&self.parent
+	}
+
+	pub fn transactions(&self) -> &BTreeSet<Transaction> {
```
mzabaluev, line 174: This exposes the internal collection as part of the API.
Would it be better to return an iterator?

If you move the `Block` definitions in `mod block` so that the iterator can be named `block::Transactions`:

```rust
use crate::Transaction;
use std::collections::btree_set;

pub type Transactions<'a> = btree_set::Iter<'a, Transaction>;

impl Block {
    pub fn transactions(&self) -> Transactions {
        self.transactions.iter()
    }
}
```

util/movement-types/src/lib.rs
```diff
@@ -118,27 +159,55 @@ impl Block {
...
 impl Commitment {
+	pub fn new(data: [u8; 32]) -> Self {
+		Self(data)
+	}
+
 	pub fn test() -> Self {
 		Self([0; 32])
 	}
 
+	pub fn inner(&self) -> &[u8; 32] {
```
mzabaluev, line 207: Also here, I think `as_bytes` is more descriptive.

util/movement-types/src/lib.rs
```diff
@@ -1,11 +1,21 @@
 use aptos_types::state_proof::StateProof;
 use core::fmt;
 use serde::{Deserialize, Serialize};
+use std::cmp::Ordering;
+use std::collections::BTreeSet;
 
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct Id(pub [u8; 32]);
```
mzabaluev, line 6: I think it's good to derive `Copy` as well. We can commit to `Id` being of fixed size, so we don't need to gum the code with clone calls.

util/movement-types/src/lib.rs
```diff
@@ -50,25 +60,52 @@ impl Transaction {
 		Self { data, sequence_number, id }
 	}
 
-	pub fn id(&self) -> Id {
-		self.id.clone()
+	pub fn id(&self) -> &Id {
```
mzabaluev, line 63: This can return `Id` by value, to avoid a lot of inconvenience at the call sites. The value is supposed to be small and `Copy` as suggested in another comment.

To benefit from optimizations that should largely make these distinctions irrelevant, put an `#[inline]` attribute on the method. This will make the implementation visible outside of the crate, the compiler won't do it for non-generic types otherwise and I'm not sure LTO can make up for this in all cases.

util/movement-types/src/lib.rs
```diff
@@ -118,27 +159,55 @@ impl Block {
 		Self { metadata, parent, transactions, id }
 	}
 
-	pub fn id(&self) -> Id {
-		self.id.clone()
+	pub fn into_parts(self) -> (BlockMetadata, Vec<u8>, BTreeSet<Transaction>, Id) {
+		(self.metadata, self.parent, self.transactions, self.id)
+	}
+
+	pub fn id(&self) -> &Id {
```
mzabaluev, line 166: Return `Id` by value.

util/movement-types/src/lib.rs
```diff
@@ -118,27 +159,55 @@ impl Block {
...
+		&self.id
+	}
+
+	pub fn parent(&self) -> &Vec<u8> {
```
mzabaluev, line 170: This is questionable in multiple ways now.
At the surface level, this should return `&[u8]`.
But how does a byte array represent a parent? Is it [supposed to be](https://github.com/movementlabsxyz/movement/blob/31386f74a2700920fec093775c221acdc9c6c7a2/protocol-units/sequencing/memseq/sequencer/src/lib.rs#L100) an `Id` as well?
I don't see this member used much.

util/movement-types/src/lib.rs
mzabaluev: To get it even more encapsulated, have the iterator as a newtype wrapping the inner `btree_set::Iter` and implement `Iterator`, `DoubleEndedIterator`, `ExactSizeIterator`, and all the rest of the iterator marker traits on the newtype, but it's too much boilerplate for now, so maybe leave that for 1.0.

util/movement-types/src/lib.rs
```diff
@@ -118,27 +159,55 @@ impl Block {
...
 
 	pub fn add_transaction(&mut self, transaction: Transaction) {
-		self.transactions.push(transaction);
+		self.transactions.insert(transaction);
 	}
 }
 
 #[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct Commitment(pub [u8; 32]);
+pub struct Commitment([u8; 32]);
```
mzabaluev, line 196: This should implement `Copy` as well.

util/movement-types/src/lib.rs
```diff
@@ -184,9 +253,31 @@ impl fmt::Display for Commitment {
...
+impl BlockCommitment {
+	pub fn new(height: u64, block_id: Id, commitment: Commitment) -> Self {
+		Self { height, block_id, commitment }
+	}
+
+	pub fn height(&self) -> u64 {
+		self.height
+	}
+
+	pub fn block_id(&self) -> &Id {
```
mzabaluev, line 270: Return `Id` by value.

util/movement-types/src/lib.rs
```diff
@@ -184,9 +253,31 @@ impl fmt::Display for Commitment {
...
+		&self.block_id
+	}
+
+	pub fn commitment(&self) -> &Commitment {
```
mzabaluev, line 274: Return `Commitment` by value.

util/movement-types/src/lib.rs
mzabaluev: Also there should better be `block::Id`, `transaction::Id` etc., for type-level safety against accidental mixing of IDs for different entities.

l-monninger: We can include it here. There is a question of whether you might want to do higher order compositions, e.g., in the shared sequencer where blocks look like transactions and their Ids are converted. But, I agree this is better if done explicitly. 

Would you prefer a base `Id` that is then rewrapped in each module? Or, just the raw `u8` slice?

l-monninger: > If Id is only supposed to be created in movement-types, this could be further restricted to pub(crate).

It is used elsewhere. In theory, this is also building towards an SDK.

util/movement-types/src/lib.rs
l-monninger: Good with me.

util/movement-types/src/lib.rs
l-monninger: So, just the `pub type Transactions<'a> = btree_set::Iter<'a, Transaction>` is your recommendation?

util/movement-types/src/lib.rs
l-monninger: See questions above about whether we want a shared base `Id`. I'd wonder about reuse and extensions of the Id.

util/movement-types/src/lib.rs
l-monninger: Oh, really? I wasn't aware that the compiler needed so much help inlining owned accessors of this sort. Can you link me to something for my own knowledge?

util/movement-types/src/lib.rs
l-monninger: It hasn't been used yet. It's better as an `Id`. But, the outstanding question per share-sequencer discussions was whether this might need to be something more than just the hash of the previous block. That is, you might want to do something like...

```rust
pub struct Parent {
    chain: Id,
    block: Id
}
```

...for atomic transaction bundles where shared ordering amongst chains is temporary.

util/movement-types/src/lib.rs
mzabaluev: Yes, that would do for now. I would split up into `block`, `transaction`, etc. modules already though, so we can have path-distinguished naming for these subsidiary types, rather than long C-programmers-disease affected names at the crate level.

util/movement-types/src/lib.rs
mzabaluev: Should we have this as an opaque type for now? I'd prefer not to have anything if it's not used, but if it needs to be on-chain, perhaps a newtype wrapping a dummy ID would do for now.

util/movement-types/src/lib.rs
mzabaluev: It's more nuanced than that. You should return the borrowed form of data in accessors by reference, so for `Vec<T>` it is `&[T]`, for `String` you return `&str`, etc. The standard owned containers add no useful API to their `Deref` target when they are immutably borrowed.

util/movement-types/src/lib.rs
mzabaluev: It's more about what metadata is available in the compiled crates. `#[inline]` forces the compiler to provide the body for inlining across crates (at some cost to compile time), otherwise the non-generic method is just linked as a symbol. There may be also LTO tricks nowadays to make this unnecessary.

util/movement-types/src/lib.rs
l-monninger: Yeah, let's go with wrapping a dummy ID.

util/movement-types/src/lib.rs
l-monninger: Sounds good!

util/movement-types/src/lib.rs
l-monninger: Okay

util/movement-types/src/lib.rs
l-monninger: There is potentially when considering the shared sequencer applications. I supposed there is also already some when we're looking at the exact way the passthrough is used. For now though, I'll stick with case-by-case `into` methods as opposed to having a shared inner domain-specific type that those are routed through.

protocol-units/da/m1/light-node-client/src/test/e2e/raw/sequencer.rs
```diff
@@ -1,5 +1,5 @@
 use crate::*;
-use movement_types::Block;
+use movement_types::block::Block;
```
mzabaluev, line 2: The "main type" in the module, and any other unambiguously named types, could be re-exported to the top level.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -50,7 +53,7 @@ impl RocksdbMempool {
 
 	fn internal_get_mempool_transaction_key(
 		db: Arc<DB>,
-		transaction_id: &Id,
+		transaction_id: &transaction::Id,
```
mzabaluev, line 56: If `Id` is `Copy`, there's no reason to pass it by reference.

mzabaluev (commented):
Good!

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -86,7 +92,10 @@ impl RocksdbMempool {
 }
 
 impl MempoolTransactionOperations for RocksdbMempool {
-	async fn has_mempool_transaction(&self, transaction_id: Id) -> Result<bool, Error> {
+	async fn has_mempool_transaction(
+		&self,
+		transaction_id: transaction::Id,
```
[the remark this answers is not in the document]
l-monninger: Yeah, I was noticing that as I was going through.

util/movement-types/src/block.rs
```diff
@@ -0,0 +1,213 @@
...
+
+impl TryFrom<Vec<u8>> for Commitment {
+	type Error = std::array::TryFromSliceError;
+
+	fn try_from(data: Vec<u8>) -> Result<Self, Self::Error> {
+		Ok(Self(data[..32].try_into()?))
+	}
+}
+
+impl From<[u8; 32]> for Commitment {
```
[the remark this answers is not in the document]
l-monninger: It looks like these aren't being used anywhere, so I've removed them.

mzabaluev (approved):
Looks good!

l-monninger:
@mzabaluev let me just trigger the labels again, then we can go ahead.

PR #503, opened 2024-09-03
Stronger assertions for counterparty integration tests

andygolay:
# Summary

- Bridge transfer details are fetched from chain using a view function, and assertions are made about transfer details.
- `BridgeTransferDetails` struct is updated to include `state` and to have correct data types.

# Changelog

- Complete `get_bridge_transfer_details()` function with `view` request in `protocol-units/bridge/chains/movement/src/lib.rs`
- Add calls to `get_bridge_transfer_details()` after lock, complete, and abort, along with assertions about bridge transfer details in `protocol-units/bridge/integration-tests/eth_movement.rs` functions `test_movement_client_should_successfully_call_lock_and_complete` and `test_movement_client_should_successfully_call_lock_and_abort`
- Add `state` field to `BridgeTransferDetails` struct throughout codebase.

# Testing

To test lock and complete, run 

```
rust_backtrace=1 cargo test --test eth_movement test_movement_client_should_successfully_call_lock_and_complete -- --nocapture --test-threads=1
```

To test lock and abort, run

```
rust_backtrace=1 cargo test --test eth_movement test_movement_client_should_successfully_call_lock_and_abort -- --nocapture --test-threads=1
```

Bridge transfer details are fetched after each function in each test

# Outstanding issues

- I skipped assertions about time_lock as it seemed unnecessary and was a little complicated, since the time_lock value in the Move module depends on the timestamp when the bridge transfer is locked. So if I say the time_lock is 3600 in my Rust code, then the actual time_lock will be `timestamp::now_seconds() + 3600`. Because in the Move code, `abort_bridge_transfer` asserts that the time_lock has expired, it didn't seem necessary to make further assertions about time.

- Everywhere on the Eth side where I added `state` to `BridgeTransferDetails` should be checked carefully. I attempted to use correct values but extra sets of eyes are appreciated.


protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
```diff
@@ -107,6 +108,17 @@ module atomic_bridge::atomic_bridge_counterparty {
         )
     }
 
+    #[view]
+    public fun get_bridge_transfer_state(bridge_transfer_id: vector<u8>): u8 acquires BridgeTransferStore, BridgeConfig {
```
0xPrimata, line 112: why adding a specific get bridge transfer state function? why not just extract it from the tuple? Is this mostly just for testing or is bridge_transfers unable to return data?

andygolay: was for testing; i got bridge_transfers returning data now

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -75,91 +80,196 @@ async fn test_movement_client_should_publish_package() -> Result<(), anyhow::Err
...
-		faucet_client
-		.fund(movement_client.address(), 100_000_000)
-		.await?;
 
-		// Check the balance of movement_client after funding
-		let balance = coin_client
-		.get_account_balance(&movement_client.address())
-		.await?;
+#[tokio::test]
+async fn test_movement_client_should_successfully_call_lock_and_abort() -> Result<(), anyhow::Error> {
```
0xPrimata, line 180: This test currently fails with the provided command. From what I'm reading, could this be an issue to hard coded environment variables?
```
024-09-03T22:14:34.734579Z ERROR movement_bridge: Failed to get transfer item: Api(AptosErrorResponse { error: AptosError { message: "Table Item not found by Table handle(0xe513451190640deb971ca1c98037382f284fb240358541c91eb672d8d607206), Table key(\"0x30303030303030303030303030303030303030303030307472616e7366657231\") and Ledger version(63)", error_code: TableItemNotFound, vm_error_code: None }, state: Some(State { chain_id: 4, epoch: 2, version: 63, timestamp_usecs: 1725401673714174, oldest_ledger_version: 0, oldest_block_height: 0, block_height: 28, cursor: None }), status_code: 404 })
thread 'test_movement_client_should_successfully_call_lock_and_abort' panicked at protocol-units/bridge/integration-tests/tests/eth_movement.rs:212:14:
Failed to get bridge transfer state: ViewSerializationError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
FAILED

failures:

failures:
    test_movement_client_should_successfully_call_lock_and_abort

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 11 filtered out; finished in 33.38s
```

protocol-units/bridge/shared/src/types.rs
```diff
@@ -219,6 +224,13 @@ impl From<Uint<256, 4>> for Amount {
 	}
 }
 
+//#[derive(Debug, PartialEq, Eq, Clone)]
+//enum State {
```
0xPrimata, line 228: Currently the enum series for the Ethereum side starts with 0.
There is also a big difference between how initiator and counterparty are calling states:

```
enum MessageState {
        INITIALIZED, vs PENDING
        COMPLETED,
        REFUNDED
    }
```
   
   On Movement they differ on the first and third state:
   
```
const INITIALIZED: u8 = 1; vs const LOCKED: u8 = 3;
    const COMPLETED: u8 = 2;
    const REFUNDED: u8 = 3; vs const CANCELLED: u8 = 3;
```
    
because 0 signifies uninitialized states on Ethereum, the Ethereum contract should be defined as 1,2,3 as not initialized states respectively as it could open up for unexpected attack vectors. We also should standardize nomeclature as well. Seems like simply using refunded makes more sense.
    
    This way, this shared enum could be used again in the /shared component:
    ```
    enum State {
    UNINITIALIZED,
    INITIALIZED,
    COMPLETED,
    REFUNDED
    }
    ```

protocol-units/bridge/integration-tests/tests/eth_movement.rs
andygolay: I think you're using an old version of the branch, because I don't have "Failed to get bridge transfer state" anymore in the most up-to-date version. Mind doing a `git pull` and trying again?

protocol-units/bridge/shared/src/types.rs
andygolay: yeah this seems like a good solution. maybe good for a separate issue?

protocol-units/bridge/integration-tests/tests/eth_movement.rs
0xPrimata:
```
test test_movement_client_should_successfully_call_lock_and_abort ... STDERR: 
STDERR: 
STDERR: Readiness endpoint: http://127.0.0.1:8070/
STDERR: 
STDERR: Transaction stream is starting, please wait...
STDERR: Faucet is starting, please wait...
STDERR: Node API is starting, please wait...
STDERR: 
STDOUT: Completed generating configuration:
STDOUT:         Log file: "/Users/primata/movement/.movement/testnet/validator.log"
STDOUT:         Test dir: "/Users/primata/movement/.movement/testnet"
STDOUT:         Aptos root key path: "/Users/primata/movement/.movement/testnet/mint.key"
STDOUT:         Waypoint: 0:a83030a61d1156ba131fa924efbdfb7e312d3ae64145ad1820699754759d57bb
STDOUT:         ChainId: 4
STDOUT:         REST API endpoint: http://127.0.0.1:8080
STDOUT:         Metrics endpoint: http://127.0.0.1:9101/metrics
STDOUT:         Aptosnet fullnode network endpoint: /ip4/0.0.0.0/tcp/6181
STDOUT:         Indexer gRPC node stream endpoint: 127.0.0.1:50051
STDOUT: 
STDOUT: Aptos is running, press ctrl-c to exit
STDOUT: 
STDERR: 
STDERR: One of the services failed to start up, running shutdown steps...
STDERR: Ran shutdown steps
STDOUT: Unexpected error: 
STDOUT: One of the services crashed on startup:
STDOUT: Ok(
STDOUT:     Err(
STDOUT:         Error {
STDOUT:             context: "Service ended with an error",
STDOUT:             source: Os {
STDOUT:                 code: 48,
STDOUT:                 kind: AddrInUse,
STDOUT:                 message: "Address already in use",
STDOUT:             },
STDOUT:         },
STDOUT:     ),
STDOUT: )
STDOUT: Please check the logs in /Users/primata/movement/.movement/testnet
thread 'test_movement_client_should_successfully_call_lock_and_abort' panicked at /Users/primata/movement/protocol-units/bridge/chains/movement/src/lib.rs:173:33:
Failed to receive setup completion signal: RecvError(())
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
FAILED
```

andygolay: The "Address already in use" indicates that you already have a process running on the port. 

Run `ps aux | grep movement` to get the PID of the process. 

Then run `kill -9 <pid>`

Then try again and you shouldn't get the "Address already in use" error.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -458,13 +461,18 @@ impl BridgeContractCounterparty for MovementClient {
...
+		};
+
 		let args = vec![
-			utils::serialize_vec(&initiator.0)?,
-			utils::serialize_vec(&bridge_transfer_id.0[..])?,
-			utils::serialize_vec(&hash_lock.0[..])?,
-			utils::serialize_u64(&time_lock.0)?,
-			utils::serialize_vec(&recipient.0.0.to_vec())?,
-			utils::serialize_u64(&amount.moveth())?,
+			bcs::to_bytes(&initiator.0).map_err(|_| BridgeContractCounterpartyError::SerializationError)?,
```
0xmovses, line 470: I recall we created, or discussed using a util fn that wraps `bcs::to_bytes` and returns a `Result<T,  BridgeContractCounterpartyError::SerializationError>` reducing the repeating `.map_err` calls.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -503,15 +512,16 @@ impl BridgeContractCounterparty for MovementClient {
...
-		let _ = utils::send_and_confirm_aptos_transaction(
+		let result = utils::send_and_confirm_aptos_transaction(
 			&self.rest_client,
 			self.signer.as_ref(),
 			payload,
 		)
 		.await
 		.map_err(|_| BridgeContractCounterpartyError::CompleteTransferError);
+
+		println!("Complete bridge transfer result: {:?}", &result);
```
0xmovses, line 523: Either remove the `println!()` or use a `tracing::info`

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -520,33 +530,93 @@ impl BridgeContractCounterparty for MovementClient {
...
-		let _ = utils::send_and_confirm_aptos_transaction(
+		let result = utils::send_and_confirm_aptos_transaction(
 			&self.rest_client,
 			self.signer.as_ref(),
 			payload,
 		)
 		.await
 		.map_err(|_| BridgeContractCounterpartyError::AbortTransferError);
+
+		println!("Abort bridge transfer result: {:?}", &result);
```
0xmovses, line 550: Same with the `println!` here

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -520,33 +530,93 @@ impl BridgeContractCounterparty for MovementClient {
...
+		    .view(&view_request, None)
+		    .await
+		    .map_err(|_| BridgeContractCounterpartyError::CallError)?;
+	    
+		// Extract and parse the response
+		let values = response.inner();
+		if values.len() == 6 {
+		    let originator = values[0].as_str().ok_or(BridgeContractCounterpartyError::SerializationError)?;
+		    let recipient = values[1].as_str().ok_or(BridgeContractCounterpartyError::SerializationError)?;
+		    let amount = values[2].as_str().ok_or(BridgeContractCounterpartyError::SerializationError)?.parse::<u64>().map_err(|_| BridgeContractCounterpartyError::SerializationError)?;
```
0xmovses, line 587: Getting the value by index [0] can panic, I know you are handling the error and sure that these vals exist but its usually better to use `values.get(1)`.

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -520,33 +530,93 @@ impl BridgeContractCounterparty for MovementClient {
...
+		// Send the request to the "/view" endpoint using JSON
+		let response: Response<Vec<serde_json::Value>> = self.rest_client
+		    .view(&view_request, None)
+		    .await
+		    .map_err(|_| BridgeContractCounterpartyError::CallError)?;
+	    
+		// Extract and parse the response
+		let values = response.inner();
+		if values.len() == 6 {
+		    let originator = values[0].as_str().ok_or(BridgeContractCounterpartyError::SerializationError)?;
```
0xmovses, line 585: Let's reduce the `ok_or` calls, its verbose and repetitive. Instead you can create a small utility function. 
```rust
fn val_as_str(value: Option<&Value>) -> Result<&str, BridgeContractCounterpartyError> {
    value.as_ref().and_then(|v| v.as_str()).ok_or(BridgeContractCounterpartyError::SerializationError)
}
```
Then you can simply do:
```rust
let originator = val_as_str(values.get(0))?;
```

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -520,33 +530,93 @@ impl BridgeContractCounterparty for MovementClient {
...
+		    // Create the BridgeTransferDetails struct
+		    let details: BridgeTransferDetails<MovementAddress, [u8; 32]> = BridgeTransferDetails {
+			bridge_transfer_id,
+			initiator_address: InitiatorAddress(MovementAddress(originator_address)),
+			recipient_address: RecipientAddress(recipient_address_bytes),
+			amount: Amount(AssetType::Moveth(amount)),
+			hash_lock: HashLock(hash_lock_array),
+			time_lock: TimeLock(time_lock),
+			state,
+		    };
```
0xmovses, line 611: 👍

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -520,33 +530,93 @@ impl BridgeContractCounterparty for MovementClient {
...
+	    
+		    return Ok(Some(details));
+		}
+	    
+		Ok(None)
```
0xmovses, line 616: `if values.len() != 6` here I would return an Error here rather than the `Ok` variant of Result.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -19,13 +22,23 @@ use bridge_shared::{
...
 
+struct ChildGuard {
+	child: Child,
+    }
+    
+impl Drop for ChildGuard {
+	fn drop(&mut self) {
+	    let _ = self.child.kill();
+	}
+}
```
0xmovses, line 40: Nice.

protocol-units/bridge/shared/src/bridge_service/active_swap.rs
```diff
@@ -464,22 +464,23 @@ impl HasTimeoutError for LockBridgeTransferError {
...
+	details: BridgeTransferDetails<BFrom::Address, BFrom::Hash>
+	//BridgeTransferDetails {
+	////	bridge_transfer_id,
+	////	initiator_address,
+	////	recipient_address,
+	////	hash_lock,
+	////	time_lock,		
+	////	amount,
+	////	..
+	//}: BridgeTransferDetails<BFrom::Address, BFrom::Hash>,
```
0xmovses, line 476: Let's remove these comments.

protocol-units/bridge/shared/src/types.rs
0xmovses: Yes 0 would be `UNINITIALIZED`, on Eth, but by the time an event is emitted I don't think the state could ever possibly be 0 / unitialized.

0xmovses (changes requested):
Thanks for these changes, some suggestions for improvements.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
0xmovses: To confirm, the tests are working locally for me.

protocol-units/bridge/integration-tests/tests/eth_movement.rs
```diff
@@ -59,7 +64,7 @@ async fn test_movement_client_build_and_fund_accounts() -> Result<(), anyhow::Er
 #[tokio::test]
 async fn test_movement_client_should_publish_package() -> Result<(), anyhow::Error> {
 	let _ = tracing_subscriber::fmt()
-        .with_max_level(tracing::Level::ERROR)
+        .with_max_level(tracing::Level::DEBUG)
```
0xmovses, line 67: Let's not set the tracing level in the source code, as the output is quite noisy. Instead leave it up to the developer to set this env var if they want to debug.

0xmovses (changes requested):

protocol-units/bridge/chains/movement/src/lib.rs
andygolay: Yeah those utility functions were was causing a `CallError` but I just figured out why and fixed it. Good to go.

protocol-units/bridge/chains/ethereum/src/event_logging.rs
```diff
@@ -215,6 +215,10 @@ fn decode_log_data(
 					.as_uint()
 					.map(|(u, _)| u.into())
 					.ok_or_else(|| anyhow::anyhow!("Failed to decode TimeLock"))?;
+				let state = decoded.indexed[6]
```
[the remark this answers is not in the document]
andygolay: Do you mean fix it for all of these `indexed[#]` or just for `state`?

```
		let bridge_transfer_id = decoded.indexed[0]
					.as_fixed_bytes()
					.map(coerce_bytes)
					.ok_or_else(|| anyhow::anyhow!("Failed to decode BridgeTransferId"))?;
				let initiator_address = decoded.indexed[1]
					.as_address()
					.map(EthAddress)
					.ok_or_else(|| anyhow::anyhow!("Failed to decode InitiatorAddress"))?;
				let recipient_address = decoded.indexed[2]
					.as_fixed_bytes()
					.map(coerce_bytes)
					.ok_or_else(|| anyhow::anyhow!("Failed to decode RecipientAddress"))?;
				let amount = decoded.indexed[3]
					.as_uint()
					.map(|(u, _)| u.into())
					.ok_or_else(|| anyhow::anyhow!("Failed to decode Amount"))?;
				let hash_lock = decoded.indexed[4]
					.as_fixed_bytes()
					.map(coerce_bytes)
					.ok_or_else(|| anyhow::anyhow!("Failed to decode HashLock"))?;
				let time_lock = decoded.indexed[5]
					.as_uint()
					.map(|(u, _)| u.into())
					.ok_or_else(|| anyhow::anyhow!("Failed to decode TimeLock"))?;
				let state = decoded.indexed.get(6);

```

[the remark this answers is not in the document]
0xmovses: Hmm, its ok this is unrelated to your PR, it will come up in my event indexing work. Let's leave it.

0xmovses (approved):

PR #506, opened 2024-09-03
Integrate Indexer to bridge `TestHarness` 

0xmovses:
# Summary
Adds Indexer support to the `TestHarness`, it has just been discussed to go with the original approach of polling the rest endpoints on the full node rather than relying on the indexer to get events, this is because the indexer has no WS streaming capabilities either. 

However, we still may want to use the indexer with the bridge in the way discussed in our sync. I.e to use the indexer processor to filter events on the gRPC Transaction Stream. Therefor, I think these changes are still worthy to merge.

# Changelog
- Updates TestHarness to provide pre-defined testing args (reduce dup redefinitions)
- Adds Indexer process to TestHarness for bridge integration testing with the indexer 

# Testing
CI Should pass


andygolay:
@0xmovses @mzabaluev I created this PR on `aptos-core` to fix the issue causing the CI run to fail: 

https://github.com/movementlabsxyz/aptos-core/pull/65

0xmovses:
A different path was chosen

Issue #486, opened 2024-08-30
Add bridge Movement-side initiator integration tests


andygolay:
**Is your feature request related to a problem? Please describe.**

Tests are needed for Movement initiator to Eth counterparty swaps. This issue is specifically for the Movement-side initiator tests.

**Describe the solution you'd like**
- Create `movement_eth.rs` 
- Add the following tests:
  - `test_movement_client_call_initiate_bridge_transfer`
  - `test_movement_client_call_complete_bridge_transfer`
  - `test_movement_client_call_refund_bridge_transfer`
  - `test_movement_client_get_bridge_transfer_details`

**Describe alternatives you've considered**

**Additional context**
- Eth side tests will also be needed.

PR #523, opened 2024-09-04
Add bridge Movement-side initiator integration tests

andygolay:
# Summary
- Addresses all of #486 except the test of `refund_bridge_transfer` which I discovered issues in the Move module.

# Changelog

- Add `impl BridgeContractInitiator for MovementClient` in `protocol-units/bridge/chains/movement/src/lib.rs`
- Add `burn_moveth` function in `protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move` because it looks like the resource signer can only be used in one module. The apparent need for this workaround provides support for the idea of one unified move module for both initiator and counterparty, because then one resource signer could be associated with one module's `BridgeConfig`.
- Fix `complete_bridge_tranfer` in  `protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move`. It wasn't an entry function and it took an extra signer. Burning was not done through the resource account. Fixed all those issues. 
- Add `BridgeContractInitiator::get_bridge_transfer_detailsget_bridge_transfer_details` method
- Add `protocol-units/bridge/integration-testing/tests/movement_eth.rs` including `test_movement_client_initiate_and_complete_transfer` which initiates a transfer, gets bridge transfer details, completes the transfer, and gets transfer details again.
- Add utility functions in testing lib to streamline tests
- Modified `MovementClient::new_for_test function` to kill any running `movement` process and delete `.movement` if present, at beginning of tests

# Testing

Test:

```
rust_backtrace=1 cargo test --test movement_eth test_movement_client_initiate_and_complete_transfer -- --nocapture --test-threads=1
```

If a test fails, check whether the `movement` local network process is still running:

```
ps aux | grep movement
```

If you see something like `movement node run-localnet`, copy the process ID (`<pid>`) and kill it:

```
kill -9 <pid>
```

# Outstanding issues
- Still need to add feature flags for testing code, created issue #564 

- Putting utility functions in the counterparty module is not a very nice solution in that it's not symmetric. One module for all Move atomic bridge functions (both counterparty and initiator) would solve this. @andyjsbell is already working on something to that effect so I'm not sure how worthwhile it would be to combine the modules for this branch of development.
- I was not able to fix `refund_bridge_transfer` in the initiator Move module to use a resource account pattern, without it feeling like the PR was getting too big. I believe I can do it but this should be a separate issue, along with integration testing for that fix. Created Issue #525


0xmovses (changes requested):
This looks good, the code reads well, the assertions are tight and you've followed all the patterns elsewhere in the bridge code. Could you add an additional job in the CI for `movement_eth.rs` integration tests?

andygolay:
> This looks good, the code reads well, the assertions are tight and you've followed all the patterns elsewhere in the bridge code. Could you add an additional job in the CI for `movement_eth.rs` integration tests?

Oh yeah good call, will do rn.

.github/workflows/checks-all.yml
```diff
@@ -262,6 +262,49 @@ jobs:
...
+        cd -
+
+    - name: Install Nix
+      uses: DeterminateSystems/nix-installer-action@main
+
+    - name: Run eth_movement tests
+      run: |
+        nix develop --command bash -c "
+          rust_backtrace=1 cargo test --test movement_eth -- --nocapture --test-threads=1 
+        "
```
0xmovses, line 306: Hmm rather than a second job you could just append to the `nix develop --command` for `eth_movement` job and rename it `bridge client integration`, then we don't have to install the CLI twice, which is quite a resource heavy op.

0xmovses (changes requested):

.github/workflows/checks-all.yml
0xmovses: As in just a `&&` in the command so it will run both integration files first for eth_movement then movement_eth.

andygolay: Sure just did it as 

```
      run: |
        nix develop --command bash -c "
          rust_backtrace=1 cargo test --test eth_movement -- --nocapture --test-threads=1 
          && rust_backtrace=1 cargo test --test movement_eth -- --nocapture --test-threads=1
        "
```

andygolay: Needed the line breaks i think... should pass this time 

```
      run: |
        nix develop --command bash -c " \
          rust_backtrace=1 cargo test --test eth_movement -- --nocapture --test-threads=1 && \
          rust_backtrace=1 cargo test --test movement_eth -- --nocapture --test-threads=1 \
        "
```

andygolay: CI passes, good for re-review @0xmovses

protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -308,33 +303,35 @@ impl MovementClient {
...
-				}
-				_ if line.starts_with("origin_addr = ") => {
-					format!(r#"origin_addr = "{}""#, address)
-				}
-				_ if line.starts_with("source_account = ") => {
-					format!(r#"source_account = "{}""#, address)
-				}
-				_ => line.to_string(),
-			})
+			.map(|line| {
```
0xmovses, line 306: As this is testing code, can we move this lofgic out of the main `lib.rs` into utils and feature flag the calling method, I think its `publish_for_test` as #[cfg(test)]`, then this code will only get build when running `cargo test.

protocol-units/bridge/integration-tests/tests/movement_eth.rs
```diff
@@ -0,0 +1,463 @@
...
+
+		let mut bridge_transfer_id: Option<String> = None;
+
+		let rest_client = movement_client.rest_client();
+		let transactions = rest_client
+			.get_account_transactions(sender_address, Some(sequence_number), Some(20))
+			.await
+			.map_err(|e| anyhow::Error::msg(format!("Failed to get transactions: {:?}", e)))?;
+
+		if let Some(transaction) = transactions.into_inner().last() {
```
0xmovses, line 166: Let's put this in a utils/ inside integration tests. I don't see it as "belonging" to `TestHarness`. Currently it obstructs the main logic of the code.

0xmovses (changes requested):
Great work. I think these integration tests are doing too much for each case. For example 
`test_movement_client_initiate_and_refund_transfer`

Should be broken into to tests
`test_movement_client_should_initiate_transfer` and then
`test_movement_client_should_refund_transfer`

I know that initiate is required before refund. But doing both ops in one test makes the test hard to read and quite verbose. 

What you can do is create a fn further down that does the processing for initiate, then reuse that fn in both initiate and refund. 

Additionally for the values that you use each time for each test, 

```
let initiator_addr = AccountAddress::new(*b"0x123456789abcdef123456789abcdef");
	let recipient = b"0x123456789abcdef".to_vec();
        let preimage = "secret".to_string();
        let serialized_preimage = bcs::to_bytes(&preimage).unwrap();  // BCS serialization
        let hash_lock = *keccak256(&serialized_preimage);
	let time_lock = 1;
	let amount = 100;
```

These are repeating and also obfuscates the core thing that is being tested. In my indexer PR I actually improved this by created a method on TestHarnes that returns the default test args. 

I'm going to PR them into main and you can pull them into this branch.

0xmovses:
One thing I just noticed when checking out this PR. Are you getting clippy and rust-analyzer warnings in your IDE? I see them for me, there were quite a few unused import warning and some clippy suggestions. Let me know.

andygolay:
> One thing I just noticed when checking out this PR. Are you getting clippy and rust-analyzer warnings in your IDE? I see them for me, there were quite a few unused import warning and some clippy suggestions. Let me know.

Yes, I'm seeings the warnings. I've been ignoring them but I'll start making the suggested changes instead.

0xmovses:
Just one conflict to resolve then this is ready go in.

andygolay:
> Just one conflict to resolve then this is ready go in.

Resolved, and updated test names. It wouldn't let me re-ping you for review but should be good as soon as CI passes.

0xmovses (approved):
🚢

PR #524, opened 2024-09-04
fix: fmt

0xmovses:
## Summary
Please make sure your IDE is using `rust-analyzer` / cargo as the default formatter. It will use the defaults set in `rustfmt.toml`, otherwise this makes diffs horrible. 

Ran `find protocol-units/bridge -name "*.rs" | xargs cargo fmt --`

## Testing
No regression should be introduced


0xPrimata (approved):
LGTM

PR #531, opened 2024-09-05
fix: timelock on Initiator must be greater than counterparty

0xmovses:
# Summary
As raised by @0xPrimata the rule in the RFC was not being held in the contracts. 
https://github.com/movementlabsxyz/rfcs/blob/main/0040-atomic-bridge/rfc-0040-atomic-bridge.md#duration

# Changelog
- Updates ETH initiator and its tests 
- Updates Move initiator and its tests 

# Testing
Tests have been updated, should pass in CI


l-monninger:
@0xmovses this checks out for the counterparty claim. But, can you link to how the time-lock is constructed on the destination chain for my awareness? That time-lock should also expire a smaller fraction than either here.

0xmovses:
@l-monninger I've added a divide by 2 step on the counterparty side for eth and move destination chains. The timelock must be divisible by 2 for this to not revert.

andygolay (commented):
I don't see how this enforces end-to-end timelock management, as it's not being tested end-to-end with an initiator on one chain and counterparty on another chain. 

In the initiator Move contract, `initiate` function we have:

```
time_lock = time_lock * 4;
```

which seems fine but it's arbitrary. There's no guarantee that someone will call `complete` on the Eth side with a time lock that is equal to time_lock * 2. 

As far as I can tell these tests don't cover that. These multiplications are done in isolation of one another, and someone could feed in whatever args they want into the functions.

We could do assertions by fetching `BridgeTransferDetails` from the initiator chain after `Initiate_bridge_transfer` and comparing those details to a `BridgeTransferDetails` on the counterparty chain, and enforcing logic there, in an E2E testing context.

It depends on how much freedom we're giving users to choose their time lock. If it's always the same duration then that could help simplify things. Have we decided on that?

If not, I would think just letting the `BridgeTransferDetails` on the initiator side have the time_lock from the initiator would make sense from a UX perspective. 

And then just fetch that timelock, round down to the nearest even number, and divide by 2 as the value to feed into `lock_bridge_transfer`, rather than doing `time_lock = time_lock * 2;`

Up to you, I just think because this involves cross-chain logic, doing the multiplying within the modules and contracts won't be as effective as doing it and checking it in an end-to-end context. 

Either way I'm sure you're planning to include assertions about correct use of time in E2E testing. 

But it seems to me that this is over-complicating the contracts and modules, and it's better to handle this more purely E2E.

0xPrimata:
I would do `time_lock = time_lock * 2;` on initiator and no multiplier on counterparty.

andygolay:
> I would do `time_lock = time_lock * 2;` on initiator and no multiplier on counterparty. @andygolay it must have a higher timelock on the initiator, agree that the *4 is weird. Issue with assertions on the server stack is that that is a lot more flexible and exploitable. This is mostly a matter of removing attack vectors.
> 
> yes the idea from the MD is to always have the same timelock.

Yeah if we want to do anything on-chain for this, a simple multipying by 2 on the initiator side should cover it. Still not sure that it's needed.

When the originator is prompted to give a time lock, it could be like, "How much time do you want to give the bridge to make the swap?" 

And then if the user enters say 24 hours, tell them "The bridge has 24 hours to claim your ETH. Once it does, you'll have at least 24 hours to claim your MovETH". Ideally it happens more automatically than that but that's the basic deal.

After `initiate_brige_transfer`, what we could name `BridgeTransferDetails(originator)` is created and an event is emitted. 

The next step, "Bob responds by locking his funds with a 24-hour timelock" is a crucial piece, which can be verified E2E in a cross-chain context.

The attestors and E2E tests can check that:

```
BridgeTransferDetails(originator).time_lock = BridgeTransferDetails(counterparty).timelock * 2
```

To validate pairs of originator and counterparty bridge transfer details after `lock_bridge_transfer` is completed on the counterparty side, before the transfer is allowed to complete. If the timing is wrong (if Alice doesn't have 2x the time as Bob), then the transfer is automatically aborted and refunded.

Just one possible solution.

0xPrimata:
one issue: what if user calls with a very short timelock like 5 seconds, bridge service picks it up at second 4, and locks at time 6, this opens up an exploit.
That's why having the referred state variable or constant variable in the MD written to the contract is a much better approach. It's not like a timelock is impeding them from performing the bridge quickly, it's only impeding them from refunding, so it's not a big issue to have it "hardcoded" to the contract.

0xPrimata:
```
BridgeTransferDetails(originator).time_lock = BridgeTransferDetails(counterparty).timelock * 2
```
Believe this part is probably wrong because the timelock parameter is the deltaTimestamp and the stored timelock is the stipulated timestamp in the future, so deltaTimestamp + currentTimestamp. A bit messy but you have to check for 
```
BridgeTransferDetails(originator).time_lock = BridgeTransferDetails(counterparty).timelock + contract.timelock
```
All of this is meaningless if we just have it written to the contract. There is just no need to rely on the Bridge Relayer imo.

andygolay:
> ```
> BridgeTransferDetails(originator).time_lock = BridgeTransferDetails(counterparty).timelock * 2
> ```
> 
> Believe this part is probably wrong because the timelock parameter is the deltaTimestamp and the stored timelock is the stipulated timestamp in the future, so deltaTimestamp + currentTimestamp. A bit messy but you have to check for
> 
> ```
> BridgeTransferDetails(originator).time_lock = BridgeTransferDetails(counterparty).timelock + contract.timelock
> ```
> 
> All of this is meaningless if we just have it written to the contract. There is just no need to rely on the Bridge Relayer imo.

Ah gotcha, in the context of the MD of timelock being hard-coded this makes a lot of sense.

l-monninger:
The x, 2x, 4x is because the difference in time-lock you guys are adding is not actually the one I believe is most important.
My understanding from the original proposal:

The **INITIATOR** contract on the source chain is used for Alice to bridge in assets.

The **COUNTERPARTY** contract on the source chain is used for Bob to finally claim Alice's bridged assets.

The **CLAIM** contract on the destination chain is used for Alice to claim her bridge assets.

It's the time-lock on the **CLAIM** contract that from the original proposal needs to be x, where the initiator time-lock needs to be 2x.

This property means that the counterparty/Bob/bridge service has at least x to finally claim Alice's bridged assets.

If you want to create an additional y, 2y requirement between the **COUNTERPARTY** contract and the **INITIATOR** contract, this places the assets in limbo for y. That is, neither the **COUNTERPARTY** nor the **INITIATOR** can claim them.

This could perhaps be a desirable property if:
- Governance has the ability to intervened here or similar.
- This plays into some monitoring or claim priority scheme.

protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
```diff
@@ -117,6 +117,9 @@ module atomic_bridge::atomic_bridge_counterparty {
         recipient: address,
         amount: u64
     ) acquires BridgeTransferStore {
+        // Initiator multiplies timelock by 4, making the timelock here half the duration
+        time_lock = time_lock * 2;
```
l-monninger, line 121: Yeah, so these can be 1 and 2 if the counterparty is the destination chain.

l-monninger (changes requested):

0xmovses:
Constant of base timelock 24 hours added and tests updated.

protocol-units/bridge/contracts/src/AtomicBridgeCounterparty.sol
```diff
@@ -39,19 +43,21 @@ contract AtomicBridgeCounterparty is IAtomicBridgeCounterparty, OwnableUpgradeab
...
+        // The timelock for the counterparty is just 24 hours (constant value)
+        uint256 timeLock = block.timestamp + TIME_LOCK_DURATION;
+
         bridgeTransfers[bridgeTransferId] = BridgeTransferDetails({
             recipient: recipient,
             originator: originator,
             amount: amount,
             hashLock: hashLock,
-            timeLock: block.number + timeLock, // using block number for timelock
+            timeLock: timeLock,
```
andygolay, line 60: Since counterparty duration is fixed as 24 hours, would it make sense to name the constant COUNTERPARTY_TIME_LOCK_DURATION on the counterparty side? And INITIATOR_TIME_LOCK_DURATION on the initiator side, so its' crystal clear which time lock duration is which?

protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
```diff
@@ -10,12 +10,13 @@ module atomic_bridge::atomic_bridge_counterparty {
     use aptos_framework::aptos_hash::keccak256;
     use aptos_std::smart_table::{Self, SmartTable};
     use moveth::moveth;
-    
 
     const LOCKED: u8 = 1;
     const COMPLETED: u8 = 2;
     const CANCELLED: u8 = 3;
 
+    const TIME_LOCK_DURATION: u64 = 24 * 60 * 60; // 24 hours in seconds
```
andygolay, line 18: Why not name this COUNTERPARTY_TIME_LOCK_DURATION just so it's clearly distinct from the initiator time lock duration, then you won't need to do any multiplying on the initiator side.

protocol-units/bridge/move-modules/sources/atomic_bridge_initiator.move
```diff
@@ -17,6 +17,7 @@ module atomic_bridge::atomic_bridge_initiator {
     const COMPLETED: u8 = 2;
     const REFUNDED: u8 = 3;
 
+    const TIME_LOCK_DURATION: u64 = 24 * 60 * 60; // 24 hours in seconds
```
andygolay, line 20: What if you name this INITIATOR_TIME_LOCK_DURATION and do the multiplication by 2 here to get 48 hours? Then you won't have to multiply by 2 anywhere else and you'll have more descriptively named timelock constants.

protocol-units/bridge/contracts/src/AtomicBridgeInitiator.sol
```diff
@@ -31,6 +31,9 @@ contract AtomicBridgeInitiator is IAtomicBridgeInitiator, OwnableUpgradeable {
     IWETH9 public weth;
     uint256 private nonce;
 
+    // Constant time lock of 24 hours in seconds (86400 seconds)
+    uint256 public constant TIME_LOCK_DURATION = 24 * 60 * 60;
```
andygolay, line 35: What if you name this constant INITIATOR_TIME_LOCK_DURATION and do the multiplication by 2 here to handle the multiplication all in one place and name the constant so that it's distinct from COUNTERPARTY_TIME_LOCK_DURATION. Just maybe a little clearer for readers and avoids the need to multiply by 2 in various places throughout the modules / contracts.

andygolay (commented):
Looks great, added comments about time lock variable naming to help make the different timelock durations (counterparty vs initiator) more easily distinguishable, and so you can do the multiplication by 2 as part of the constant definition, rather than needing to do the multiplication through the module, tests, etc.

protocol-units/bridge/contracts/src/AtomicBridgeCounterparty.sol
0xmovses: That's been added

protocol-units/bridge/move-modules/sources/atomic_bridge_counterparty.move
0xmovses: Done

l-monninger (approved):
Apart from my comments about the naming, LGTM!

0xmovses:
@l-monninger yes Re. renaming I agree. But it also means renaming components in the relayer, so a rename PR incoming.

l-monninger:
> @l-monninger yes Re. renaming I agree. But it also means renaming components in the relayer, so a rename PR incoming.

Yeah, don't worry about it here.

Issue #529, opened 2024-09-05
MST-7 acceptGenesisCeremony Function Can Be Called Multiple Times


SA124:
MST-7 acceptGenesisCeremony Function Can Be Called Multiple Times
Auditor: Movebit
Code: MCR Settlement Contract
Severity: Discussion
Status: Pending

Code Location:
protocol-units/settlement/mcr/contracts/src/staking/MovementStaking.sol#61

Descriptions:
The acceptGenesisCeremony function does not check to see if it has already been called, so it can be called multiple times by the domain causing an error in the number of stakes recorded for the attester.

Suggestion:
It is recommended to check that the function has been called by a domain.

l-monninger:
See comment on #536

PR #536, opened 2024-09-06
Genesis can only be ran once if successful per domain

0xPrimata:
# Summary

https://github.com/movementlabsxyz/movement/issues/529

# Changelog

Added a mapping between domain and bool to check if a genesis ceremony has been done for the specified domain.

# Testing

Checks if running acceptGenesisCeremony() twice on the same domain reverts.

# Outstanding issues
Clarify how a genesis ceremony could fail and ran again.


l-monninger (approved):
This is reasonable. But, the original intent was that the calling domain would manage calling this method and perhaps calling it again.

PR #540, opened 2024-09-09
fix: kill `movement` local testnet process before running `eth-movement` tests

andygolay:
# Summary
- `bridge-eth-movement` CI check has been failing for a couple PRs to `main` due to process already running. The code added to the checks is designed to find the process and kill it.
- `movement init` sometimes fails. For more consistency in CI, delete `.movement` before running CI checks.

# Changelog

- added to [`.github/workflows/checks-all.yml`](https://github.com/movementlabsxyz/movement/compare/andygolay/kill-movement-local-testnet?expand=1#diff-55859d6a2123609588a7f5d13328343f6e4db169d68206d9ed338e926263a1ef):

```
    - name: Kill running movement process
      run: |
        PID=$(ps aux | grep 'movement node run-local-testnet' | grep -v grep | awk '{print $2}' | head -n 1)
        if [ -n "$PID" ]; then
          echo "Killing process with PID: $PID"
          kill -9 $PID
        else
          echo "No movement process found."
        fi
```

and

```
    - name: Delete .movement directory if present
      run: |
        cd $GITHUB_WORKSPACE
        if [ -d "protocol-units/bridge/integration-tests/.movement" ]; then
          echo ".movement directory found, deleting it."
          rm -rf protocol-units/bridge/integration-tests/.movement
        else
          echo ".movement directory not found."
        fi
```  

# Testing

- Passing CI is the test for this. I tried locally but testing packages were not satisfactory, so it seems that pushing and seeing that CI passes is the best approach.

# Outstanding issues


0xmovses:
I just merged main into this, should fix the failing solidity test, as the missing lib file was causing the CI to fail 👍

.github/workflows/checks-all.yml
```diff
@@ -253,6 +253,26 @@ jobs:
...
+    - name: Delete .movement directory if present
+      run: |
+        cd $GITHUB_WORKSPACE
+        if [ -d "protocol-units/bridge/integration-tests/.movement" ]; then
+          echo ".movement directory found, deleting it."
+          rm -rf protocol-units/bridge/integration-tests/.movement
+        else
+          echo ".movement directory not found."
+        fi
+
```
0xmovses, line 275: Ah, Now I'm looking at this. I wouldn't run execute this logic from the CI level but rather in the rust integration test logic itself, we want to keep the .yml files clean.

0xmovses (changes requested):

.github/workflows/checks-all.yml
andygolay: Okay I'll close this and add the logic into #523.

andygolay:
Close this to add the logic into #523.

PR #545, opened 2024-09-09
feat: add bridge label to workflow

andygolay:
# Summary
Added bridge label to `solidity-bridge-tests` workflow.

# Changelog

```
  solidity-bridge-tests:
    if: contains(github.event.label.name, 'bridge') 
```

# Testing

Hope it passes CI

# Outstanding issues

### DIFF
--- a/.github/workflows/checks-all.yml
+++ b/.github/workflows/checks-all.yml
@@ -142,7 +142,7 @@ jobs:
       run: nix develop --command bash  -c "just mcr-client native build.local.test -t=false"
 
   move-modules-test:
-    if: contains(github.event.label.name, 'bridge') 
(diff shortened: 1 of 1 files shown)


0xmovses:
We would want to apply this for all the bridge integration tests and move module tests also.

0xmovses:
@andygolay Just noticed there is a naming convention we should stick to. Use the label name `cicd:bridge`

andygolay:
> @andygolay Just noticed there is a naming convention we should stick to. Use the label name `cicd:bridge`

Will just close this to avoid overlapping with https://github.com/movementlabsxyz/movement/pull/549

0xmovses:
Let's merge this as #549 wont go in, @musitdev is going to bring back the missing method instead.

musitdev (approved):

PR #549, opened 2024-09-10
fix: wrong method name, add `cicd:bridge` label

0xmovses:
# Summary
- `dot_movement.try_get_or_create_config_from_json` has somehow disappeared. 
Quick fix here to keep main building until it is revived.  

# Changelog
- use `try_get_config_from_json`
- Adds bridge label to skip flaky CI

# Testing
CI should pass

### DIFF
--- a/networks/suzuka/indexer/src/main.rs
+++ b/networks/suzuka/indexer/src/main.rs
@@ -16,8 +16,8 @@ fn main() -> Result<(), anyhow::Error> {
 		.init();
 
 	let dot_movement = dot_movement::DotMovement::try_from_env()?;
-	let maptos_config = dot_movement
-		.try_get_or_create_config_from_json::<maptos_execution_util::config::Config>()?;
+	let maptos_config =
+		dot_movement.try_get_config_from_json::<maptos_execution_util::config::Config>()?;
 
 	let default_indexer_config = build_processor_conf("default_processor", &maptos_config)?;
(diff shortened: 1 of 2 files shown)


andygolay (approved):
LGTM

networks/suzuka/indexer/src/main.rs
```diff
@@ -16,8 +16,8 @@ fn main() -> Result<(), anyhow::Error> {
 		.init();
 
 	let dot_movement = dot_movement::DotMovement::try_from_env()?;
-	let maptos_config = dot_movement
-		.try_get_or_create_config_from_json::<maptos_execution_util::config::Config>()?;
+	let maptos_config =
+		dot_movement.try_get_config_from_json::<maptos_execution_util::config::Config>()?;
```
musitdev, line 20: It's not the right correction. If you do that the config is not created if absent. The methos to call is `try_get_or_create_config_from_json` I don't know why it diseapper.

musitdev (commented):
Perhaps I can do a specific PR for the `try_get_or_create_config_from_json` and you merge the CI change?

networks/suzuka/indexer/src/main.rs
0xmovses: This PR is just to ensure `main` will build, but yes if you could add that follow up PR that would be great.

0xmovses:
This was just caused my some cached files. After running `cargo clean` the error went away.

PR #574, opened 2024-09-12
Feat: Add `bridge-test` feature flag for `movement-bridge` lib test functions

andygolay:
# Summary
- A new `bridge-test` feature flag is applied to test functions in the `movement-bridge` crate.

# Changelog

- Define feature flag in `protocol-units/bridge/chains/movement/Cargo.toml`
- Add feature flag dependency declaration into project root `Cargo.toml`
- Add `#[cfg(feature = "bridge-tests")]` annotation to three test functions in `protocol-units/bridge/chains/movement/src/lib.rs`

# Testing

Passes tests, eg:

```
rust_backtrace=1 cargo test --test eth_movement --features bridge-tests -- --nocapture --test-threads=1
```

# Outstanding issues
- There's a CI/CD MCR check failure happening on globally when merging PRs. Curious as to whether it affects this PR.
- In the example test command and others similarly structured, the `--features` option must be placed before the `--` so it's considered a Cargo option.


protocol-units/bridge/chains/movement/src/lib.rs
```diff
@@ -116,6 +116,7 @@ impl MovementClient {
 		})
 	}
 
+	#[cfg(feature = "bridge-tests")]
```
musitdev, line 119: To avoid putting the test code in the prod code you can use 2 impl like : 
for the prod code
```
impl MovementClient {
  pub async fn new(_config: Config) -> Result<Self, anyhow::Error> {
    ....
}
```
And later in the code for the test, and you feature flag the whole impl.
```
#[cfg(feature = "bridge-tests")]
impl MovementClient {
   pub async fn new_for_test(....
}
```
You can do the same for config also.

andygolay: Thanks @musitdev, can the impls with the feature flag go in the same file (protocol-units/bridge/chains/movement/src/lib.rs) or is it better to refactor into a file specifically for tests?

musitdev: I don't know, there's no best way. My opinion is that we can leave them in the same file.

0xmovses: Same file is fine yes. But move the impl block for testing down, beneath the more important methods, as its a long method.

0xmovses:
As agreed with @andygolay we will close this and re-implement the feature flagging due to large changes in main.

Issue #573, opened 2024-09-12
Fix maptos-opt-executor tests, add to CI


mzabaluev:
maptos-opt-executor unit tests are currently broken in main:

> thread 'transaction_pipe::tests::test_repeated_pipe_mempool_from_api' panicked at /Users/me/.cargo/registry/src/index.crates.io-6f17d22bba15001f/tokio-1.39.2/src/runtime/blocking/shutdown.rs:51:21:
Cannot drop a runtime in a context where blocking is not allowed. This happens when a runtime is dropped from within an asynchronous context.

Fix the tests and add a job to CI to run them with every PR update.

Issue #553, opened 2024-09-10
Feature-gate integration tests and and enroll them in CI


mzabaluev:
Discussion: https://github.com/movementlabsxyz/movement/discussions/394

Any protocol tests that require external setup and cannot be modified to use mocks, should be feature-gated to not be run
by default `cargo test`. Instead, a test workflow should set up and run these tests in CI.

PR #576, opened 2024-09-12
Fix test crashes on dropping indexer runtime

mzabaluev:
# Summary
- **Categories**: `protocol-units`, `cicd`.

Fixes: #573

# Changelog

* Revert `Executor::background` to previous style returning two objects. The `IndexerRuntime` can be spun with a separate method.

# Testing

Now exercised in CI 🙌 

`cargo test -p maptos-fin-view` (currently requires the Movement env)

# Outstanding issues

Tests in `maptos-fin-view` remain to be adapted as per #553


l-monninger (changes requested):
CI appears to be hanging.

mzabaluev:
> CI appears to be hanging.

Odd, it's only one run in the matrix. I'll check if a merge with newer commits in main (which is what the workflow triggered by `pull_request` checks out) is causing this.

BTW, should we restrict the `push` trigger to well-known branches (currently 'main')? Now the checks are run largely redundantly, except for the difference in checkout as explained above.

mzabaluev:
> CI appears to be hanging.

In main, there have only been Solidity changes, this should not affect the unit tests.

Looks like we've hit a random deadlock.

mzabaluev:
```
```

mzabaluev:
The cause of the lockup is, as seen in the lower stack frame, dropping a registry entry results in a recursion where a referenced resource is removed from the registry while the lock is already held on its registry shard.

I've been trying to determine how likely is this recursion to occur in normal operation. In my testing on Mac, it occurs once after several dozen iterations. On a PC in CI it might be easier to trigger.

mzabaluev:
Tests fixed by https://github.com/movementlabsxyz/aptos-core/pull/67

l-monninger (approved):

.github/workflows/checks-all.yml
```diff
@@ -129,8 +156,8 @@ jobs:
     - name: Run Suzuka Full Node Tests Against Holesky and Local Celestia
       env: 
         CELESTIA_LOG_LEVEL: FATAL # adjust the log level while debugging
+        MCR_DEPLOYMENT_ACCOUNT_PRIVATE_KEY: ${{ secrets.MCR_DEPLOYMENT_ACCOUNT_PRIVATE_KEY }}
```
l-monninger, line 159: Why did you switch this from using the generated key?

mzabaluev: No switch, just moved this from the run script where the secret could leak into the logs (more easily) or `ps` output.

PR #597, opened 2024-09-13
High Sequence Number Gas DOS

### LINKED ISSUE #490: No gas fees for high sequence number transactions allows resource abuse
Auditor: Halborn
Severity:[CRITICAL] 
Code: Aptos Core & Suzuka

No gas fees for high sequence number transactions allows resource abuse
Transactions with sequence numbers significantly higher than expected can be added to the mempool and processed by the nodes. These transactions go through the initial validation and queuing phases without incurring any gas fees, despite ultimately failing during block execution due to their unexpected sequence numbers. This behavior allows for the repeated submission of such transactions without cost.
The absence of gas charges for these transactions presents a potential attack vector where a malicious actor can submit thousands of these high sequence number transactions, causing nodes to process them without incurring any costs. This behavior can lead to resource exhaustion as nodes allocate CPU, memory, and I/O resources to handle these transactions. Over time, this can degrade node performance, disrupt normal transaction processing, and potentially lead to denial-of-service (DoS) conditions. The vulnerability undermines the economic incentives designed to prevent network abuse, as it allows attackers to exploit node resources without any financial deterrent.
Proof of Concept:
=============
Setup: In images 1 and 2, macros are included to print where a transaction is validated and where the gas is charged during the block execution phase.
Test code:
Here it's the code of the create_fake_signed_transaction function that allows to create transactions with high sequence number and customized expiration timestamp:

Screenshot 2024-08-30 at 12 28 07 PM

Here it's the code for 2 different kind of tx that will eventually fail:
TEST 1: Sending a tx trying to transfer more coins than Alice has (including gas fees).
TEST 2: Sending 1000 txs with high sequence number

Screenshot 2024-08-30 at 12 37 06 PM
Screenshot 2024-08-30 at 12 37 15 PM
Screenshot 2024-08-30 at 12 37 21 PM

Results in the node side:
Image 3 shows that both FLAG 1 and FLAG 2 were printed for the TEST 1, indicating that the tx sent was validated during block execution and the corresponding gas fees charged.
Image 4 shows that only FLAG 1 was printed for the TEST 2, indicating that the txs sent were validated during block execution, but  the corresponding gas fees were not charged.
Results in the test:
Image 5 shows the following:
Because of the failed tx in TEST 1, gas fee was charged (700).
The 1000 txs in TEST 2 failed, but no gas was charged.

image (6)
image (7)
image (8)
image (9)
image (10)

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Resolves #490 DoS attack by simply checking by two non-exclusive methods:

1. Prohibits high sequence numbers and prevents redundant sequence numbers from entering the mempool on a single node in #661. This effectively means that attacks can only run `TOO_NEW_TOLERANCE` transactions.
2. ~~Charges base gas fees for validation only transactions in #665, i.e., transactions that fail after validation still incur the base gas fee.~~ #665 has been excluded until further verification. 

# Testing

1. Adds e2e test to match #490 attack scripts. These are grouped under `cicd:suzuka-full-node-malicious`. 
2. Adds unit tests to `TransactionPipe`.

# Outstanding issues
None


flake.nix
```diff
@@ -54,6 +54,7 @@
           gcc
           rust
           postgresql
+          tesseract4
```
mzabaluev, line 57: What is this for?

l-monninger: This is for OCRing the code from the issue screenshots.

protocol-units/execution/opt-executor/src/transaction_pipe.rs
```diff
@@ -149,7 +151,28 @@ impl TransactionPipe {
...
+		// this checks that the sequence number is too old or too new
+		let sequence_number =
+			vm_validator::get_account_sequence_number(&state_view, transaction.sender())?;
+		if transaction.sequence_number() < sequence_number {
+			let status = MempoolStatus::new(MempoolStatusCode::InvalidSeqNumber);
+			println!("Transaction sequence number too old: {:?}", transaction.sequence_number());
+			return Ok((status, Some(DiscardedVMStatus::SEQUENCE_NUMBER_TOO_OLD)));
+		}
+
+		if transaction.sequence_number() > (sequence_number + TOO_NEW_TOLERANCE) {
```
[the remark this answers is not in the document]
l-monninger: Yes, but they won't be able to send very many--depending on how low the tolerance is.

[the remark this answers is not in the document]
l-monninger: Oh, wait no, that also doesn't work.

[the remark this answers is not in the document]
l-monninger: Let me send in the more invasive PR.

flake.nix
mzabaluev: This is a one-off then, no need to force every developer shell to install this.

networks/suzuka/suzuka-client/src/bin/e2e/gas_dos.rs
```diff
@@ -0,0 +1,193 @@
...
+	// assert gas fee charged
+	let failed_balance = coin_client
+		.get_account_balance(&alice.address())
+		.await
+		.context("Failed to get Alice's account balance")?;
+	println!("\n=== After Failed Tx#1 ===");
+	println!("Alice: {:?}", failed_balance);
+	assert!(initial_balance > failed_balance);
+
+	// TEST 2: Sending n transactions with a high sequence number
```
mzabaluev, line 166: I see a structural problem here: two tests are rolled into one binary, perhaps because creating binary targets for each test is work-intensive and add a lot boilerplate.

If each of the tests had a `#[test]` entry point, and maybe delegated the setup to a common function, this could be developed as a normal test suite.

process-compose/suzuka-full-node/process-compose.gas-dos.yml
```diff
@@ -0,0 +1,16 @@
+version: "3"
+
+environment:
+
+processes:
+
+  gas-dos-test:
```
mzabaluev, line 7: Do we really need an overlay for each small test area? The test run could be rolled into `process-compose.test.yml` since it needs the same process-compose environment.

flake.nix
l-monninger: I think it's useful. Let me put a Polly up for a couple hours and see.

.github/workflows/checks-all.yml
```diff
@@ -101,9 +101,11 @@ jobs:
         CELESTIA_LOG_LEVEL: FATAL # adjust the log level while debugging
       run: |
         nix develop --command bash  -c "just suzuka-full-node native build.setup.eth-local.celestia-local.test -t=false"
+        nix develop --command bash  -c "just suzuka-full-node native build.setup.eth-local.celestia-local.test -t=false"
```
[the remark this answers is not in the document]
l-monninger: It checks that we pick up persisted state.

networks/suzuka/suzuka-client/src/bin/e2e/gas_dos.rs
l-monninger: All of these conditions should hold to prove the `gas_dos` does not occur. But, if you want to adopt a standard of minimization, I can do so. 

I thought we were trying to avoid writing e2e tests under the `#[test]` macro.

process-compose/suzuka-full-node/process-compose.gas-dos.yml
l-monninger: I personally prefer it for targeting different tests. One thing I was thinking is that we could update the scripts to have a `/test` directory where it reads additional overlays. It might help with the organization.

networks/suzuka/suzuka-client/src/bin/e2e/gas_dos.rs
mzabaluev: I'd still prefer `#[test]` entry points because they are easier to maintain in the long run and they plug into the standard `cargo test` tool.

OK with this being a two-step test, though.

.github/workflows/checks-all.yml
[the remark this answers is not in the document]
mzabaluev: Right; I remember there was a comment about this in one of @musitdev's PRs, but it's not been merged yet.

protocol-units/execution/opt-executor/src/gc_account_sequence_number.rs
```diff
@@ -0,0 +1,157 @@
...
+			sequence_number_ttl_ms,
+			gc_slot_duration_ms,
+			sequence_number_lifetimes: BTreeMap::new(),
+		}
+	}
+
+	/// Gets a sequence number for an account
+	pub(crate) fn get_sequence_number(&self, account: &AccountAddress) -> Option<u64> {
+		// check each slot for the account
+		for (_slot, lifetimes) in self.sequence_number_lifetimes.iter().rev() {
```
mzabaluev, line 28:
```suggestion
		for lifetimes in self.sequence_number_lifetimes.values().rev() {
```

protocol-units/execution/opt-executor/src/gc_account_sequence_number.rs
```diff
@@ -0,0 +1,157 @@
...
+			}
+		}
+
+		None
+	}
+
+	/// Removes the sequence number for an account.
+	pub(crate) fn remove_sequence_number(&mut self, account_address: &AccountAddress) {
+		// check each slot for the account
+		for (_slot, lifetimes) in self.sequence_number_lifetimes.iter_mut().rev() {
```
mzabaluev, line 45:
```suggestion
		for lifetimes in self.sequence_number_lifetimes.values_mut().rev() {
```

protocol-units/execution/opt-executor/src/gc_account_sequence_number.rs
```diff
@@ -0,0 +1,157 @@
...
+	/// This should be called periodically.
+	pub(crate) fn gc(&mut self, current_time_ms: u64) {
+		let gc_slot = current_time_ms / self.gc_slot_duration_ms;
+
+		// remove all slots that are too old
+		let slots_to_remove: Vec<u64> = self
+			.sequence_number_lifetimes
+			.range(..gc_slot - self.sequence_number_ttl_ms / self.gc_slot_duration_ms)
+			.map(|(slot, _)| *slot)
+			.collect();
```
mzabaluev, line 82:
```suggestion
		let slot_cutoff = gc_slot - self.sequence_number_ttl_ms / self.gc_slot_duration_ms;
		let slots_to_remove: Vec<u64> = self
			.sequence_number_lifetimes
			.keys()
			.take_while(|slot| *slot < slot_cutoff)
			.cloned()
			.collect();
```

mzabaluev (approved):

PR #604, opened 2024-09-17
Centralized Signed Blobs

l-monninger:
# Summary
- **RFCs**: [MD-6](https://github.com/movementlabsxyz/MIP/pull/6)
- **Issues:** #563 

Implements trusted-only blob reading and writing with the following:
- Signs blocks sent to DA.
- Checks signature on block was made by well-known signers. 
- Moves compression and decompression of blobs for Celestia down into DA. 

# Testing

1. e2e tests confirm Suzuka Behavior is unchanged.
2. Unit tests for formatting and basic verification included. 

# Outstanding issues
1. Lots of potential for unit tests here. 
2. Though this is using `secp256k1`, the key serialization and deserialization format is not yet compatible with ETH hex encoding: https://github.com/movementlabsxyz/movement/issues/710
3. There should probably be an e2e test for a malicious signer that is ignored: https://github.com/movementlabsxyz/movement/issues/711


networks/suzuka/suzuka-full-node/src/da_db.rs
```diff
@@ -32,7 +32,7 @@ impl DaDB {
 		Ok(Self { inner: Arc::new(db) })
 	}
 
-	pub async fn add_executed_block(&self, id: String) -> Result<(), anyhow::Error> {
+	pub async fn add_executed_block(&self, id: Vec<u8>) -> Result<(), anyhow::Error> {
```
mzabaluev, line 35: Should we just use an `Id` domain type?

proto/movementlabs/protocol_units/da/m1/light_node/v1beta1.proto
```diff
@@ -4,17 +4,12 @@ package movementlabs.protocol_units.da.m1.light_node.v1beta1;
 
 // Request and response messages
 message Blob {
-    string blob_id = 1;
+    bytes blob_id = 1;
     bytes data = 2;
     uint64 height = 3;
-    // bytes signature = 4; // at some point a signature will be added here
+    bytes signature = 4; // at some point a signature will be added here
```
mzabaluev, line 10: Remove the comment, that point is now.

```suggestion
    bytes signature = 4;
```

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
```diff
@@ -1,79 +1,70 @@
-use crate::Verifier;
-use celestia_rpc::{BlobClient, Client, HeaderClient};
+use crate::{Error, Verified, VerifierOperations};
+use celestia_rpc::Client;
+#[cfg(feature = "pessimistic")]
+use celestia_rpc::{BlobClient, HeaderClient};
 use celestia_types::{nmt::Namespace, Blob};
-use m1_da_light_node_grpc::VerificationMode;
+use m1_da_light_node_util::inner_blob::InnerBlob;
```
mzabaluev, line 6: `InnerBlob` is an interesting name for public API...

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
```diff
@@ -1,79 +1,70 @@
...
-		_height: u64,
-	) -> Result<bool, anyhow::Error> {
-		unimplemented!()
-	}
+#[tonic::async_trait]
+impl VerifierOperations<Blob, InnerBlob> for Verifier {
+	/// Verifies a Celestia Blob as a Valid InnerBlob
+	async fn verify(&self, blob: Blob, _height: u64) -> Result<Verified<InnerBlob>, Error> {
+		//@l-monninger: the light node itself does most of the work of verify blobs. The verification under the feature flag below is useful in zero-trust environments.
+		#[cfg(feature = "pessimistic")]
```
mzabaluev, line 26: Changing behavior with features is an antipattern: features are additive in a workspace, so if this library has a client A that expects the `not(feature = "pessimistic")` behavior, but there is another crate B that enables the "pessimistic" feature, the client A may get unexpected results or unwanted performance hit.
I think it's better to have a separate method that invokes the pessimistic verification.

protocol-units/da/m1/light-node/src/v1/manager.rs
```diff
@@ -12,7 +24,14 @@ where
...
 
 // Implements a very simple manager using a marker strategy pattern.
-impl Manager<LightNodeV1> {
+impl<C> Manager<LightNodeV1<C>>
+where
+	C: PrimeCurve + CurveArithmetic + DigestPrimitive + PointCompression,
+	Scalar<C>: Invert<Output = CtOption<Scalar<C>>> + SignPrimitive<C>,
+	SignatureSize<C>: ArrayLength<u8>,
+	AffinePoint<C>: FromEncodedPoint<C> + ToEncodedPoint<C> + VerifyPrimitive<C>,
+	FieldBytesSize<C>: ModulusSize,
```
mzabaluev, line 33: Perhaps it would be good to add a helper trait that represents these lengthy bounds.

protocol-units/da/m1/util/src/config/common.rs
```diff
@@ -1,5 +1,20 @@
...
+		generic_array::ArrayLength,
+		ops::Invert,
+		point::PointCompression,
+		sec1::{FromEncodedPoint, ModulusSize, ToEncodedPoint},
+		subtle::CtOption,
+		AffinePoint, CurveArithmetic, FieldBytesSize, PrimeCurve, Scalar,
+	},
+	hazmat::{DigestPrimitive, SignPrimitive, VerifyPrimitive},
+	SignatureSize,
+};*/
```
mzabaluev, line 14: Just remove this.

networks/suzuka/suzuka-full-node/src/da_db.rs
l-monninger: Yeah, we should; was going to leave some comments about that.

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
l-monninger: I thought you were going to say that. I'll split it out into another verifier.

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
l-monninger: Yeah, I wasn't sure what we would best do here. We already have `Blob` from gRPC and `Blob` from Celestia. My thinking here was that this would be a consistent intermediate type that would be leveraged regardless of the underlying DA.

protocol-units/da/m1/light-node/src/v1/manager.rs
l-monninger: Was going to call it `SigningCurve` btw.

protocol-units/da/m1/util/src/config/common.rs
l-monninger: I need to come back to it for serialization and deserialization.

protocol-units/da/m1/light-node/src/v1/manager.rs
l-monninger: Yeah, I think it's the associated types that are making this tricky, but here are all of the variants I have tried...

```rust
pub trait SigningCurve: PrimeCurve + CurveArithmetic + DigestPrimitive + PointCompression
where
	SignatureSize<Self>: ArrayLength<u8>,
	AffinePoint<Self>: FromEncodedPoint<Self> + ToEncodedPoint<Self> + VerifyPrimitive<Self>,
	FieldBytesSize<Self>: ModulusSize,
{
	// You can define other methods or associated types specific to curve operations
}

pub trait SigningKeyRequirements<C: SigningCurve>
where
	Scalar<C>: Invert<Output = CtOption<Scalar<C>>> + SignPrimitive<C>,
{
	// You can define any signing-specific logic here if needed
}
```

```rust
pub trait SigningCurve: PrimeCurve + CurveArithmetic + DigestPrimitive + PointCompression
where
	Scalar<Self>: Invert<Output = CtOption<Scalar<Self>>> + SignPrimitive<Self>,
	SignatureSize<Self>: ArrayLength<u8>,
	AffinePoint<Self>: FromEncodedPoint<Self> + ToEncodedPoint<Self> + VerifyPrimitive<Self>,
	FieldBytesSize<Self>: ModulusSize,
{
}
```

```rust
pub trait SigningCurve<C> :
where
	C: PrimeCurve + CurveArithmetic + DigestPrimitive + PointCompression,
	Scalar<C>: Invert<Output = CtOption<Scalar<C>>> + SignPrimitive<C>,
	SignatureSize<C>: ArrayLength<u8>,
	AffinePoint<C>: FromEncodedPoint<C> + ToEncodedPoint<C> + VerifyPrimitive<C>,
	FieldBytesSize<C>: ModulusSize,
{
}
```

mzabaluev: Yeah, I've played with the second variant and did not get far.

The `ecdsa` API is ungodly abstract, could we achieve the same with [ed25519-dalek](https://docs.rs/ed25519-dalek/latest/ed25519_dalek/struct.SigningKey.html)?

mzabaluev: Maybe just replace the generic parameter with the [specific curve](https://docs.rs/k256/latest/k256/struct.Secp256k1.html), and get into all this abstract bound crap when we actually need it.

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
mzabaluev: Or "transaction payload", but we need to name it consistently throughout then.

protocol-units/da/m1/light-node-verifier/src/permissioned_signers/mod.rs
```diff
@@ -45,7 +45,7 @@ where
 	pub fn new(
 		celestia_client: Arc<Client>,
 		celestia_namespace: Namespace,
-		known_signers_sec1_bytes: HashSet<Vec<u8>>,
+		known_signers_sec1_bytes: HashSet<String>,
 	) -> Self {
```
mzabaluev, line 49: This could be more generic to allow passing in e.g. an array of string literals:

```rust
	pub fn new<T>(
		celestia_client: Arc<Client>,
		celestia_namespace: Namespace,
		known_signers_sec1_bytes: T,
	) -> Self
        where
            T: IntoIterator,
            T::Item: Into<String>,
        {
```

protocol-units/da/m1/util/Cargo.toml
```diff
@@ -43,6 +43,7 @@ zstd = { workspace = true }
 bcs = { workspace = true }
 ecdsa = { workspace = true, features = [ "signing", "verifying", "der"] }
 k256 = { workspace = true }
+rand = { version = "0.8.5" }
```
mzabaluev, line 46: Use the workspace version?

protocol-units/da/m1/light-node/src/main.rs
```diff
@@ -13,7 +14,8 @@ async fn main() -> Result<(), Box<dyn std::error::Error>> {
 	let dot_movement = dot_movement::DotMovement::try_from_env()?;
 	let config_path = dot_movement.get_config_json_path();
 	let config_file = tokio::fs::File::open(config_path).await?;
-	let manager = Manager::<LightNodeV1>::new(config_file).await?;
+	// todo: consider whether LightNode implementation should encapsulate signing type
```
0xPrimata, line 17: should probably be generic if target is multiple possible DAs, correct?

protocol-units/da/m1/util/src/config/local/m1_da_light_node.rs
```diff
@@ -1,13 +1,90 @@
...
+		Ok(val) => {
+			// decode from hex to bytes 32
+			let hex_bytes = hex::decode(val).expect("Invalid hex string");
+
+			// todo: maybe remove the unwrap and catch for a random signing key
+			let signing_key_bytes: &[u8; 32] =
+				hex_bytes.as_slice().try_into().expect("Slice with incorrect length");
+			SigningKey::from_bytes(signing_key_bytes.into()).unwrap()
+		}
+		Err(_) => SigningKey::random(
```
mzabaluev, line 30: This case could be for the unset value:

```suggestion
		Err(env::VarError::NotPresent) => SigningKey::random(
```

And then the other error case could be panicked on, but that's finessing.

protocol-units/da/m1/util/src/config/local/m1_da_light_node.rs
```diff
@@ -1,13 +1,90 @@
...
+			// rand_core
+			&mut rand::rngs::OsRng,
+		),
+	}
+}
+
+pub fn default_da_signers_sec1_keys() -> HashSet<String> {
+	match std::env::var("DA_SIGNERS_SEC1_KEYS") {
+		Ok(val) => val.split(',').map(|s| s.to_string()).collect(),
+		Err(_) => HashSet::new(),
```
mzabaluev, line 40: Same here.

0xPrimata (commented):
Not sure if I could be of much help to evaluate code.

protocol-units/da/m1/util/src/config/local/m1_da_light_node.rs
```diff
@@ -1,13 +1,90 @@
...
 };
+use ecdsa::SigningKey;
+use k256::Secp256k1;
 use serde::{Deserialize, Serialize};
 use std::collections::HashSet;
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct DaSigners {
+	pub da_signing_private_key_hex: String,
+	pub da_signers_public_keys_hex: HashSet<String>,
```
mzabaluev, line 15: These names are quite a handful to type if they are meant to be used in JSON and similar schemas. Why do we need to nag the user every time that this is hex, when it should be documented (in the doc comments for this structure and in the example config)?

mzabaluev: The containing structure does not have `#[serde(flatten)]` as a field in `Config`, so
the full field name would be resolved like `da_signers.da_signers_public_keys_hex`. I think it's poor UX. Plug for more discussion in #533.

protocol-units/da/m1/light-node/src/main.rs
l-monninger: At the level of running in `main.rs`, this is where we have to decide. So, it's generic in the lib, but the bin shan't be.

protocol-units/da/m1/util/Cargo.toml
l-monninger: It conflicts with Aptos. Aptos is on `0.7` ecdsa is on `0.8`. I will leave a comment.

protocol-units/da/m1/light-node-verifier/src/permissioned_signers/mod.rs
l-monninger: Will update.

networks/suzuka/suzuka-full-node/src/tasks/execute_settle.rs
```diff
@@ -130,17 +131,11 @@ where
...
 		}
 
-		// decompress the block bytes
-		let block = tokio::task::spawn_blocking(move || {
-			let decompressed_block_bytes = zstd::decode_all(&block_bytes[..])?;
-			let block: Block = bcs::from_bytes(&decompressed_block_bytes)?;
-			Ok::<Block, anyhow::Error>(block)
-		})
-		.await??;
+		let block: Block = bcs::from_bytes(&block_bytes[..])?;
```
mzabaluev, line 134: I think the removal of compression should be noted in the description to the PR.

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+	/// Verifies a Celestia Blob as a Valid InnerBlob
+	async fn verify(&self, blob: Blob, _height: u64) -> Result<Verified<InnerBlob>, Error> {
+		// Only assert that we can indeed get an InnerBlob from the Blob
+		let inner_blob = InnerBlob::try_from(blob).map_err(|e| Error::Internal(e.to_string()))?;
+
+		Ok(Verified::new(inner_blob))
+	}
+}
+
+pub mod pessimistic {
```
mzabaluev, line 32: Would be cleaner in a separate file. and what is `super::Verifier` now, "optimistic"? Maybe better to put it in a module alongside to keep the code balanced.

protocol-units/da/m1/light-node-verifier/src/lib.rs
```diff
@@ -1,44 +1,44 @@
...
+	#[error("verifier internal error: {0}")]
+	Internal(String),
+	#[error("verifier validation error: {0}")]
+	Validation(String),
+}
+
+/// thiserror for validation and internal errors
+#[derive(thiserror::Error, Debug)]
+
+/// A verified outcome. Indicates that input of A is verified as valid instance of B, or else invalid instance.
```
mzabaluev, line 20: What's A here?

networks/suzuka/suzuka-full-node/src/tasks/execute_settle.rs
l-monninger: It isn't removed it's moved down into the DA, which is noted in the linked discussions. But, I will add.

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
l-monninger: I will move it out.

protocol-units/da/m1/light-node-verifier/src/lib.rs
l-monninger: Used to have a Phantom Type `A` indicated `B` was verified from `A`. I removed for immediate convenience. Will updated docs.

l-monninger: Actually, that usage is still implied by `VerifierOperations`. I'm not going to revert the type signatures until we have better agreement, but I will just slightly modify the comment.

protocol-units/da/m1/util/src/ir_blob.rs
```diff
@@ -95,38 +95,38 @@ impl InnerSignedBlobV1 {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub enum InnerBlob {
+pub enum IntermediateBlobRepresentation {
```
mzabaluev, line 98: Now you went overboard, I think `IntermediateBlob` would be fine, but OK :)

networks/suzuka/suzuka-full-node/src/da_db.rs
mzabaluev: I see, the thing is we don't have domain types to represent these values yet, so we just operate with the raw types that prost-build gives us.

But since da_db is an internal module of suzuka-full-node at this point, it's not that important.

networks/suzuka/suzuka-full-node/src/tasks/execute_settle.rs
```diff
@@ -130,17 +131,11 @@ where
...
 
 		// get the transactions
 		let transactions_count = block.transactions().len();
-		let span = info_span!(target: "movement_timing", "execute_block", id = %block_id);
+		let span = info_span!(target: "movement_timing", "execute_block", id = ?block_id);
```
mzabaluev, line 138: The debug representation of `Vec` is going to be a weird microformat in telemetry (which #697 will turn this span into); do a hex here as well?

protocol-units/da/m1/light-node-verifier/src/celestia/mod.rs
```diff
@@ -1,82 +1,35 @@
...
-		_verification_mode: VerificationMode,
-		_blob: &[u8],
-		_height: u64,
-	) -> Result<bool, anyhow::Error> {
-		unimplemented!()
+		Ok(Verified::new(ir_blob))
 	}
 }
 
+pub mod pessimistic;
```
mzabaluev, line 32: Good.

protocol-units/da/m1/light-node-verifier/src/lib.rs
```diff
@@ -1,44 +1,44 @@
...
+	pub fn into_inner(self) -> B {
+		self.0
+	}
+}
+
+#[tonic::async_trait]
+pub trait VerifierOperations<A, B>
+where
+	A: Send + Sync + 'static,
+	B: Send + Sync + 'static,
```
mzabaluev, line 41: The `Sync` is probably falling out of some structuring of async methods and is not really required, but no need to obsess about it.

protocol-units/da/m1/light-node-verifier/src/signed/mod.rs
```diff
@@ -0,0 +1,143 @@
...
+			return Err(Error::Validation("signer not in known signers".to_string()));
+		}
+
+		Ok(ir_blob)
+	}
+}
+
+#[cfg(test)]
+pub mod tests {
+	/*use ecdsa::SigningKey;
```
mzabaluev, line 122: Why is this commented out?

mzabaluev: There is a good way to leave tests around that don't work, but you want to revisit some day: `#[ignore]`.

But if it does not compile at all, there should be a TODO comment explaining why it's left here.

networks/suzuka/suzuka-full-node/src/tasks/execute_settle.rs
l-monninger: Sure, I can do a hex.

protocol-units/da/m1/util/src/ir_blob.rs
l-monninger: Do you want me to make this change or just stick with the above?

protocol-units/da/m1/light-node-verifier/src/signed/mod.rs
l-monninger: Yeah, it needs to be rewritten due to changes in the API, but I will simply remove.

protocol-units/da/m1/util/src/ir_blob.rs
mzabaluev: Whatever feels right.

mzabaluev (approved):

PR #610, opened 2024-09-18
Use script to init submodules

0xmovses:
# Summary
Adding this script allows us to have duplicate git submodules around the workspace with `--force`. It's not possible to achieve this with the normal `.gitmodules` config file. For example:

With `.gitmodules` if you have a repo `bar` set to two paths :
`bar` path = `/baz`
`bar`  path = `/foo`

It will ignore foo, meaning `bar` will not get installed at `/foo`. In our case we want this to happen when we want openzeppelin lib to be present in both `mcr/` and `bridge/` directories. 

Doing it with this script seemed like the best way to get what we need with the most minimal change. We may want to put all our solidity code in one place later, but that would be a bigger change and may have some undesirable consequences. 

NB After running the script this will autogenerate a `.gitmodules` file. 

# Changelog
- Adds script for initialising gitmodules 

# Testing
`forge build` should work for mcr and bridge directories. 

# Outstanding issues


andygolay (commented):
`forge build` worked for me in the bridge contracts dir:

```
[⠘] Compiling...
[⠒] Compiling 47 files with 0.8.26
[⠢] Solc 0.8.26 finished in 1.20s
Compiler run successful!
```

In the mcr contracts dir the output for `forge build` was:

```
[⠃] Compiling...2024-09-18T23:14:13.280531Z ERROR foundry_compilers::artifacts: error=".../movement/protocol-units/settlement/mcr/contracts/lib/forge-std/src/Script.sol": No such file or directory (os error 2)
[⠊] Compiling...
Error: 
failed to resolve file: ".../movement/protocol-units/settlement/mcr/contracts/lib/forge-std/src/Script.sol": No such file or directory (os error 2); check configured remappings
        --> .../movement/protocol-units/settlement/mcr/contracts/script/DeployMCRDev.s.sol
        forge-std/Script.sol
        ```

0xmovses:
Ah I think the script needs updating, will do in the morn, thanks!

0xmovses:
Interestingly @andygolay I get the same error on main when building in the MCR directory. And doing a `git submodule update --remote --recursive`

PR #617, opened 2024-09-20
Bridge swap state logic.

musitdev:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/state_logic.md
+++ b/protocol-units/bridge/state_logic.md
@@ -1,0 +1,57 @@
+# Bridge swap state logic definition
+
+## Normal swap process state changes
+
+```mermaid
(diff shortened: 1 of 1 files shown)


protocol-units/bridge/state_logic.md
andygolay, line 1: - Is "Done" specific enough? Or will we want to know whether the state is "Completed" or "Cancelled/Aborted/Refunded"

- Are we using "swap" as a term equivalent tho "transfer" in the contracts? Or is there any subtle difference between how we use "swap" and "transfer"? I think "swap" makes more sense if we were to choose one term or another and implement it consistently throughout the codebase.

Issue #404, opened 2024-08-21
Rework async processing in m1-da-light-node to actor pattern


mzabaluev:
Rework the  m1-da-light-node crate to the actor pattern as discussed in #340.

Code smells to address:
* `Clone` on state that is also used by background tasks.
* `Arc<_>` and `Arc<RwLock<_>>` members.
* `run_backgound_tasks` as method taking `&self`.

PR #620, opened 2024-09-24
Load shedding in memseq

### LINKED ISSUE #413: SEQ-1 Lack of Garbage Collection Mechanism in DA light node mempool
SEQ-1 Lack of Garbage Collection Mechanism in Mempool
Auditor: Movebit
Code: Aptos Core

Severity: Major
Discovery Methods: Manual Review
Status: Pending
Code Location:
protocol-units/da/m1/light-node/src/v1/sequencer.rs#19
Descriptions:
A complete mempool should have a garbage collection mechanism, also known as a
transaction aging mechanism, which periodically removes expired and unused transactions to
prevent the mempool from growing indefinitely.
Since Movement's mempool is based on RocksDB and stored on disk, the issue becomes
more severe. In this case, a node crash or server restart will not result in the deletion of the
old, expired transactions.
Suggestion:
It is recommended to establish a mempool garbage collection mechanism by referencing
Aptos.

mzabaluev:
Fixes #413

# Summary

Limit uncontrollable growth of sequencer's mempool by maintaining
a simple in-memory limit.

# Changelog

* Limit the number of transactions in DA sequencer's mempool.

# Testing

Tests to be added in the memseq and m1-da-light-node to test behavior against a configured limit.

# Outstanding issues

### DIFF
--- a/protocol-units/da/m1/light-node/src/v1/sequencer.rs
+++ b/protocol-units/da/m1/light-node/src/v1/sequencer.rs
@@ -270,11 +270,10 @@ impl LightNodeV1 {
 				}
 				Err(e) => {
 					info!("block proposer failed: {:?}", e);
+					return Err(e);
 				}
 			}
 		}
-
-		Ok(())
 	}
 
 	pub fn to_sequenced_blob_block(
--- a/protocol-units/da/m1/light-node/src/v1/sequencer.rs
+++ b/protocol-units/da/m1/light-node/src/v1/sequencer.rs
@@ -1,27 +1,28 @@
-use std::{
-	sync::{atomic::AtomicU64, Arc},
-	time::Duration,
+use std::boxed::Box;
+use std::fmt;
+use std::path::PathBuf;
+use std::pin::Pin;
+use std::sync::{atomic::AtomicU64, Arc};
+use std::time::Duration;
+
+use tokio::{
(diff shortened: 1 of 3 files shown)


mzabaluev:
The draft code does not work, because `Memseq` follows the same cloneable immutable god object pattern as the other units I have refactored in #308. This can be justified by a need for parallelized operation, and I can see tests for this, but I'm not sure if there is a real benefit because the performance of the underlying RocksDB mempool should be filesystem-bound. So I can either turn the `transactions_in_flight` counter into an atomic, or go into refactoring the whole thing into actor pattern as requested in #404.

l-monninger:
> can either turn the transactions_in_flight counter into an atomic, 
Just go for this for now. I will review in detail in about an hour. 

@mzabaluev

mzabaluev:
> > can either turn the transactions_in_flight counter into an atomic,
> Just go for this for now. I will review in detail in about an hour.

Synchronization of an atomic counter with potentially fallible ops is also tricky, to the point that I'm considering a global lock. Which will, of course, negate the last benefits of parallel operation.

How important is it to have `Memseq` cloneable and `Send` so that it can be populated concurrently from multiple gRPC service tasks?

mzabaluev:
Suspended in favor of #628.

PR #621, opened 2024-09-24
Atomically update transaction mempool database

### LINKED ISSUE #322: LIB-2 Lack of Atomicity in Multiple Database Write Operations in mempool transaction handling methods
Severity: Informational
Status: Pending
Auditor: Movebit
Topic:Suzuka-only custom components

LIB-2 Lack of Atomicity in Multiple Database Write Operations in mempool transaction handling methods

Code Location:
protocol-units/mempool/move-rocks/src/lib.rs#92,93,107,111,152,158

Descriptions:
In the current implementation of the RocksdbMempool, the mempool transaction handling
methods add_mempool_transaction, remove_mempool_transaction, and
pop_mempool_transaction perform multiple write operations to the database. However,
these operations are not atomic, meaning that if an error occurs during one of these write
operations, it can leave the database in an inconsistent state. Specifically, the issue arises
because each method makes multiple calls to delete_cf and put_cf without wrapping
them in a batch operation to ensure atomicity.

Suggestion:
To ensure atomicity and maintain database consistency, it is recommended to use RocksDB's
batch operations for all database writes within these methods. By grouping multiple write
operations into a single batch, we can guarantee that either all operations succeed or none
of them are applied, thus preserving the integrity of the database.

mzabaluev:
Fixes #322

# Summary

Use `WriteBatch` to atomically update transaction column families in RocksDB mempool.

# Testing

`cargo test -p move-rocks`

Existing tests should pass in CI.

### DIFF
--- a/protocol-units/mempool/move-rocks/src/lib.rs
+++ b/protocol-units/mempool/move-rocks/src/lib.rs
@@ -120,6 +120,10 @@ impl MempoolTransactionOperations for RocksdbMempool {
 				.cf_handle(cf::TRANSACTION_LOOKUPS)
 				.ok_or_else(|| Error::msg("CF handle not found"))?;
 
+			// Add the transactions and update the lookup table atomically
+			// in a single write batch.
+			// https://github.com/movementlabsxyz/movement/issues/322
+
 			let mut batch = WriteBatch::default();
 
 			for transaction in transactions {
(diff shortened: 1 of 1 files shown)


l-monninger (changes requested):
This is correct, but please add comments referring to the original motivation for these changes, s.t. this is all preserved.

l-monninger (approved):
LGTM

PR #624, opened 2024-09-25
Add state logic to bridge code

musitdev:
# Summary

All new code is put in the service crate. I copy the client and type code from bridge code crates and updated them in the service crate. 

# Changelog

# Testing

# Outstanding issues


protocol-units/bridge/integration-tests/tests/relayer_eth_movement.rs
andygolay, line 1: @musitdev is this file supposed to be empty, just as a placeholder for now?

PR #628, opened 2024-09-26
Garbage collection in DA light node sequencer

mzabaluev:
Fixes: #413

# Summary

Add functionality to periodically iterate through RocksDB mempool transactions, earliest first, and delete transactions that have not made it into a block before the timestamp threshold. In the DA light node sequencer, use the heuristic of setting the threshold to twice the block building time.

# Changelog

* Perform periodic database sweeps to remove outdated transactions from the DA light node mempool.

# Testing

TODO: add unit testing in mempool.

# Outstanding issues

* Need to define a realistic e2e test scenario where this behavior could be verified.

### DIFF
--- a/protocol-units/mempool/move-rocks/src/lib.rs
+++ b/protocol-units/mempool/move-rocks/src/lib.rs
@@ -405,11 +405,12 @@ impl MempoolBlockOperations for RocksdbMempool {
 }
 
 #[cfg(test)]
-pub mod test {
+pub mod tests {
 
 	use super::*;
 	use movement_types::transaction::Transaction;
 	use tempfile::tempdir;
+	use tokio::time::{sleep, Duration};
 
 	#[tokio::test]
 	async fn test_rocksdb_mempool_basic_operations() -> Result<(), Error> {
@@ -452,6 +453,32 @@ pub mod test {
 		assert_eq!(Some(transaction.clone()), transaction2);
 		mempool.remove_transaction(transaction_id.clone()).await?;
 		assert!(!mempool.has_transaction(transaction_id.clone()).await?);
+
+		Ok(())
+	}
+
+	#[tokio::test]
+	async fn test_rocksdb_gc() -> Result<(), Error> {
+		let temp_dir = tempdir().unwrap();
+		let path = temp_dir.path().to_str().unwrap();
+		let mempool = RocksdbMempool::try_new(path)?;
+
+		let transaction1 = MempoolTransaction::at_time(Transaction::new(vec![1], 0), 2);
+		let transaction1_id = transaction1.id();
+		mempool.add_mempool_transaction(transaction1).await?;
+		assert!(mempool.has_transaction(transaction1_id).await?);
+
+		sleep(Duration::from_secs(2)).await;
+
+		let transaction2 = MempoolTransaction::at_time(Transaction::new(vec![2], 0), 64);
+		let transaction2_id = transaction2.id();
+		let transaction2_timestamp = transaction2.timestamp;
+		mempool.add_mempool_transaction(transaction2).await?;
+
+		mempool.gc_mempool_transactions(transaction2_timestamp).await?;
+
+		assert!(!mempool.has_transaction(transaction1_id).await?);
+		assert!(mempool.has_transaction(transaction2_id).await?);
 
 		Ok(())
 	}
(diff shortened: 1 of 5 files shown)


protocol-units/mempool/util/src/lib.rs
```diff
@@ -57,6 +59,11 @@ pub trait MempoolTransactionOperations {
 		Ok(mempool_transactions)
 	}
 
+	fn gc_mempool_transactions(
+		&self,
```
l-monninger, line 63: I don't think this is a product of the refactor here. But, the `Ord` implementation is incorrect. The slot seconds comparison is not useful. It should be compared using the floored timestamp. 

This is also not corrected by `Transaction` type in the parent: https://github.com/movementlabsxyz/movement/blob/4947b75ce0f5b2f2927745423ebe8291fd1c41e5/util/movement-types/src/transaction.rs#L76

I think there is a gap in our unit tests that should have caught this. 

I will put in a PR to fix shortly.

protocol-units/sequencing/memseq/sequencer/src/lib.rs
```diff
@@ -444,6 +457,13 @@ pub mod test {
 			Err(anyhow::anyhow!("Mock pop_mempool_transaction"))
 		}
 
+		async fn gc_mempool_transactions(
+			&self,
+			_timestamp_threshold: u64,
+		) -> Result<(), anyhow::Error> {
+			Err(anyhow::anyhow!("Mock gc_mempool_transaction"))
```
l-monninger, line 464: @mzabaluev Can you outline the approach you intend to take here again? That's the most important thing for me to verify currently.

l-monninger (changes requested):

protocol-units/mempool/util/src/lib.rs
mzabaluev: `slot_seconds` is the floored timestamp. But the field is not very useful and with the chosen slot time, it's only a factor of two of the timestamp because the timestamp in `MempoolTransaction` is rounded up to seconds.

protocol-units/sequencing/memseq/sequencer/src/lib.rs
mzabaluev: The GC routine makes a pass of the `mempool_transactions` CF in the key order, stopping at a key prefix that's obtained from the timestamp parameter. All transactions with preceding keys are removed. This works because the transaction timestamp is the leftmost part of the key. In the sequencer, the timestamp cutoff is chosen as twice the block forming period before the current time. The heuristic can be tweaked, of course.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -245,39 +273,71 @@ impl MempoolTransactionOperations for RocksdbMempool {
...
+				.cf_handle(cf::MEMPOOL_TRANSACTIONS)
+				.ok_or_else(|| Error::msg("CF handle not found"))?;
+			let lookups_cf_handle = db
+				.cf_handle(cf::TRANSACTION_LOOKUPS)
+				.ok_or_else(|| Error::msg("CF handle not found"))?;
+			let mut read_options = ReadOptions::default();
+			read_options
+				.set_iterate_upper_bound(construct_timestamp_threshold_key(timestamp_threshold));
+			let mut iter = db.iterator_cf_opt(&cf_handle, read_options, IteratorMode::Start);
+			let mut batch = WriteBatch::default();
```
l-monninger, line 317: So, I think ideally, we would be using [`DeleteRange`](https://rocksdb.org/blog/2018/11/21/delete-range.html) here.

Alternatively, we could build manual index by writing into GC designated column families. But, I don't think this is any better than the above.

protocol-units/mempool/util/src/lib.rs
l-monninger: No, this got garbled over the refactors. You can see we're using the `slot_now` constructor which under the hood is just passing the `Self::SLOT_SECONDS` constant: https://github.com/movementlabsxyz/movement/blob/551687ea77596ad6ada55b59669effcfcb146cea/protocol-units/mempool/util/src/lib.rs#L70

l-monninger: But, this is fixed in #631

l-monninger (changes requested):

protocol-units/mempool/move-rocks/src/lib.rs
mzabaluev: I thought of using that operation, but we also need to remove the "transaction_lookups" entries, and I don't see how we can do it without enumerating the removed transactions first.

l-monninger (approved):

PR #631, opened 2024-09-26
Timestamp Ordering

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Corrects timestamp ordering and adds test. 

# Changelog

# Testing

# Outstanding issues
1. I feel like this was also corrected and tested in another PR. I'm going to check for that to link and make sure we aren't doing some unhigenic things here


protocol-units/mempool/util/src/lib.rs
```diff
@@ -147,12 +147,12 @@ impl PartialOrd for MempoolTransaction {
 impl Ord for MempoolTransaction {
 	fn cmp(&self, other: &Self) -> Ordering {
 		// First, compare by slot_seconds
-		match self.slot_seconds.cmp(&other.slot_seconds) {
```
mzabaluev, line 150: Good!
Is the `slot_seconds` field left with any use whatsoever now?

protocol-units/settlement/mcr/client/Cargo.toml
```diff
@@ -46,8 +50,6 @@ serde = { workspace = true }
 godfig = { workspace = true }
 
 dot-movement = { workspace = true }
-
-[dev-dependencies]
```
mzabaluev, line 50: Why remove this? We don't want test dependencies to pollute the build graph for anything using the mcr client crate.

process-compose/mcr-client/process-compose.test.yml
```diff
@@ -7,7 +7,7 @@ processes:
   mcr-client-tests:
     command: |
       set -e
-      cargo test -p mcr-settlement-client --features="e2e" -- --nocapture
+      cargo run --bin test-mcr-settlement-client
```
mzabaluev, line 10: This commit is unrelated to the timestamps. Was it added to fix the CI for this PR?

mzabaluev (changes requested):

protocol-units/mempool/util/src/lib.rs
l-monninger: Only insofar as we might want to make these structs self-describing.

process-compose/mcr-client/process-compose.test.yml
l-monninger: Yeah, CI fix.

protocol-units/settlement/mcr/client/Cargo.toml
mzabaluev: Ah yes, that's why I don't like tests as binaries. Consider moving them to a separate crate e.g. mcr-settlement-tests, or finding a way to have them as integration tests in the crate `./tests` directory.

protocol-units/mempool/util/src/lib.rs
```diff
@@ -147,12 +147,12 @@ impl PartialOrd for MempoolTransaction {
 impl Ord for MempoolTransaction {
 	fn cmp(&self, other: &Self) -> Ordering {
 		// First, compare by slot_seconds
```
mzabaluev, line 149: This comment is outdated now.
```suggestion
		// First, compare by timestamps
```

protocol-units/mempool/util/src/lib.rs
mzabaluev: OK, I missed `Debug`. Is it just for that?

protocol-units/settlement/mcr/client/Cargo.toml
mzabaluev: Or, perhaps, do add an "integration-tests" feature flag guarding these binaries, like we have it elsewhere, and make these dependencies conditional on that.

protocol-units/mempool/util/src/lib.rs
l-monninger: I wasn't thinking about it just in terms of debug.

protocol-units/settlement/mcr/client/Cargo.toml
0xmovses: I think we should be able to exclude the testing bin from the release build if we, in the root Cargo.toml:
```rust
[profile.release]
exclude = ["test-mcr-settlement-client"]
```

0xmovses (changes requested):
Logic looks good, one suggestion re. slim release binaries.

protocol-units/settlement/mcr/client/Cargo.toml
l-monninger: I'm not familiar with the above syntax. I couldn't find it in the docs either. 

I would assume this also wouldn't change the bin size if it does work, just drop that particular binary from the workspace build. 

I couldn't actually find a good way to do this with profiles, so our best option is indeed to stick with a flag and just make sure everyone is using it. Sound good to you?

0xmovses: Ah, I meant it would not include the binary in the release build. But fine with the flag.

mzabaluev: I'm not sure about profiles, it's a weird way to affect the build graph.

I agree that the feature flag is not the best solution, either.
So my alternative suggestions are:
1) a separate test-only crate, or
2) find a way to build and run it as regular Rust tests.

It's less important that the tests require external setup.

l-monninger: I think we can move this to a separate PR: https://github.com/movementlabsxyz/movement/discussions/643

PR #369, opened 2024-08-16
l-monninger/open-network

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `networks`

A PR for collaborating on modifications, bugs, and usability of Suzuka Syncing APIs and the `l-monninger/open-network` Suzuka Environment.

If you are looking to join this network, following the guides [here](https://github.com/movementlabsxyz/movement/tree/l-monninger/open-network/docs/movement-node-experimental/l-monninger/open-network).

# Changelog

1. Changes to `syncador-v2` to increase usability of syncing API.

# Testing

# Outstanding issues

PR #641, opened 2024-09-30
Elevate Pruning Config

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `misc`.

Adds pruning config parameters for Aptos state  to the Maptos Config. Applies minimal defaults. Opens DB in pruning mode. 

# Testing

1. All tests pass.
2. e2e testing for #369 sync speed shows improvement which reasonably facilitates testnet adoption, i.e., adoption against large states. 

# Outstanding issues
1. Does not prune `suzuka-da-db` as this remains small. 
2. Does not add tests to confirm pruning behavior. These would have to be somewhat long-running; we will have to be cleverer about how we add these.

### DIFF
--- a/protocol-units/execution/util/src/config/common.rs
+++ b/protocol-units/execution/util/src/config/common.rs
@@ -120,17 +120,17 @@ env_default!(
 	10
 );
 
-env_default!(default_maptos_ledger_pruning_window, "MAPTOS_LEDGER_PRUNING_WINDOW", u64, 50_000_000);
+env_default!(default_maptos_ledger_prune_window, "MAPTOS_LEDGER_PRUNING_WINDOW", u64, 50_000_000);
 
 env_default!(
-	default_maptos_state_merkle_pruning_window,
+	default_maptos_state_merkle_prune_window,
 	"MAPTOS_STATE_MERKLE_PRUNING_WINDOW",
 	u64,
 	100_000 // see comment in aptos-core, this essentially needs to exceed TPS
 );
 
 env_default!(
-	default_maptos_epoch_snapshot_pruning_window,
+	default_maptos_epoch_snapshot_prune_window,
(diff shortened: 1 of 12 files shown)


protocol-units/execution/util/src/config/common.rs
```diff
@@ -120,6 +120,22 @@ env_default!(
...
+	"MAPTOS_STATE_MERKLE_PRUNING_WINDOW",
+	u64,
+	100_000 // see comment in aptos-core, this essentially needs to exceed TPS
+);
+
+env_default!(
+	default_maptos_epoch_snapshot_prune_window,
+	"MAPTOS_EPOCH_SNAPSHOT_PRUNING_WINDOW",
+	u64,
+	50_000_000
```
0xmovses, line 136: What does this number represent? transactions? I need a little more context.

0xmovses (approved):
This looks fine. From my understanding by reading the code, this will create additional fields on the config json file that is stored inside `./movement` these additional fields allow for pruning data in the node. 

Could you drop the command for e2e testing as mentioned in the PR description? I'd like to run them for my own knowledge later on.

protocol-units/execution/util/src/config/common.rs
l-monninger: Yeah, `ledger_version`. It's easiest to read the upstream comments if you want me to permalink them.

l-monninger: e2e testing is...

```
just suzuka-full-node native build.setup.eth-local.celestia-local.test
```

0xmovses: ah its the normal just e2e test, I wondered if some other test was added. Ok thanks.

PR #645, opened 2024-10-01
Add state and storage proof rest APIs

aeryz:
# Summary
- **RFCs**: $\emptyset$.

In order for Movement to support IBC and light clients, it should expose some API's for getting transaction, storage and state proofs. This PR adds the necessary API's for supporting IBC. Please see `Changelog` for more details.

# Changelog

The following API's have been added:
1. `/movement/v1/table-item-with-proof/:table_handle/:key/:blockheight`: Returns a table item and it's storage proof at a specific height. This is being used to verify whether some data `N` is really committed. The proof is verified against the [state_checkpoint_hash](https://github.com/aptos-labs/aptos-core/blob/d10449df4f68f9af0f5ee3620963fe5e3bfaf5ac/api/types/src/transaction.rs#L367) of the transaction which is the final transaction at the block at `blockheight`.
2.   `/movement/v1/state-proof/:blockheight`: Returns the state proof that is committed by Movement to L1 and also the transaction proof and info which contains the `state_checkpoint_hash`. The transaction proof is verified against the [executed_state_id](https://github.com/aptos-labs/aptos-core/blob/d10449df4f68f9af0f5ee3620963fe5e3bfaf5ac/types/src/block_info.rs#L35) which is a part of the `StateProof`. Note that in order to verify the validity of the given `StateProof`, the light clients would have to check it against the hash that is committed to Ethereum L1.

# Testing

# Outstanding issues

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -95,6 +95,9 @@
               # rust toolchain
               (toolchain pkgs)
 
+              # for `hostname -I`
+              hostname-debian
+
               # build dependencies
               llvmPackages.bintools openssl openssl.dev libiconv pkg-config
               libclang.lib libz clang pkg-config protobuf rustPlatform.bindgenHook
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9058,8 +9058,12 @@ version = "0.0.2"
 dependencies = [
  "anyhow",
  "aptos-api",
+ "aptos-types",
  "futures",
+ "hex",
  "poem",
+ "serde",
+ "serde_json",
  "tokio",
  "tracing",
 ]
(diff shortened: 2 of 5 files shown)


flake.nix
```diff
@@ -95,6 +95,9 @@
               # rust toolchain
               (toolchain pkgs)
 
+              # for `hostname -I`
+              hostname-debian
```
mzabaluev, line 99: Is this added just for convenience, or is there a need for this in the development shell?

aeryz: required here: https://github.com/movementlabsxyz/movement/blob/main/scripts/movement/run#L23

PR #640, opened 2024-09-30
Feat: Bridge Godfig

andygolay:
# Summary
- **RFCs**: [Link to RFC](#./link/to/rfc), [Link to RFC](#./link/to/rfc), or $\emptyset$.
- **Categories**: any of `protocol-units`, `networks`, `scripts`, `util`, `cicd`, or `misc`.

Integrate the same GodFig configuration as Suzuka node for Bridge Service.

# Changelog
* Add config and setup crate to build the bridge configuration file using GodFig.
* In setup create the step needed to start a local Eth and Movmeent network for the test.
* Move integration test in integration_test crate.
* Update test and TestHarness to use the new Config and Setup.
* Pass the tests.

# Testing
1. `cargo test --test client_movement_eth -- --nocapture --test-threads=1 `
2.`cargo test --test client_eth_movement -- --nocapture --test-threads=1 ` 
3.`cargo test --test bridge_e2e-test -- --nocapture --test-threads=1 ` 

# Outstanding issues

PR #646, opened 2024-10-01
Bridge CLI updates for production environment

0xmovses:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3675,6 +3675,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "bridge-cli"
+version = "0.0.2"
+dependencies = [
+ "alloy",
+ "anyhow",
+ "bridge-service",
+ "clap 4.5.18",
+ "directories",
+ "eyre",
+ "serde",
+ "serde_json",
+ "tokio",
+ "tracing",
+ "tracing-subscriber 0.3.18",
+ "url",
+ "uuid",
+]
+
+[[package]]
 name = "bridge-service"
 version = "0.0.2"
 dependencies = [
@@ -5249,12 +5268,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "directories"
+version = "5.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9a49173b84e034382284f27f1af4dcbbd231ffa358c0fe316541a7337f376a35"
+dependencies = [
+ "dirs-sys 0.4.1",
+]
+
+[[package]]
 name = "dirs"
 version = "3.0.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "30baa043103c9d0c2a57cf537cc2f35623889dc0d405e6c3cccfadbc81c71309"
 dependencies = [
- "dirs-sys",
+ "dirs-sys 0.3.7",
 ]
 
 [[package]]
@@ -5276,6 +5304,18 @@ dependencies = [
  "libc",
  "redox_users",
  "winapi 0.3.9",
+]
+
+[[package]]
+name = "dirs-sys"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "520f05a5cbd335fae5a99ff7a6ab8627577660ee5cfd6a94a6a929b52ff0321c"
+dependencies = [
+ "libc",
+ "option-ext",
+ "redox_users",
+ "windows-sys 0.48.0",
 ]
 
 [[package]]
@@ -9557,6 +9597,12 @@ dependencies = [
  "pkg-config",
  "vcpkg",
 ]
+
+[[package]]
+name = "option-ext"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "04744f49eae99ab78e0d5c0b603ab218f515ea8cfe5a456d7629ad883a3b6e7d"
 
 [[package]]
 name = "ordered-float"
@@ -14034,6 +14080,7 @@ version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "81dfa00651efa65069b0b6b651f4aaa31ba9e3c3ce0137aaad053604ee7e0314"
 dependencies = [
+ "getrandom 0.2.15",
  "serde",
 ]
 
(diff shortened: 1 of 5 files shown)


musitdev (commented):
I don't understand why you need to open a grpc access to the bridge onchain call. The cli don't need it because it can get all these call directly using a chain access, and it adds security risk on the bridge (mostly DDOS attack).
I thought that some part of the bridge client code will be implemented or share in the CLI code.
The smart contract access for the bridge has 3 domains:
 * bridge domain that implement lock_transfer, initiator_completed, refund_transfer
 * the test domain with the get_details call I suppose these methods are only use for test or are specific for the test.
 * CLI/SDK domain with initiate_transfer and counterpart_complete and perhaps some other specific get_details calls.
 All these are implemented by a domain client that call the associated blockchain using its domain related access.
These distinctions are important because the onchain access is different for each domain. For bridge domain we connect to a node managed by us, for the test it's a test node (Anvil or movement) and for the CLI/SDK it's a public access or anything managed by the user.
 So I'm not sure we need a grpc access.

0xmovses:
The CLI is for the bridge operator to do any call, such as refunding and aborting and also starting the bridge service. It runs as a separate process completely to the service so it needs a way to communicate with the ETH and Movement Clients, the best way is via gRPC. 

The CLI is not for users, but for the bridge operator (us).

0xmovses:
Users make calls to the Bridge via the  Front End.

andygolay:
When I run `cargo run -p bridge-cli ` I get:

```
Command line interface to perform atomic bridge transfers

Usage: bridge-cli <COMMAND>

Commands:
  eth-to-movement  Ethereum to Movement bridge commands
  movement-to-eth  Movement to Ethereum bridge commands
  help             Print this message or the help of the given subcommand(s)
```

But then when I run `bridge-cli` or ` ./target-debug/bridge-cli` I get: 

```
zsh: command not found: bridge-cli
```

andygolay:
> This is what we expect. Not sure about running it from the target I didn't test that. Is it the correct path?
> 
> ```
> Command line interface to perform atomic bridge transfers
> 
> Usage: bridge-cli <COMMAND>
> 
> Commands:
>   eth-to-movement  Ethereum to Movement bridge commands
>   movement-to-eth  Movement to Ethereum bridge commands
>   help             Print this message or the help of the given subcommand(s
> ```

I think so (see screenshot). Should the subcommands run in general?
 
 Or should it be expected to return `command not found: bridge-cli`?
 
image

0xmovses:
One thing, I'm going to remove all the calls that the relayer will call automatically, I can't think of any reason why the CLI would need to support them. Let me trim this down to remove bloat and I'll check your issue.

0xmovses:
A command I run with the executable that works: 
```
./target/debug/bridge-cli eth-to-movement initiate "0x8bcdbe40eeb01c7451f359318e5709c16ab2f23c3a9fa71531cca57920aa828c" 100 "2bb80d537b1da3e38bd30361aa855686bde0b2f16f48e5b536b0f7625a529f33"
```

We are missing config still but the args get sent to the client correctly. The client currently does not connect to anvil it. Once https://github.com/movementlabsxyz/movement/pull/640 is in, it will just look to the config to initialise the EthClient and Movement Client. 

I created a README.md with some instructions also.

protocol-units/bridge/cli/src/subcommands.rs
```diff
@@ -0,0 +1,80 @@
...
+	/// Ethereum to Movement bridge commands
+	#[command(subcommand)]
+	EthToMovement(TransferSubcommands),
+
+	/// Movement to Ethereum bridge commands
+	#[command(subcommand)]
+	MovementToEth(TransferSubcommands),
+}
+
+/// The Transfer C
```
andygolay, line 23:
```suggestion
/// Transfer commands
```

protocol-units/bridge/cli/src/subcommands.rs
```diff
@@ -0,0 +1,80 @@
...
+		/// The hash lock for the bridge transfer
+		#[arg(help = "keccak256 hash of the secret")]
+		hash_lock: String,
+	},
+
+	#[command(
+		about = "Completes the transfer by calling complete_bridge_transfer on the destination chain and revealing the preimage"
+	)]
+	Complete {
+		/// The transfer id of the bridge transfer
```
andygolay, line 46:
```suggestion
		/// The transfer ID of the bridge transfer
```

protocol-units/bridge/cli/src/subcommands.rs
```diff
@@ -0,0 +1,80 @@
...
+		about = "Refunds a transfer back to the origin chain. Only callable by the Owner of the contract"
+	)]
+	InitiatorRefund {
+		/// The transfer ID of the bridge transfer
+		#[arg(help = "The transfer id of the bridge transfer")]
+		transfer_id: String,
+	},
+
+	CounterpartyAbort {
+		/// The Transfer ID of the bridge transfer
```
andygolay, line 66:
```suggestion
		/// The transfer ID of the bridge transfer
```

protocol-units/bridge/cli/src/subcommands.rs
```diff
@@ -0,0 +1,80 @@
...
+#[command(name = "Movement Atomic Bridge CLI")]
+#[command(about = "Command line interface to perform atomic bridge transfers", long_about = None)]
+pub struct CliOptions {
+	/// The bridge command to run (Ethereum to Movement or Movement to Ethereum)
+	#[command(subcommand)]
+	pub command: BridgeCommands,
+}
+
+#[derive(Subcommand)]
+pub enum BridgeCommands {
```
andygolay, line 13:
```suggestion
/// Bridge commands
pub enum BridgeCommands {
```

protocol-units/bridge/service/src/chains/ethereum/grpc_handler.rs
```diff
@@ -0,0 +1,158 @@
...
+		Self {
+			eth_client: EthClient::new(Config::build_for_test())
+				.await
+				.expect("Failed to create EthClient"),
+		}
+	}
+}
+
+// More Admin only utility functions can be added to get more information.
+// For example list of ongoing tasks, list of completed tasks, etc.
```
andygolay, line 31: Will the gRPC server be for read-only activity?

Probably good for @musitdev to review the gRPC code, as I'm a bit unfamiliar with its role in the CLI.

andygolay (commented):
Looks solid as a foundation for the CLI! 

Added some comments and questions for clarification.

0xmovses:
Admin CLI can be useful, but draft for now as this is for the HTLC protocol.

PR #663, opened 2024-10-07
Client tests for framework

andygolay:
# Summary
Add client integration tests for testing against the framework modules

# Changelog

- Added `framework` bool arg to client functions
- Added `extract_bridge_transfer_details` to test utils

# Testing

Currently failing to get the event, with this tests

```
rust_backtrace=1 cargo test --test client_l2move_l1move test_movement_client_initiate_transfer -- --nocapture --test-threads=1
```

# Outstanding issues
Still debugging


protocol-units/bridge/service/src/chains/ethereum/client.rs
```diff
@@ -274,6 +276,7 @@ impl crate::chains::bridge_contracts::BridgeContract<EthAddress> for EthClient {
 		&mut self,
 		bridge_transfer_id: BridgeTransferId,
 		pre_image: HashLockPreImage,
+		framework: bool
```
0xmovses, line 279: I know this is still in draft, but I don't think we need framework bool arg here, we don't want to update the traits. When I suggested framework: bool arg, I was thinking of using that in the testing setup code to conditionally call the .move script, was just a loose idea. Feel free to disregard, looks like that arg is not being used anyway in the implementation of the trait methods here.

protocol-units/bridge/integration-tests/tests/client_movement_eth.rs
```diff
@@ -36,6 +36,7 @@ async fn test_movement_client_build_and_fund_accounts() -> Result<(), anyhow::Er
 async fn test_movement_client_initiate_transfer() -> Result<(), anyhow::Error> {
 	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
 
+
```
andygolay, line 39:
```suggestion
```

Issue #555, opened 2024-09-10
Emit OpenTelepathy events for ingress transactions


mzabaluev:
Discussions: #422 and #423.

On each transaction received, currently in maptos-opt-executor's [transaction_pipe](https://github.com/movementlabsxyz/movement/blob/68cffed5e55f46fd29b6ce758ccdc6be0551f342/protocol-units/execution/opt-executor/src/transaction_pipe.rs#L90), emit a distinctively identified OpenTelepathy event tagged with the transaction hash, sender, and sequence number.

[tracing-opentelemetry](https://github.com/tokio-rs/tracing-opentelemetry) can be used to emit the events using tracing macros.
Consider repurposing the tracing span and events currently used for "movement_timing", rather than adding redundant tracing.
The separation of concerns is that the process-global tracing subscriber configuration is responsible for dispatching events of interest, while at the logging site events are tagged to enable the subscriber filtering. 

The events will be aggregated by Grafana Alloy or other suitable agent.

Issue #556, opened 2024-09-10
Emit OpenTelemetry events for transactions sent to DA


mzabaluev:
Discussions: #422 #423

On each transaction sent to DA, currently in the [transaction_ingress task](https://github.com/movementlabsxyz/movement/blob/68cffed5e55f46fd29b6ce758ccdc6be0551f342/networks/suzuka/suzuka-full-node/src/tasks/transaction_ingress.rs#L65) of the node, emit a distinctively identified OpenTelepathy event tagged with the transaction hash, sender, and sequence number.

Consider repurposing the tracing span and events currently used for "movement_timing", rather than adding redundant tracing.
The separation of concerns is that the process-global tracing subscriber configuration is responsible for dispatching events of interest, while at the logging site events are tagged to enable the subscriber filtering. 

The events will be aggregated by Grafana Alloy or other suitable agent.

PR #670, opened 2024-10-08
OpenTelemetry OTLP setup for tracing

mzabaluev:
Resolves: #555 #556

# Summary
- **Categories**: `networks`, `util`.

Integrate an OTLP exporter for tracing events matching the "movement_tracing" target.
The OTLP gRPC endpoint is configured with the `MOVEMENT_OTLP` environment variable.

This replaces the earlier "movement_timing" tracing layer, as analysis of OpenTelemetry should be more versatile.

# Changelog

* Added OpenTelemetry OTLP exporter, with the endpoint URL configured with the `MOVEMENT_OTLP` environment variable.
* Removed support for timing JSON logs configured with `MOVEMENT_TIMING`.

# Testing

TODO: add docker-compose setup for jaeger or similar.

Local process-compose testing hasn't been conclusive.


process-compose/suzuka-full-node/process-compose.telemetry.yml
```diff
@@ -0,0 +1,11 @@
+version: "3"
+
```
l-monninger, line 2: If you can add an OTLP runner, that would be great.

l-monninger: Our nix flake does package in Docker, so you can just spawn the container from process compose.

l-monninger (changes requested):

process-compose/suzuka-full-node/process-compose.telemetry.yml
l-monninger:
```
just suzuka-full-node native build.setup.eth-local.celestia-local.test -t=false
```

protocol-units/execution/opt-executor/src/transaction_pipe.rs
```diff
@@ -123,16 +131,14 @@ impl TransactionPipe {
...
 		let in_flight = self.transactions_in_flight.load(std::sync::atomic::Ordering::Relaxed);
-		info!(
-			target: "movement_timing",
-			in_flight = %in_flight,
-			"transactions_in_flight"
+		let otel_cx = OtelContext::current();
+		let otel_span = otel_cx.span();
+		otel_span.add_event(
+			"transactions_in_flight",
+			vec![KeyValue::new("in_flight", in_flight as i64)],
```
mzabaluev, line 138: This event looks like it wants to be a proper metric.

protocol-units/execution/opt-executor/src/executor/mod.rs
```diff
@@ -46,7 +46,7 @@ impl Executor {
 		self.transactions_in_flight
 			.fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
 				info!(
-					target: "movement_timing",
+					target: "movement_telemetry",
 					count,
 					current,
 					"decrementing_transactions_in_flight",
```
mzabaluev, line 52: Also this one looks like a metric would be better (I haven't converted it to OpenTelemetry yet).

util/tracing/src/telemetry.rs
```diff
@@ -0,0 +1,75 @@
...
+		Ok(Self { otlp_grpc_url })
+	}
+}
+
+/// Global initialization of the OpenTelemetry tracer provider.
+///
+/// This function should be called at the start of the program before any
+/// threads are able to use OpenTelemetry tracers. The function will panic
+/// if not called within a Tokio runtime.
+pub fn init_tracer_provider(
```
l-monninger, line 50: I would prefer these providers are fungible. Ideally, the content of a particularly (set) of environment variables, args, or other would tell an instance of the application which to use. How difficult is this to set up from where we are currently?

protocol-units/da/m1/light-node/src/v1/passthrough.rs
```diff
@@ -146,7 +146,7 @@ impl LightNodeV1 {
 		Ok(verified_blobs)
 	}
 
-	#[tracing::instrument(target = "movement_timing", level = "debug")]
+	#[tracing::instrument(target = "movement_telemetry", level = "debug")]
```
l-monninger, line 149: I would like to have comments on each of these explaining why they are needed to achieve a certain metric as outlined here: 
- https://github.com/movementlabsxyz/movement/discussions/367
- https://github.com/movementlabsxyz/sre/discussions/63

l-monninger (changes requested):

l-monninger:
Will try to test this e2e shortly.

protocol-units/da/m1/light-node/src/v1/passthrough.rs
mzabaluev: I'm not sure if we need all of these for telemetry. Originally they were added for perf research.
Do you think we can just remove those that don't contribute to a particular metric goal, or convert them to a plain spans/events with the default target?

util/tracing/src/telemetry.rs
mzabaluev: As for fungibility in the API, tracing-opentelemetry was supposed to make OpenTelemetry exports just another sort of tracing macro call sites, but it does not work due to https://github.com/tokio-rs/tracing-opentelemetry/issues/159.
This is why I made `telemetry` a separate module with its own API.

mzabaluev:
Superseded by #697.

PR #677, opened 2024-10-10
Client integration tests against framework modules

andygolay:
# Summary
- Update client tests so they run against the Movement framework in our fork of `aptos-core`

# Changelog

- Added `client_l2move_l1move` and  `client_l1move_l2move` files with tests against framework.
- Fix `get_bridge_transfer_details_counterparty` which was returning a struct with initiator and recipient address types reversed. Implemented a `BridgeTransferDetailsCounterparty` struct to achieve this.

# Testing

To run client tests:

1. Navigate to the root of the `movement` repo, checkout the `andygolay/framework-client-tests` branch, and start a local Suzuka node:

```
CELESTIA_LOG_LEVEL=FATAL CARGO_PROFILE=release CARGO_PROFILE_FLAGS=--release nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just bridge native build.setup.eth-local.celestia-local --keep-tui"
```

Note, if you have any `.movement` directories present, e.g. when re-running tests, the directories must be deleted before starting the node. 

2. In the generated `.movement` directory, there will be a `config.yaml`. In that file, change `f90391c81027f03cdea491ed8b36ffaced26b6df208a9b569e5baf2590eb9b16` to `0xA550C18` so the file looks like:

```
---
profiles:
  default:
    network: Custom
    private_key: "0x0000000000000000000000000000000000000000000000000000000000000001"
    public_key: "0x4cb5abf6ad79fbf5abbccafcc269d85cd2651ed4b885b5869f241aedf0a5ba29"
    account: "0xA550C18"
    rest_url: "http://0.0.0.0:30731/"
    faucet_url: "http://0.0.0.0:30732/"
``` 

3. Run each set of Movement client tests:

L1 Move to L2 Move:
```
rust_backtrace=1 cargo test --test client_l1move_l2move -- --nocapture --test-threads=1
```

L2 Move to L1 Move:
```
rust_backtrace=1 cargo test --test client_l2move_l1move -- --nocapture --test-threads=1
```

# Outstanding issues
- The `refund` and `abort` tests have a sleep of 20 seconds before calling `refund` or `abort`. Just a heads up.
- If your node has been running for awhile, L1 --> L2 tests are sometimes flaky at the `extract_bridge_transfer_id_framework` call, where it won't find a matching transaction. I don't know why that happens but restarting the node and re-running steps 1 and 2 of the above test directions appears to fix it so tests pass.
- There was an issue with not finding a coin store for the recipient in `client_l1move_l2_move test_movement_client_complete_transfer`. The fix I implemented for it is rather hacky. That test should be cleaned up but perhaps in a later iteration.


0xmovses:
This looks good, I will do a thorough review once we have the governance on testnet devnet issue resolved for local client tests.

protocol-units/bridge/integration-tests/README.md
```diff
@@ -0,0 +1,128 @@
...
+
+```
+movement move compile --package-dir protocol-units/bridge/move-modules
+```
+
+2. Run the `enable_bridge_feature` script:
+
+```
+ movement move run-script --compiled-script-path protocol-units/bridge/move-modules/build/bridge-modules/bytecode_scripts/enable_bridge_feature.mv --profile local_root
+ ```
```
0xmovses, line 128: Does this script need to be ran manually? I think the integration test handles this? Maybe it's worth adding a note that these steps are for manual testing and they take place automatically in the integration test

protocol-units/bridge/integration-tests/src/lib.rs
```diff
@@ -49,29 +52,42 @@ pub struct MovementToEthCallArgs {
...
+                    .take(6)
+                    .map(char::from)
+                    .collect();
+
+                // Construct the bridge_transfer_id with the random suffix
+                let mut bridge_transfer_id = b"00000000000000000000000tra".to_vec();
+                bridge_transfer_id.extend_from_slice(random_suffix.as_bytes());
+
+                Self {
+                        initiator: b"32Be343B94f860124dC4fEe278FDCBD38C102D88".to_vec(),
```
0xmovses, line 68: Where does this address come from?

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+use tokio::time::{sleep, Duration};
+use tokio::{self};
+use tracing::info;
+
+#[tokio::test]
+async fn test_movement_client_lock_transfer(
+) -> Result<(), anyhow::Error> {
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
+
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
```
0xmovses, line 31: Dup line here, we only need to init `tracing_subscriber` once for a a given context.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+	.await;
+
+	test_result
+}
+
+#[tokio::test]
+async fn test_movement_client_complete_transfer(
+) -> Result<(), anyhow::Error> {
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
```
0xmovses, line 90: Dup line here.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+	let test_result = async {
+		let coin_client = CoinClient::new(&mvt_client_harness.rest_client);
+		let movement_client_signer = mvt_client_harness.movement_client.signer();
+		{
+			let faucet_client = mvt_client_harness.faucet_client.write().unwrap();
+			faucet_client.fund(movement_client_signer.address(), 100_000_000).await?;
+			faucet_client.fund(AccountAddress::from_hex_literal("0xface")?, 100_000_000).await?;
+			faucet_client.fund(AccountAddress::from_hex_literal("0x1")?, 100_000_000).await?;
+			faucet_client.fund(AccountAddress::from_hex_literal("0x3078303030303030303030303030303030303030303030303030303066616365")?, 100_000_000).await?;
+
```
0xmovses, line 104: Why this address? Small comment in code about this `0x307` addy.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+	test_result
+}
+
+#[tokio::test]
+async fn test_movement_client_abort_transfer(
+) -> Result<(), anyhow::Error> {
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
+
+	let _ = tracing_subscriber::fmt().with_max_level(tracing::Level::DEBUG).try_init();
```
0xmovses, line 181: Dup line here.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+	];
+
+	let provider = eth_client_harness.rpc_provider().await;
+	let accounts = provider.get_accounts().await.expect("Failed to get accounts");
+	assert_eq!(accounts.len(), expected_accounts.len());
+
+	for (account, expected) in accounts.iter().zip(expected_accounts.iter()) {
+		assert_eq!(account, expected);
+	}
+}
```
0xmovses, line 294: Remove this `test_eth_client_should_build_and_fetch_accounts` as it exists in the `eth_movement` integration test, we don't want to waste CI compute.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
```diff
@@ -0,0 +1,455 @@
...
+			BridgeAddress(recipient),
+			HashLock(EthHash(hash_lock).0),
+			Amount(AssetType::EthAndWeth((1, 0))), // Eth
+		)
+		.await
+		.expect("Failed to initiate bridge transfer");
+}
+
+#[tokio::test]
+async fn test_eth_client_should_successfully_call_initiate_transfer_only_weth() {
```
0xmovses, line 346: These should be testing the MOVE.sol contracts seems like they are initialised currently with the old Initiator contracts. In other words this it `AtomicBridgeInitiator` -> COIN. Right?

0xmovses (commented):
Thank you for this!

Main thing here is that the `client_l1move_l2move.rs` tests are testing with the old `AtomicBridgeInitiator`/ Counterparty contract I believe. The abis used in the client handler fns are those contracts not `AtomicBridgeInitiatorMOVE`. 

However, this successfully proves that clients are able to to call the framework functions as expected. 

The lack of configurability of ABI on ETH side could cause issues. We assume right now they are the same, which is why this works, but if there were updates to the `AtomicBridgeInitiatorMOVE` contract, the Clients would break, becase they are always using the sol type `AtomicBridgeInitiator`.

Before testing e2e we should add this configurability. I'm happy to add this onto this PR if you like while we switch.

protocol-units/bridge/integration-tests/README.md
andygolay: Thanks, that was before I had the scripts implemented. Updated the README to a simpler version.

protocol-units/bridge/integration-tests/src/lib.rs
andygolay: It's the dummy valid EIP-55 address used in the framework modules.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
andygolay: Removed

0xmovses:
On second, thoughts let me add the configurability in a follow up PR, to reduce the scope creep on this. I will add a fast followup PR after this goes in.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
andygolay: I didn't touch the Eth client tests, I think I had gotten it in my mind that we were using the E2E tests to tests the `.sol` contracts; I was just using the client tests to check the Movement client against the framework.

One thing I notice on `main` is that there aren't Eth tests in ` `client_movement_eth` anymore. Not sure why or how they disappeared.

protocol-units/bridge/integration-tests/src/lib.rs
andygolay: I'll add a comment.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
0xmovses: Don't worry about this, I'm going to make sure the .sol MOVE contrats are supported by the relayer in a seperate PR. Right now it only uses the old ABIs for WETH contracts.

protocol-units/bridge/integration-tests/tests/client_l1move_l2move.rs
0xmovses: Ignore this

andygolay:
> On second thought, let me add the configurability in a follow up PR, to reduce the scope creep on this. I will add a fast followup PR after this goes in.

Sounds good, looking into the diff to see how to fix the E2E test

0xmovses:
I followed the testing steps written above and all the client tests are passing 🥳

0xmovses:
I've triggered all the CI to ensure that bumping aptos doesn't introduce any failure, it shouldn't as its only the framework tests. After those pass will be good to approve this.

0xmovses (approved):

PR #678, opened 2024-10-10
Bridge Eth-> Mvt happy path transfer

musitdev:
# Summary

Maje the Eth -> Mvt transfer work
# Changelog
* Set the e2e Test for  Eth -> Mvt transfer 
* Add timelock to the configuration. Default 48h. Only initialize the Initiator contract.
* Rebuild Eth contract and Update ABI Initiator files. 
* Log the current pending transfers in the bridge every minute.
* Correct a logic issue in the MVT event monitoring. If an event fails during deserialization, it is always fetched. It should be fetched only one time. 
* Implement the todo part of the event verification in the state logic.

# Testing

1.Start the Bridge.
```
```
2. To start the e2e test:
```
DOT_MOVEMENT_PATH="$(pwd)/.movement" cargo test --test bridge_e2e_test test_bridge_transfer_eth_movement_happy_path -- --nocapture --test-threads=1
```
The test uses the bridge config in .movement/bridge.
To succeed the test uses several accounts.
For MVT it uses the config.movement.movement_signer_address to fund the counterpart account.
For Eth it takes the private key at index 2 in the config. testing.eth_well_known_account_private_key for the initiator account. This account must be funded for the transfer amount and the gas for the 2 tx.

# Outstanding issues

* Transfer are not bootstraped from the 2 blockchain. If the bridge crash it lost the current pending transfer state.
* Eth event state for monitoring aren't persisted. When the bridge restart some event sent during the stop can be missed. It always starts at the last block. For Movmeent, it's persisted so it restart at the same seq_number when the bridge was stop. It save its mvt monitoring state in the file .movement/bridge/pullstate.store


protocol-units/bridge/integration-tests/tests/client_movement_eth.rs
```diff
@@ -89,7 +89,7 @@ async fn test_movement_client_initiate_transfer() -> Result<(), anyhow::Error> {
 }
 
 #[tokio::test]
-#[ignore] // Test fail when run with the other test: https://github.com/movementlabsxyz/movement/issues/656
+//#[ignore] // Test fail when run with the other test: https://github.com/movementlabsxyz/movement/issues/656
```
andygolay, line 92: This passes locally now with `rust_backtrace=1 cargo test --test client_movement_eth -- --nocapture --test-threads=1` 👍

andygolay (approved):
:shipit:

0xmovses (approved):
Looks good, and test is passing for me.

PR #679, opened 2024-10-10
fix: move erc20 transferFrom, update tests

0xmovses:
# Summary
The solidity MOVE initiator contract was missing `transferFrom` call on the initiator side, these are present in the WETH contracts, so probably why we missed it.

# Changelog
- Adds the missing call
- Adds missing unit test logic

# Testing
CI solidity-tests check should pass

# Outstanding issues


andygolay (approved):
:shipit:

PR #683, opened 2024-10-10
Remove Sync Task Participation from Followers

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Cherry-picks from #660 to remove followers from sync task participation. This is a hotfix to reduce the chance for forks during reboot after network asynchrony and to reduce warn logging for third-party operators. 

# Testing

1. Native e2e tested against dynamic environment and the Movement Testnet. 

# Outstanding issues
1. Ideally, we would also at unit or e2e tests confirming the syncing is not taking place. But, this should be fairly trivial.


0xPrimata (approved):
CI passes but I'm unable to reproduce.

PR #688, opened 2024-10-11
Create Bridge readme with run steps

0xmovses:
# Summary
Creates a readme describing how to run the bridge and manual steps to test it outside of the e2e tests. This allows manual testing without the need of the frontend UI.

# Changelog
- Creates a README.md for `protocol-units/bridge`.

# Testing
Try out the commands :)

# Outstanding issues
New contract deployments will happen, README's should be updated.

### DIFF
--- a/protocol-units/bridge/README.md
+++ b/protocol-units/bridge/README.md
@@ -1,0 +1,120 @@
+# Current Contract Deployments
+## Holesky Ethereum Deployments
+
+| Contract                         | Address                                      |
+|-----------------------------------|----------------------------------------------|
+| `AtomicBridgeInitiatorMove.sol`   | 0xb33d9fA868054FD3F8F02ba9Fc04B2a1862Ab855   |
+| `AtomicBridgeCounterpartyMove.sol`    | 0xA6bB0d5eA03f3bEe3510EA3Ec39e85Cd7395Bc37   |
+
+
+## testnet.movementlabs Deployments
+the movement atomic-bridge and all of its modules are deployed under the same package:
+`0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c`
+
+Verify this by running:
+```
+ movement account list --account 0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c 
+```
+## Running the Atomic Bridge Relayer In Production
+The atomic bridge relayer services requires a `config.json` at the root of the application inside a `./movement`
+
+Set these values to your needs to configure the relayer.
+
+```json
+{
+  "eth": {
+    "eth_rpc_connection_protocol": "http",
+    "eth_rpc_connection_hostname": "0.0.0.0",
+    "eth_rpc_connection_port": 8090,
+    "eth_ws_connection_protocol": "ws",
+    "eth_ws_connection_hostname": "0.0.0.0",
+    "eth_ws_connection_port": 8090,
+    "eth_chain_id": 3073,
+    "eth_initiator_contract": "0xb33d9fA868054FD3F8F02ba9Fc04B2a1862Ab855",
+    "eth_counterparty_contract": "0xA6bB0d5eA03f3bEe3510EA3Ec39e85Cd7395Bc37",
+    "eth_weth_contract": "0xB7f8BC63BbcaD18155201308C8f3540b07f84F5e",
+    "signer_private_key": "<ETH_PRIVATE_KEY>",
+    "time_lock_secs": 60,
+    "gas_limit": 10000000000000000,
+    "transaction_send_retries": 10
+  },
+  "movement": {
+    "movement_signer_address": "<MOVEMENT_PRIVATE_KEY>",
+    "movement_native_address": "<MOVEMENT_PUBLIC_KEY>",
+    "mvt_rpc_connection_protocol": "http",
+    "mvt_rpc_connection_hostname": "0.0.0.0",
+    "mvt_rpc_connection_port": 30731,
+    "mvt_faucet_connection_protocol": "http",
+    "mvt_faucet_connection_hostname": "0.0.0.0",
+    "mvt_faucet_connection_port": 30732,
+    "mvt_init_network": "custom"
+  },
+  "testing": {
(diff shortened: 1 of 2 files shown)


protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+## Holesky Ethereum Deployments
+
+| Contract                         | Address                                      |
+|-----------------------------------|----------------------------------------------|
+| `AtomicBridgeInitiatorMove.sol`   | 0xb33d9fA868054FD3F8F02ba9Fc04B2a1862Ab855   |
+| `AtomicBridgeCounterpartyMove.sol`    | 0xA6bB0d5eA03f3bEe3510EA3Ec39e85Cd7395Bc37   |
+
+
+## testnet.movementlabs Deployments
+the movement atomic-bridge and all of its modules are deployed under the same package:
```
andygolay, line 11:
```suggestion
the movement atomic-bridge and all of its modules are deployed under the same package address:
```

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+`0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c`
+
+Verify this by running:
+```
+ movement account list --account 0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c 
+```
```
andygolay, line 17: Maybe here add something like:

Note: this step requires [Movement CLI](https://docs.movementnetwork.xyz/devs/movementcli#install-movement-cli) initialized to the following endpoints. 

You can do this by running `movement init --skip-faucet`. When prompted, enter `custom` network and the following REST endpoint:
```
https://testnet.movementnetwork.xyz
```

Your `config.yaml` should look like:

```
profiles:
  default:
    network: Custom
    private_key: "<your-private-key>"
    public_key: "<your-public-key>"
    account: "<your-account-address>"
    rest_url: "https://testnet.movementnetwork.xyz/"
```

To fund your account, visit the [faucet UI](https://faucet.movementnetwork.xyz/?network=movement) at and select Movement.

andygolay (changes requested):
Seems to be some issue finding that account you deployed to earlier. Left comment in Slack.

andygolay (approved):
:shipit: once network is defo up and stable

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+## testnet.movementlabs Deployments
+the movement atomic-bridge and all of its modules are deployed under the same package address:
+`0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c`
+
+Verify this by running:
+```
+ movement account list --account 0xf7c46960ef7aa1b25cd5d48efef064525aa4ccec286fc3b3a7176bbfec09b84c 
+```
+## Running the Atomic Bridge Relayer In Production
+The atomic bridge relayer services requires a `config.json` at the root of the application inside a `./movement`
```
musitdev, line 19: The path to the bridge config file is `.movement/bridge`.  The config file in .movement is the suzuka node one. I made 2 separate config files because I didn't find how to mix both without mixing the Rust struct.

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+    "mvt_rpc_connection_protocol": "http",
+    "mvt_rpc_connection_hostname": "0.0.0.0",
+    "mvt_rpc_connection_port": 30731,
+    "mvt_faucet_connection_protocol": "http",
+    "mvt_faucet_connection_hostname": "0.0.0.0",
+    "mvt_faucet_connection_port": 30732,
+    "mvt_init_network": "custom"
+  },
+  "testing": {
+    "eth_well_known_account_private_keys": [
```
musitdev, line 53: The `eth_well_known_account_private_keys` field must be empty. It's only use for testing purpose. You can put an empty vec.

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+
+For example:
+```
+aptos move call \
+  --function 0x1::BridgeModule::complete_bridge_transfer \
+  --type-args "" \
+  --args 0x1234 676d6f7665 \
+  --profile default
+```
+You'll notice that the args in the module are of type `vec<u8>`, so you have to correctly convert any string into its hex representation.
```
musitdev, line 112: Not clear. The arg is a 0x string, not a `Vec<u8>`. I would say that to fill the args parameters, the preimage that is a 32 bytes array should be encoded in a 0x string using `xxd` for example.

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+    ]
+  }
+}
+
+```
+Once the relayer is up and listening for events, you can initiate a transfer.
+
+## Bridging MOVE from Ethereum to Movement without a UI
+
+First we need to create the `preimage` or secret, for the transfer.
```
musitdev, line 66: You can add: This preimage is an arbitrary array of 32 bytes. For each transfer, the premiage should be regenerated and be unique.
to describe what is a preimage.

protocol-units/bridge/README.md
```diff
@@ -0,0 +1,120 @@
...
+- <YOUR_ADDRESS>: Your wallet address, which will be the originator of the transaction.
+- <RPC_URL>: The RPC URL of your Ethereum node or test network.
+
+For example:
+```
+cast send 0xYourContractAddress "initiateBridgeTransfer(uint256,bytes32,bytes32)" 1000000000000000000 0x000000000000000000000000000000000000move 0x1234567890abcdef1234567890abcdef1234567890abcdef1234567890abcdef --from 0xYourWalletAddress --rpc-url http://127.0.0.1:8545
+```
+The relayer will now call functions on the movement atomic bridge contract. You can inspect the logs of the service to see its progress. 
+
+Copy the `BridgeTransferId` in the logs.
```
musitdev, line 93: It's not very clear how you get the bridge transfer Id from the logs. Perhaps you should add which part of the logs to copy.

0xmovses:
Unneeded for now.

Issue #559, opened 2024-09-10
Emit OpenTelemetry events for discarded transactions


mzabaluev:
Discussion: #428

Emit a distinctively identified OpenTelepathy event tagged with the transaction hash, sender, and sequence number, on each of the transaction rejection points in suzuka-full-node, including at least:

* Load shedding in the transaction pipe task;
* Pre-validation;
* Aptos mempool rejection;
* Execution (on transaction status).

PR #697, opened 2024-10-15
OpenTelemetry OTLP setup for tracing, take 2

mzabaluev:
Resolves: #555 #556 #559

# Summary
- **Categories**: `protocol-units`, `networks`, `util`.

Integrate an OTLP exporter for tracing events matching the "movement_tracing" target.
The OTLP gRPC endpoint is configured with the MOVEMENT_OTLP environment variable.

This replaces the earlier "movement_timing" tracing layer, as analysis of OpenTelemetry data should be more versatile.

# Changelog

* Added OpenTelemetry OTLP exporter, with the endpoint URL configured with the `MOVEMENT_OTLP` environment variable.
* Removed support for timing JSON logs configured with `MOVEMENT_TIMING`.

# Testing

The `telemetry` overlay for process-compose runs the jaeger all-in-one Docker container as a local OpenTelemetry collector.


util/tracing/src/tracing.rs
```diff
@@ -0,0 +1,27 @@
+use crate::telemetry::{self, ScopeGuard};
+use crate::Config;
+use tracing_subscriber::filter::{EnvFilter, LevelFilter};
+use tracing_subscriber::prelude::*;
+use tracing_subscriber::{fmt, Registry};
+
+/// Sets up the tracing subscribers for a Movement process. This should be
+/// called at the beginning of a process' `main` function.
+///
+/// If successful, returns a guard object that should be dropped at the end
```
l-monninger, line 10: Explain what the guard guards and what is released when it is dropped.

process-compose/suzuka-full-node/process-compose.telemetry.yml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+
+environment:
+
+processes:
+  otlp-collector:
```
l-monninger, line 6: I don't seem to be getting anything served by this and the e2e tests for the basic simulation are failing. Is this how I should be using this?:

```
just suzuka-full-node native build.setup.eth-local.celestia-local.test.telemetry --keep-project
```

l-monninger (changes requested):

process-compose/suzuka-full-node/process-compose.telemetry.yml
l-monninger: Screenshot 2024-10-18 at 10 49 29 AM

This is all I'm seeing from the collector process when I start with the above.

l-monninger: I added a `feed` overlay in case for some reason this is not in fact staying open with the `--keep-project` flag. 

```
just suzuka-full-node native build.setup.eth-local.celestia-local.feed.telemetry --keep-project
```

l-monninger:
@mzabaluev Last commit adds a test. It would be good to add to CI if you like it.

mzabaluev:
> @mzabaluev Last commit adds a test. It would be good to add to CI if you like it.

I have added the overlays to the local test job, as we need some tests to drive the exporting. Do you think it would be better to test separately?

Cargo.toml
```diff
@@ -248,6 +248,10 @@ move-vm-ext = { path = "types/move-vm-ext" }
 num-derive = "0.4.2"
 num-traits = "0.2.14"
 once_cell = "1.8.0"
+opentelemetry = { version = "0.26" }
```
andygolay, line 251: @mzabaluev can this PR be merged soon?

I'm adding OpenTelemetry to the bridge so I'll need the same deps.

mzabaluev: Let me resolve the conflicts with latest main and we can bump it up for review.

elliottdehn:
Make sure that downstream tools and dashboards (or any filtering logic) are updated to reflect the change from "movement_timing" to "movement_telemetry" naming.

elliottdehn:
Can we document logging/tracing practices and how-to in the public repo somewhere? Or at least in internal Notion. Just so people know how to do it, best practices, examples, etc.

0xmovses:
Might be good to commit a README.md to this PR outlining the architecture and implementation, with assistance from Mikhail. Good to start with some solid docs.

elliottdehn:
I'd like us to be able to keep track of:
- Which contracts are calling other contracts (kind of like recording a micro-service graph)
- Perhaps which functions are calling which functions, and if we can swing it (recording a call graph)
- The exclusive/inclusive cost of a function (gas of code within the function, vs. gas of code within the function + transitive calls).
This way we can actually understand and potentially optimize for the workload that we observe over time, and better see what optimization techniques at the node level might yield the best results for our specific use case. Much less speculation. There may even be cases where we can approach a specific team with optimization notes from our observations.

Is that possible to do/demonstrate in this PR, or should it be kicked down the road into another PR? This kind of data was existential at a previous company I worked at.

networks/movement/movement-full-node/src/node/tasks/execute_settle.rs
```diff
@@ -166,7 +168,7 @@ where
 				}
 			}
 		} else {
-			info!(block_id = ?block_id, "Skipping settlement");
+			info!(block_id = block_id_hex, "Skipping settlement");
```
elliottdehn, line 171: `target: "movement_telemetry"` is missing here, is that intentional?

networks/movement/movement-full-node/src/node/tasks/execute_settle.rs
```diff
@@ -110,9 +110,10 @@ where
 			}
 		};
 
+		let block_id_hex = hex::encode(&block_id);
 		info!(
-			block_id = %hex::encode(block_id.clone()),
-			da_height = da_height,
+			block_id = block_id_hex,
```
elliottdehn, line 115: `target: "movement_telemetry"` is missing here too. What's the logic for including/not including?

l-monninger:
@Icarus131 @mzabaluev what happens to this PR now?

mzabaluev:
This has been superseded by #1087.

PR #699, opened 2024-10-16
feat: test_eth_client_lock_then_complete_transfer passes?

andygolay:
# Summary
I tinkered a bit with the contract, but don't think I made any actual change... and this appears to pass now?

```
RUST_BACKTRACE=1 cargo test --test client_eth_movement test_eth_client_lock_then_complete_transfer -- --nocapture --test-threads=1
```

using local bridge service and node started with 

```
 ```

# Changelog

# Testing

# Outstanding issues

### CODE REVIEW

File: protocol-units/bridge/contracts/src/AtomicBridgeCounterparty.sol [hunk 1]
```diff
@@ -53,16 +53,16 @@ contract AtomicBridgeCounterparty is IAtomicBridgeCounterparty, OwnableUpgradeab
         bytes32 hashLock,
         address recipient,
         uint256 amount
-    ) external onlyOwner returns (bool) {
+    ) external returns (bool) {
```
0xmovses, line 56: ah but this needs to be `onlyOwner`

0xmovses (changes requested):

protocol-units/bridge/contracts/src/AtomicBridgeCounterparty.sol
andygolay: It's passing for me without `onlyOwner` ... so it seems like that's the issue, tests must not be satisfying that requirement as far as I can tell. Will close this. Thanks for looking into it!

0xmovses: I did add a trace to check the owner by calling `.owner()` on the contract and it seemed to be the correct address 🤔

PR #700, opened 2024-10-16
Implement relayer contract switch on ETH side calls 

0xmovses:
# Summary
Adds support for both `AtomicBridgeInitiator` (WETH) and `AtomicBridgeInitiatorMOVE` via the use of config field `asset` which is set by the env var `ASSET` managed by the config. 

I realised, that even though the function signatures may be the same for these contract, because of their different names the solidity bytecode will indeed be different, therefore we have to explicitly call one of the two contracts when calling.

# Changelog
- Sets the default config `asset` to be `MOVE`
- Adds the missing ABIs
- Updates EthClient to support match calls depending on value of `AssetKind`
- Optimisation: updated the client to not initialise the contract on every call but rather use the already initialised contracts on the `EthClient` fields.

# Testing
Integration tests are all passing, but do run them again. 

# Outstanding issues
I did try and use generics but I couldn't get the alloy trait bounds quite right, rather than wasting time I thought enums would be faster to implement. We may want to refactor to use generics later instead, but it makes no difference to the operation of the clients. 

We may want to update the movement side to have a similar type of switch, but I'm not sure it will work out the same way because of the steps required, so on movement side, we just hardcode to framework for now


musitdev (commented):
I would have made 2 clients implementation one per token. 
Only one impl is used per Bridge, so you can define the client to use during the start of the bridge. if you need to change one and not the other it can be cumbersome because both impl will always be mixed.
To do that you need to move the sol! contract import in each client instead of putting them in utils too.
I think it's better that we define one client per contract implementation.

0xmovses:
@musitdev I think that would result in a larger diff for the same functionality. This works tests are passing. We could get many more contracts, potentially, making a new client for each contract seems verbose and clunky. A single client should be generalised and flexible enough to deal with any number of asset types.

0xmovses:
The ultimate goal would be to have one contract that could support many asset types, then the client would always call just a single contract, but for now we need this to ensure either contract is correctly called.

andygolay:
Regarding the one client for multiple asset types, I could see that being a viable path, if it's one client per chain, and then there can be a contract on that chain for the different asset types.

All client integration tests pass except one which I put in an issue because I believe it was a bug introduced in 677: https://github.com/movementlabsxyz/movement/issues/701 

I do think @musitdev has a good point about organization, just not sure how many different assets we would need to accommodate per chain? It may be more convenient to put them all in one contract on each chain?

0xmovses:
One client per chain yes. Yes a single contract would be more idea, but for now we have two have contracts. Until these are unified we need to support both. About how many assets it's kind of not related to the code we have here, right now we have two contracts on eth and want to support both, that's what this PR does.

I need to make one additional change before this goes in.

musitdev:
I think the client granularity is Initiator/counterpart contract pair. There should be one client for these pairs. It's a question of lifecycle of the code. The client lifecycle is the same as the one of these 2 contracts. We only change the client if the contracts change. It's important to respect code lifecycle to avoid too complex update. For example, if tomorrow we add batch locking and decide that as one contract is going to be dropped, with 2 clients, we can decide how to implement this update and define a specific strategy for each. With one client, you'll implement both strategies in a mixed code that can become hard to do.
Domain definition should be study and elaborate after some evaluation of the interactions between the party. Lifecycle is a good way to define what is linked or not and define the boundary.
If you keep both contract pair in one client, you make a constraint on both to evolve the same wait.
About multi asset, currently the client interface only allow mono asset. To allow multi asset, we need to alter it or add a new interface. Here we can see another time why the domain boundary is Initiator/counterpart contract pair and not blockchain or asset. If both contract have their own client, you can evolve one and add multi asset and not the others.
It's not a big deal now to have one client, but little by little you'll add more code to manage both in the same client. Even now, you need to add some enum to manage both. With 2 clients, they don't exist.
So do as you prefer. If you keep like that, will see during the next evolution how it behaves. We still can split it latter.

0xmovses:
Not sure how 
> With one client, you'll implement both strategies in a mixed code that can become hard to do.
Your just dealing with one match arm at a time, it seems the same thing but two different approaches. You could have a different strategy for each match arm or case. Seems reasonable. 

Anyway this is not a big deal it just will enable our e2e tests to run both for the current contracts and the framework contracts. The objective of the PR was to allow flexibility for our e2e.

0xmovses:
I don't mind which we way go, but we need to e2e test the framework today and this supports that. We can refine the domains later, if indeed we find ourselves needing to.

andygolay (approved):
It sounds like @0xmovses and @musitdev agree that we can potentially refactor this in the next iteration. Approving with the understanding that further discussion is needed for long-term design.

PR #705, opened 2024-10-17
Add REST health endpoint and configure gRPC for Bridge 

0xmovses:
# Summary
Adds REST health endpoints and configures `EthClient` to accept gRPC requests. 

# Changelog
- Adds `rest.rs`
- Adds `grpc.rs`
- Adds health calls for both
- Updates config
- Adds `grpc/` package for `bridge/`
- Creates a `proto` definition
- Imports necessary tonic crates with features
- Adds simple test for REST health endpoint
- Adds test for gRPC health 

# Testing
Adds the new test :
`cargo test --test bridge_rest`
`cargo test --test bridge_grpc`


protocol-units/bridge/config/src/common/movement.rs
```diff
@@ -113,15 +160,24 @@ impl MovementConfig {
 
 	pub fn suzuka() -> Self {
 		MovementConfig {
-			movement_signer_key: Ed25519PrivateKey::from_encoded_string("0x0000000000000000000000000000000000000000000000000000000000000001").unwrap(),
-			movement_native_address: "0xf90391c81027f03cdea491ed8b36ffaced26b6df208a9b569e5baf2590eb9b16".to_string(),
+			movement_signer_key: Ed25519PrivateKey::from_encoded_string(
+				"0x0000000000000000000000000000000000000000000000000000000000000001",
```
l-monninger, line 164: Just want to get some implications right here. The type here should indeed be a signing key, right? Not the framework address or the genesis signer address for test (which is also not 0x1 it's 0xa12...).

protocol-units/bridge/integration-tests/tests/bridge_rest.rs
```diff
@@ -0,0 +1,29 @@
...
+			EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+		)
+		.init();
+
+	let mock_config = Config::default();
+	let rest_service = Arc::new(BridgeRest::new(&mock_config.movement).unwrap());
+	let rest_service_for_task = Arc::clone(&rest_service);
+
+	let rest_service_future = tokio::spawn(async move {
+		rest_service_for_task.run_service().await.unwrap();
```
l-monninger, line 20: You know my comments about unwraps in tests, right? I would prefer to use results.

protocol-units/bridge/integration-tests/tests/bridge_rest.rs
andygolay, line 1: The health endpoint test passes. Is there a test for the gRPC configuration? I'm not sure that I'm qualified to evaluate gRPC config. Maybe better for @l-monninger or @musitdev to take a look at that part.

protocol-units/bridge/service/src/chains/ethereum/client.rs
```diff
@@ -244,6 +266,24 @@ impl EthClient {
 	}
 }
 
+/// Implement the gRPC service for the Ethereum Bridge Client
+#[tonic::async_trait]
+impl Bridge for EthClient {
+	async fn get_bridge_transfer_details_initiator(
+		&self,
+		_request: Request<GetBridgeTransferDetailsRequest>,
```
l-monninger, line 274: Okay, yeah, leaving this for future implementation is good.

protocol-units/bridge/integration-tests/tests/bridge_rest.rs
0xmovses: Let me add a simple test for the gRPC server also. A health check will probably only be used by REST, but I will add it for testing anyway.

protocol-units/bridge/service/src/rest.rs
```diff
@@ -0,0 +1,48 @@
...
+use poem::{
+	get, handler, listener::TcpListener, middleware::Tracing, EndpointExt, IntoResponse, Response,
+	Route, Server,
+};
+use std::future::Future;
+use std::sync::Arc;
+use tracing::info;
+
+#[derive(Debug)]
+pub struct BridgeRest {
```
l-monninger, line 14: We may want to generalize this pattern going forward. With tonic, you can directly do GRPC over HTTP1. But, it's still a little awkward for nullaries, like health checks, because a message type is still expected. 

gRPC on the other hand has a health standard that we might want to adopt: https://github.com/grpc/grpc/blob/master/doc/health-checking.md

l-monninger (changes requested):
I think you should add calls to the health check in Docker and process compose.

l-monninger:
> I think you should add calls to the health check in Docker and process compose.

Lmk if you'd prefer I just commit these updates.

protocol-units/bridge/service/src/rest.rs
0xmovses: I'll use this and add in a test, I think REST health is still useful for quick checking also. So health on both.

protocol-units/bridge/integration-tests/tests/bridge_rest.rs
0xmovses: Yes I have seen in prior reviews (still I left them here), ok will update.

protocol-units/bridge/config/src/common/movement.rs
0xmovses: This method `suzuka()` is for testing the `bridge-service` e2e against a full running suzuka node locally. It only gets called when running the e2e test. This needs to be set to this value as this is the value of `MAPTOS_PRIVATE_KEY` when running suzuka locally. 

It may be beneficial to rename this method to `suzuka_e2e_test`.

andygolay: Yeah, agreed because it's only used in testing, renaming it `suzuka_for_testing` would be good. It's used in client tests against the framework as well.

The default signer key is `0x1` when starting a local Suzuka node. Default `config.yaml` generated is:

```
---
profiles:
  default:
    network: Custom
    private_key: "0x0000000000000000000000000000000000000000000000000000000000000001"
    public_key: "0x4cb5abf6ad79fbf5abbccafcc269d85cd2651ed4b885b5869f241aedf0a5ba29"
    account: f90391c81027f03cdea491ed8b36ffaced26b6df208a9b569e5baf2590eb9b16
    rest_url: "http://0.0.0.0:30731/"
    faucet_url: "http://0.0.0.0:30732/"
```

protocol-units/bridge/integration-tests/tests/bridge_rest.rs
0xmovses: Added

0xmovses:
Unsure exactly what you have in mind for the process-compose and docker health calls, feel free to commit directly. All other changes requested have been added.

l-monninger:
> Unsure exactly what you have in mind for the process-compose and docker health calls, feel free to commit directly. All other changes requested have been added.

I will place these in a separate PR.

l-monninger (approved):

PR #717, opened 2024-10-20
Ledger Increase Check

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**:`util`

Adds an overlay that will crash the `docker-compose` application if the ledger has not increased in a given amount of time. This is useful for detecting node asynchrony and similar conditions without a more advanced monitoring setup. 

Most of the changes here are permissions changes on compose files. 

# Testing

## To validate
```
# to ensure that it will run normally
LEDGER_INCREASE_PERIOD_SECONDS=300 just suzuka-full-node docker-compose local.ledger-should-progresst will run normally

# to ensure the crash occurs
LEDGER_INCREASE_PERIOD_SECONDS=1 just suzuka-full-node docker-compose local.ledger-should-progress
```

## Added
No automated tests have yet been added. 

# Outstanding issues
1. We may want to build a monitoring util container. This is a quick solution. More advanced solutions would fall [under](https://github.com/movementlabsxyz/MIP/pull/26). 
2. Main still has issues with containers--owing to a bad setup script. Please use an older container when testing.

PR #726, opened 2024-10-21
Testnet Transaction Health

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `networks`

A simple implementation of a health check that uses the existing containers, towards [MIP-26](https://github.com/movementlabsxyz/MIP/pull/26). 

Like #717 , this check will crash the docker-compose application on the whole if the test criteria aren't meant, thereby generating an outage that will be reported for monitoring.

A slightly more sophisticated approach might be to write the exit code of the looped check into the `DOT_MOVEMENT_PATH` at a well known file and serve this status in the watcher. 

A more sophisticated approach still would adjust the running modes of the `suzuka-client-e2e-simple-interaction` to include a `healthcheck` mode which would be run to serve this through the docker compose application. When we transition to  `k8s` and `helm` throughout our infra this will be needed.

### DIFF
--- a/docker/compose/suzuka-full-node/docker-compose.testnet-healthcheck.yml
+++ b/docker/compose/suzuka-full-node/docker-compose.testnet-healthcheck.yml
@@ -1,0 +1,39 @@
+services:
+
+  # runs the simple interaction client to healthcheck the full node over and over
+  suzuka-client-e2e-simple-interaction-healthcheck:
+    image: ghcr.io/movementlabsxyz/suzuka-client-e2e-simple-interaction:${CONTAINER_REV}
+    container_name: suzuka-client-e2e-simple-interaction
+    command: >
+      sh -c '
+        while true; do
+          /app/suzuka-client-e2e-simple-interaction
+        done
+      '
+    environment:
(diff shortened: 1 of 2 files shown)


docker/compose/suzuka-full-node/docker-compose.testnet-healthcheck.yml
```diff
@@ -0,0 +1,39 @@
...
+    container_name: ledger-should-progress-watcher
+    command: >
+      sh -c '
+        while true; do
+          sleep 1000
+        done
+      '
+    depends_on:
+      suzuka-client-e2e-simple-interaction-healthcheck:
+        condition: service_healthy
```
mzabaluev, line 39: I don't understand the purpose of this "service". Is it a stub do add the ledger-should-progress script later?

l-monninger: This is "Induced Crash" above.

andygolay:
@radupopa369 this is the PR I had mentioned to you... I wasn't sure whether your monitoring already implements similar health checks, sending in transactions that simulate user interactions?

PR #714, opened 2024-10-18
Andygolay/bridge e2e test framework

andygolay:
# Summary
Currently working off `62c33771d78524ddf71dc7308f5dce0588602077` rev of our Aptos Core fork.

# Testing

1. Run the local Suzuka node and bridge:

```
CELESTIA_LOG_LEVEL=FATAL CARGO_PROFILE=release CARGO_PROFILE_FLAGS=--release  nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just bridge native build.setup.eth-local.celestia-local.bridge --keep-tui"
```

2. In `.movement/config.yaml` change the `account` field from
```
f90391c81027f03cdea491ed8b36ffaced26b6df208a9b569e5baf2590eb9b16
```
to 
```
0xA550C18
```
which allows use of the core resource account in bridge setup scripts.

The bridge operator is set to `f90391c81027f03cdea491ed8b36ffaced26b6df208a9b569e5baf2590eb9b16` in the setup, because that's the address derived from the signer key set up in Genesis.

3. Run the Eth -> Movement E2E test against the framework. 
```
DOT_MOVEMENT_PATH="$(pwd)/.movement" cargo test --test bridge_e2e_test_framework test_bridge_transfer_eth_movement_happy_path -- --nocapture --test-threads=1
```
# Outstanding issues
Still need Movement -> Eth direction E2E tests. 

Also we hard-coded `FRAMEWORK_ADDRESS` in some places, e.g. `counterparty_complete_bridge_transfer` that may be a bit of a hack.

PR #742, opened 2024-10-24
Bridge relayer, correct asset conversion and add better trace

musitdev:
# Summary

Do some corrections to the relayer for Mvt -> Eth transfer direction.
Set some better trace for the bridge relayer execution.

# Changelog

# Testing
To test follow the process in  this PR: #714 

# Outstanding issues

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,163 @@
+{
+  "nodes": {
+    "crane": {
+      "locked": {
+        "lastModified": 1729741221,
+        "narHash": "sha256-8AHZZXs1lFkERfBY0C8cZGElSo33D/et7NKEpLRmvzo=",
+        "owner": "ipetkov",
+        "repo": "crane",
+        "rev": "f235b656ee5b2bfd6d94c3bfd67896a575d4a6ed",
+        "type": "github"
+      },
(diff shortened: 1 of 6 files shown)


andygolay (dismissed):
LGTM 

```
DOT_MOVEMENT_PATH="$(pwd)/.movement" cargo test --test bridge_e2e_test_framework test_bridge_transfer_eth_movement_happy_path -- --nocapture --test-threads=1
```
passes for me locally after setting the default account to `A550C18` in `config.yaml`

andygolay:
The client integration tests `client_l1move_l2move.rs` and `client_l1move_l2move.rs` pass.

Integration tests against the old FA modules fail (`client_eth_movement` and `client_movement_eth`) due to a linker error. I raised an issue about it here:

https://github.com/movementlabsxyz/movement/issues/746

But think this is good to approve for now unless we're planning to use the FA modules in the near future.

andygolay (approved):

0xmovses:
As agreed with @musitdev I wont be merging this as we will instead be removing `AssetType` entirely as the original usecase for this type no longer exists and we will no longer have to deal with this conversion.

PR #744, opened 2024-10-24
update: docker eth and config

0xmovses:
# Summary
We'd like to try and see if Infura provided RPC endpoints allows us to pull events via RPC, in order to this we need to concatenate more values to construct the RPC connection url. This adds the logic to the config, without breaking the old way

# Changelog
- Updates docker ile
- Updates config

# Testing
Will test on the live network when we redeploy 

# Outstanding issues


andygolay:
Mind adding a quick summary to the PR description?

docker/compose/bridge/docker-compose.eth-holesky.yml
andygolay, line 1: Do we want to change `eth-holesky.yml` to `eth-sepolia.yml` in the file name, if it will be configured for Sepolia instead of Holesky?

0xmovses: We may endup switching back to Holesky as it wasn't Sepolia that was the problem, as we just discovered.

musitdev (approved):

docker/compose/bridge/docker-compose.eth-holesky.yml
```diff
@@ -1,15 +1,18 @@
-services:
+ services:
 
   setup:
     environment:
       ETH_RPC_CONNECTION_PROTOCOL: "https"
-      ETH_RPC_CONNECTION_HOSTNAME: "ethereum-holesky-rpc.publicnode.com"
+      ETH_RPC_CONNECTION_BASE_HOSTNAME: "sepolia.infura.io"
```
mzabaluev, line 6: Should we rename the overlay to `sepolia`, then?

0xmovses:
Thanks @mzabaluev, in the end, we don't need this PR. It didn't solve the issue.

PR #749, opened 2024-10-25
Application Priority in Memseq and Integration with Aptos

### LINKED ISSUE #321: LIB-1 The Transaction Sorting Algorithm In RocksdbMempool Can Lead To A Series Of Issues
Severity: Major
Status: Pending
Auditor: Movebit
Topic:Suzuka-only custom components
Discovery Methods: Manual Review

LIB-1 The Transaction Sorting Algorithm In
RocksdbMempool Can Lead To A Series Of Issues

Code Location:
protocol-units/mempool/move-rocks/src/lib.rs#37

Descriptions:
The sorting algorithm for transactions in RocksdbMempool is as follows:

First, transactions are sorted by timestamp (the timestamp is in 2-second increments).
Then, they are sorted by sequence_number.
Finally, they are sorted by transaction hash.

Screenshot 2024-08-08 at 3 08 28 PM

The light_node transaction packing algorithm works as follows:
It packs 10 transactions into one block.
If there are fewer than 10 transactions, it packs a block every second.
This could result in the following situation:
Transactions A and B belong to the same account.
Transaction A has a timestamp of 12 and a sequence_number of 11.
Transaction B has a timestamp of 10 and a sequence_number of 15.
In this case, transaction B would be executed before transaction A. However, Aptos
rules require that transaction A must be executed before transaction B.
This sorting algorithm could lead to a series of unknown issues, which would require a deep
analysis of the Aptos-core module to understand fully.

Suggestion:
It is recommended to refer to the sorting method used by Aptos-core.

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**:  `protocol-units`, 

1. Adds an `application_priority` (low to high) field to `Transaction`.
2. Orders by this field in `Transaction` collections, `MempoolTransaction` collections, and `Memseq`. 
3. Updates `maptos-block-executor` transaction pipe to send along `application_priority` from inverted `MempoolStatus::score()`--effectively implementing application priority for Aptos and Suzuka full node. 

Resolves #321 .

# Testing

1. Added unit tests to assert ordering.
2. e2e tests left unmodified. 

# Outstanding issues
1. We could add an abstraction layer for computing application priority.
2. We will likely want to add an `ApplicationPriority` type at some point to avoid misuse with other u64.


protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -20,24 +20,26 @@ pub struct RocksdbMempool {
...
 		transaction.transaction.id(),
 	))
-	.unwrap();
-	key
+	.map_err(|_| Error::msg("Error writing mempool transaction key"))?;
+	Ok(key)
 }
 
-fn construct_timestamp_threshold_key(timestamp_threshold: u64) -> String {
+fn construct_timestamp_threshold_key(timestamp_threshold: u64) -> Result<String, Error> {
```
mzabaluev, line 38: Same here, no need to make the function fallible.

protocol-units/mempool/move-rocks/src/lib.rs
```diff
@@ -469,14 +471,14 @@ pub mod tests {
 		let path = temp_dir.path().to_str().unwrap();
 		let mempool = RocksdbMempool::try_new(path)?;
 
-		let transaction1 = MempoolTransaction::at_time(Transaction::new(vec![1], 0), 2);
+		let transaction1 = MempoolTransaction::at_time(Transaction::new(vec![1], 0, 0), 2);
```
mzabaluev, line 474: `Transaction::new` now has two integer arguments that are easy to mix up. Time to introduce a builder API?

mzabaluev (changes requested):

networks/suzuka/suzuka-full-node/src/tasks/transaction_ingress.rs
```diff
@@ -14,14 +14,14 @@ use std::time::{Duration, Instant};
 const LOGGING_UID: AtomicU64 = AtomicU64::new(0);
 
 pub struct Task {
-	transaction_receiver: mpsc::Receiver<SignedTransaction>,
+	transaction_receiver: mpsc::Receiver<(u64, SignedTransaction)>,
```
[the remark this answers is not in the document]
l-monninger: Yeah, I mentioned the `ApplicationPriority` type in the summary. Do we want to introduce that and this now?

protocol-units/mempool/move-rocks/src/lib.rs
l-monninger: I can add one to prevent introducing bugs going forward. Should we remove the `::new()` method though? That might be too time consuming right now.

networks/suzuka/suzuka-full-node/src/tasks/transaction_ingress.rs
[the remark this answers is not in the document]
mzabaluev: Yes, that's an alternative to inventing yet another transaction struct e.g. `PrioritizedSignedTransaction`. I'd say the latter is less annoying to deal with, and as long as you don't need any range checks or other invariants on the priority value, an `ApplicationPriority` wrapper type is not necessary.

protocol-units/mempool/move-rocks/src/lib.rs
mzabaluev: It can be done so that `new` captures the essential data and assigns default priority. The builder will start with the data vector, and get the named setters for priority and maybe for binding the sequence number as well.

ayanwaar:
f8e6e303c5257af5b1bca4913031311bb2fdb2a7

Issue #695, opened 2024-10-15
Follower Node Block Hash Mismatch with Testnet at Specific Heights


wengych:
**Describe the bug**
I configured a follower node using the documentation provided by Movement: [Follower Node Documentation](https://github.com/movementlabsxyz/movement/blob/main/docs/movement-node/run/manual/follower-node/README.md). The corresponding `.env` configuration is as follows:

```
CONTAINER_REV=a4ba7a656f429e79bbb637898b451c21159a3dc0
DOT_MOVEMENT_PATH=./.movement
MAPTOS_CHAIN_ID=250 # change this to the chain id of the network you are running
MOVEMENT_SYNC="follower::mtnet-l-sync-bucket-sync<=>{maptos,maptos-storage,suzuka-da-db}/**" # change to the sync bucket for the network you are running
M1_DA_LIGHT_NODE_CONNECTION_PROTOCOL=https
M1_DA_LIGHT_NODE_CONNECTION_HOSTNAME="m1-da-light-node.testnet.movementnetwork.xyz"
M1_DA_LIGHT_NODE_CONNECTION_PORT=443
AWS_ACCESS_KEY_ID=xxxx
AWS_SECRET_ACCESS_KEY=xxxxxx
AWS_DEFAULT_REGION=us-west-1 # change this to match the region of the sync bucket
AWS_REGION=us-west-1 # change this to match the region of the sync bucket
```

The command to start the node is as follows:
`sudo just suzuka-full-node docker-compose follower`

When I ran it for the first time using `just`, the API returned the following data:
```
curl http://localhost:30731/v1
{"chain_id":250,"epoch":"1","ledger_version":"51","oldest_ledger_version":"0","ledger_timestamp":"1728914733620485","node_role":"validator","oldest_block_height":"0","block_height":"18","git_hash":"70be3926ff79ff4cdb0cee928f717fafcd41ecdd"}
```

After cross-referencing the block height with the testnet, I found that the hash for the same block height does not match:
- Localhost: `curl http://localhost:30731/v1/blocks/by_height/18`
  ```
  {"block_height":"18","block_hash":"0x8a12c83f0146c1a13d4be394d7071f4775c8c99dc8cffb61225780b5ade0c4ef","block_timestamp":"1728914733620485","first_version":"50","last_version":"51","transactions":null}
  ```
- Testnet: `curl https://testnet.movementnetwork.xyz/v1/blocks/by_height/18`
  ```
[...]

Issue #595, opened 2024-09-12
(HAL-30) TEST ROLL OVER GENESIS FUNCTION USED IN PRODUCTION


SA124:
(HAL-30) TEST ROLL OVER GENESIS FUNCTION USED IN PRODUCTION
Auditor: Halborn
Severity: Informational

![Screenshot 2024-09-12 at 12 33 24 PM](https://github.com/user-attachments/assets/62ec47f0-3b09-4b5f-a771-d5c3bd464597)

This misuse of a testing function could lead to unexpected behavior, potential security vulnerabilities, and inconsistencies in the blockchain's initial state.

Score
[AO:A/AC:L/AX:L/R:N/S:U/C:N/A:N/I:N/D:N/Y:N (0.0)](https://www.halborn.com/bvss?q=AO%3AA%2FAC%3AL%2FAX%3AL%2FR%3AN%2FS%3AU%2FC%3AN%2FA%3AN%2FI%3AN%2FD%3AN%2FY%3AN)

![Screenshot 2024-09-12 at 12 33 58 PM](https://github.com/user-attachments/assets/f4e9e991-1aa9-4b4e-b846-296299d5c72e)

Recommendation
It is recommended to assess whether the function is suitable for production use, and keep or change it if needed.

l-monninger:
We've assessed this a few times and believe it is safe for our purely execution-based used case. We would, however, like to re-evaluate the use of genesis key before Suzuka Genesis.

l-monninger:
We've modified the manner in which we compute attestations, so this needs to be updated. I am adding determinism and checks for consistency.

PR #750, opened 2024-10-25
Genesis Rollove Determinism

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

1. Adds deteminism to genesis rollover, addressing #595 and #695.
2. Adds tests for e2e consistency against relevant APIs. 

# Testing

# Outstanding issues


protocol-units/execution/opt-executor/src/executor/execution.rs
```diff
@@ -65,9 +64,10 @@ impl Executor {
 		})
 		.await??;
 
-		info!("Block execution compute the following state: {:?}", state_compute);
+		warn!("Block execution compute the following state: {:?}", state_compute);
```
mzabaluev, line 67: This is a normal outcome, why use the WARN level?
I'd rather bump this down to debug to eliminate the spam from dumping `state_compute`.

PR #753, opened 2024-10-28
Bridge Synchronize tx send to both client.

musitdev:
# Summary

To avoid sending 2 Tx at the same time to the same client, which could possibly generate a nonce error, Tx are synchronized per client.
# Changelog

# Testing
The bridge e2e test Eth->Mvt direct pass.
I didn't test using concurrent Tx. It needs to develop more testing code to test the relayer logic.

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/service/src/lib.rs
+++ b/protocol-units/bridge/service/src/lib.rs
@@ -13,7 +13,9 @@ use crate::types::BridgeTransferId;
 use crate::types::ChainId;
 use futures::stream::FuturesUnordered;
 use std::collections::HashMap;
+use std::sync::Arc;
 use tokio::select;
+use tokio::sync::Mutex;
 use tokio_stream::StreamExt;
 
 mod actions;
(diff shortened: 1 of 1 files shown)


protocol-units/bridge/service/src/lib.rs
```diff
@@ -70,15 +76,27 @@ where
 									ChainId::ONE => {
 										let fut = process_action(action, one_client.clone());
 										if let Some(fut) = fut {
-											let jh = tokio::spawn(fut);
+											let jh = tokio::spawn({
+												let client_lock_clone = one_client_lock.clone();
+												async move {
+													let _lock = client_lock_clone.lock().await;
+													fut.await
```
andygolay, line 83: Do you think any error handling could potentially be useful here, like

```
if let Err(e) = fut.await {
    eprintln!("Error in task: {:?}", e);
}
```

likewise in line 97?

andygolay (approved):
Looks good, I ran the E2E test locally and it passed. 

I left [one comment](https://github.com/movementlabsxyz/movement/pull/753/files#r1818738314), more of an inquiry than a change request.

protocol-units/bridge/service/src/lib.rs
musitdev: The error is managed later. The fut return is added to the list of future (line 86) to execute and after execution the error is handle. See line 145 for example.

0xmovses (approved):

PR #754, opened 2024-10-28
relayer remove asset type, fix conversion

0xmovses:
# Summary

# Changelog

# Testing

# Outstanding issues


protocol-units/bridge/service/src/types.rs
```diff
@@ -136,59 +136,13 @@ impl From<Uint<256, 4>> for TimeLock {
...
-	}
-
-	pub fn weth_value(&self) -> u64 {
-		match self.0 {
-			AssetType::EthAndWeth((_, weth_value)) => weth_value,
-			AssetType::Moveth(_) => 0,
-		}
-	}
-}
+pub struct Amount(pub u64);
```
musitdev, line 139: You use u64. If it's wei for example, we should use u128 because u64 is to small for important eth amount.

0xmovses: `Uint<256, 4>` has support for conversion to u64 via its limbs methods but not for u128. 

This is a different version but the `Uint` methods are the same https://docs.rs/ruint/1.12.3/src/ruint/lib.rs.html#146

andygolay (approved):
LGTM

PR #758, opened 2024-10-28
remove pool balance

0xPrimata:
# Summary

Removes Pool Balance from eth contracts.

# Changelog

# Testing

> cd /Users/user/movement/protocol-units/bridge/contracts

> forge test

# Outstanding issues

There are tests that are failing in main branch. Disregard them

### DIFF
--- a/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol
+++ b/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol
@@ -57,8 +57,6 @@ contract AtomicBridgeCounterpartyMOVE is IAtomicBridgeCounterpartyMOVE, OwnableU
         uint256 amount
     ) external onlyOwner returns (bool) {
         if (amount == 0) revert ZeroAmount();
-        if (atomicBridgeInitiatorMOVE.poolBalance() < amount) revert InsufficientMOVEBalance();
-
         // The time lock is now based on the configurable duration
         uint256 timeLock = block.timestamp + counterpartyTimeLockDuration;
 
--- a/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol
+++ b/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol
@@ -55,7 +55,7 @@ contract AtomicBridgeCounterpartyMOVE is IAtomicBridgeCounterpartyMOVE, OwnableU
         bytes32 hashLock,
         address recipient,
         uint256 amount
-    ) external onlyOwner returns (bool) {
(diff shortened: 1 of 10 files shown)


0xmovses:
@0xPrimata run the tests with the command:
`forge test --fork-url https://ethereum-sepolia-rpc.publicnode.com -vv`

If in doubt always check CI declarations, to find commands: https://github.com/movementlabsxyz/movement/blob/e57af879a13ed5e175f86f65af5188ba6647cbb8/.github/workflows/checks-all.yml#L249

0xmovses:
Let's check integration tests and e2e bridge tests on this before merging.

protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol
```diff
@@ -57,8 +57,6 @@ contract AtomicBridgeCounterpartyMOVE is IAtomicBridgeCounterpartyMOVE, OwnableU
         uint256 amount
     ) external onlyOwner returns (bool) {
         if (amount == 0) revert ZeroAmount();
-        if (atomicBridgeInitiatorMOVE.poolBalance() < amount) revert InsufficientMOVEBalance();
```
andygolay, line 60: Without this check, how are we sure that there's enough MOVE available to cover the transfer?

0xPrimata: Because you can only bridge back enough MOVE that was bridged before and all bridged MOVE is contained in the eth side of the contracts.

0xmovses: Yeah Primata is correct, you don't need poolBalance to do this check, the tx will revert if the contract doesn't have enough MOVE balance using the standard ERC20 checks

0xPrimata: Besides that revert, we have to work with the assumption that locked amount on Ethereum contract equals to circulating supply on Movement.

0xmovses: On Movement Barcok yes. Let's set this correct amount on the new deployments.

andygolay: For now on Bardock that wouldn't be true due to the ability to use the Bardock faucet, but for mainnet I see how that reasoning makes sense.

0xmovses (approved):
Framework e2e passing

0xmovses:
Forge tests passing

Issue #480, opened 2024-08-30
Redo config structures to support hierarchical naming


mzabaluev:
Discussion: #384

Issue #408, opened 2024-08-21
Run unit tests in CI


mzabaluev:
Discussion: #394

There shall be CI jobs to run unit tests and lightweight integration tests.
All available tests that don't require infrastructure setup should be run at every push.

See also #407, which explores the coverage aspect.

PR #765, opened 2024-10-29
Read-only mode for Suzuka node

mzabaluev:
# Summary

Add a read-only flag to configuration.
If set, disable transaction submission and related requests
in the API.

# Changelog

* Read-only mode for executor, activated with the `MAPTOS_READ_ONLY` boolean environment variable.

# Testing

TODO: add tests to enable the mode and confirm the API behavior is as expected.

### DIFF
--- a/protocol-units/execution/dof/src/v1.rs
+++ b/protocol-units/execution/dof/src/v1.rs
@@ -32,7 +32,7 @@ impl Executor {
 		Self { executor, finality_view }
 	}
 
-	pub fn try_from_config(config: &Config) -> Result<Self, anyhow::Error> {
+	pub fn try_from_config(config: Config) -> Result<Self, anyhow::Error> {
 		let executor = OptExecutor::try_from_config(config)?;
 		Ok(Self::new(executor))
 	}
@@ -53,7 +53,6 @@ impl DynOptFinExecutor for Executor {
 	fn background(
 		&self,
 		transaction_sender: Sender<SignedTransaction>,
-		config: &Config,
 	) -> Result<
 		(Context, impl Future<Output = Result<(), anyhow::Error>> + Send + 'static),
 		anyhow::Error,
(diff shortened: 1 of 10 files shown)


protocol-units/execution/util/src/config/chain.rs
```diff
@@ -27,6 +27,9 @@ pub struct Config {
 	#[serde(default = "default_maptos_private_key")]
 	pub maptos_private_key: Ed25519PrivateKey,
 
+	#[serde(default)]
+	pub read_only: bool,
```
mzabaluev, line 31: The "chain" config is perhaps not a good name for this configuration section,
since we also have the `maptos_rest_*` fields here (the private key is here too, but it should be removed as per #387). Maybe we should move all these to a separate configuration section, or rename this one.

mzabaluev: By the current naming convention, this should be `maptos_read_only`, but see #480

l-monninger: This is fine.

protocol-units/execution/dof/src/v1.rs
```diff
@@ -162,38 +162,49 @@ mod tests {
...
 			AccountAddress::random(),
 			0,
 			transaction_payload,
 			0,
-			gas_unit_price,
+			0,
 			0,
 			ChainId::test(), // This is the value used in aptos testing code.
 		);
+		// TODO: actually sign
```
andyjsbell, line 190: Is this maybe better as an issue?

mzabaluev: Filed as #814 and added to #408.

l-monninger (changes requested):

protocol-units/execution/dof/src/v1.rs
```diff
@@ -162,38 +162,49 @@ mod tests {
...
+	fn create_signed_transaction(private_key: &Ed25519PrivateKey) -> SignedTransaction {
 		let public_key = private_key.public_key();
 		let transaction_payload = TransactionPayload::Script(Script::new(vec![0], vec![], vec![]));
 		let raw_transaction = RawTransaction::new(
 			AccountAddress::random(),
 			0,
 			transaction_payload,
 			0,
-			gas_unit_price,
+			0,
```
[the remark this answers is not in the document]
mzabaluev: This is test code, it's enough that the mempool accepts the transaction for the test, and gas price is not a variable test parameter here.

Also, the way to generate these test transactions needs to be changed as discussed in #814.

l-monninger (approved):

PR #771, opened 2024-10-29
Bridge: set from before gas_estimation

musitdev:
# Summary

Update eth send_transaction to set the from(signer_address) before gas_estimation.
Correct an issue where gas_estimation doesn't set correctly the sender and revert onlyOwner contract call.
# Changelog

# Testing

# Outstanding issues


0xmovses (approved):
e2e passing, code looks fine. We love alloy don't we ..

Issue #774, opened 2024-10-30
Wakeup Relayer


0xmovses:
I observed that after a period of about 12 hours the relayer can sometimes become idle and stop receiving events. I checked with Radu, there is no special network configurations in terraform / aws that would cause this. It's probably something at the application level. 

- Update the main function or lib.rs to keep the loop running indefinitely.

PR #782, opened 2024-10-30
feat: keep relayer alive

0xmovses:
# Summary
Issue: #774 
Add a task that runs every three minutes to ensure relayer is alive. 

Note that when I noticed the relayer going idle no errors were emitted and the health endpoint still responded 200, this is why I have chosen this approach, rather than checking health within docker setup.

# Changelog
Adds periodic wakeup task

# Testing
This has to be observed over a 12 hour period in production, if the relayer continues to relay events after this time then the issue is resolved.


andygolay (approved):
LGTM once CI passes

0xmovses:
We will hold off on merging as so far relayer is still processing events now after 12 hours in production.

andygolay:
It seems like this same issue has happened again, see #806 for what I observed over the weekend for Movement -> Eth swaps. We just restarted the bridge, so we can see how it goes today, but if the same behavior persists I think it would make sense to try implementing this Keep Relayer Alive task as a possible fix.

musitdev (approved):

PR #784, opened 2024-10-30
Implement dynamic gas fees for bridge transfers

0xmovses:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/service/src/chains/ethereum/client.rs
+++ b/protocol-units/bridge/service/src/chains/ethereum/client.rs
@@ -650,11 +650,32 @@ impl GasMaster for EthClient {
 	}
 
 	async fn get_priority_gas_fee_estimate(&self) -> GasMasterResult<u64> {
-		unimplemented!()
+		let params = ("0x1", "latest", vec![0.2, 0.5, 0.9]);
+		let fee_history: serde_json::Value = self
+			.rpc_provider
+			.raw_request(Cow::Borrowed("eth_feeHistory"), params)
+			.await
+			.map_err(|_| GasMasterError::GetPriorityGasFeeEstimateError)?;
+
+		let priority_fee = fee_history["reward"]
+			.as_array()
+			.and_then(|rewards| rewards.last())
+			.and_then(|last_block_rewards| last_block_rewards[0].as_str())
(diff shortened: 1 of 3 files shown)


protocol-units/bridge/service/src/chains/ethereum/client.rs
```diff
@@ -650,11 +650,32 @@ impl GasMaster for EthClient {
...
 	}
 
 	async fn calculate_dynamic_fee(&self) -> GasMasterResult<u64> {
-		unimplemented!()
+		let base_gas_price = self.get_base_gas_price().await?;
+		let priority_fee = self.get_priority_gas_fee_estimate().await?;
+		let percentage_adjustment = 0.2; // 20% adjustment hardcoded for now
+
+		// Apply the percentage adjustment to base gas price
+		let adjusted_base_fee =
```
musitdev, line 676: This part seem a little strange. You multiply an u64 by a U256 and the result is an u64. I think you use U256 because of the precision, but the U256 result is transformed to an u64. Not sure but you should convert the `base_gas_price` in a U256 and do the calculus on u256 for all and do the conversion to u64 at the end.
Here you let the compiler decide when to do the conversion.

protocol-units/bridge/service/src/chains/ethereum/client.rs
```diff
@@ -650,11 +650,32 @@ impl GasMaster for EthClient {
 	}
 
 	async fn get_priority_gas_fee_estimate(&self) -> GasMasterResult<u64> {
```
musitdev, line 652: Perhaps we should use u128 for all gas and amount value because of the decimal use in Eth for example, you can exceed the u64 limit. I got the issue when calculating the transaction fee for Eth. The code: `let transaction_fee_wei = estimate_gas * gas_price;` has exceeded u64 during my test.

protocol-units/bridge/service/src/chains/ethereum/client.rs
musitdev: More generally, we should use checked calculus functions like `checked_mul` or `saturating_mul` (if we don't care to detect overflow) for all our calculus.

PR #786, opened 2024-10-30
Feat: Fund Movement recipient

andygolay:
# Summary
- Adds a call to fund the Movement recipient when locking, so they have some funds to cover transaction fees when they call `complete_bridge_transfer`.

# Changelog

In `process_action` add a funding step before `lock_bridge_transfer` if the `recipient` field contains a 32-byte Movement address. This guarantees the recipient will have funds to to subsequently call `complete_bridge_transfer`.

# Testing

Run bridge:

```
```

Then run the test:

```
DOT_MOVEMENT_PATH="$(pwd)/.movement" cargo test --test bridge_e2e_test_framework test_bridge_transfer_eth_movement_happy_path -- --nocapture --test-threads=1
```

Before funding you'll see an error logged due to the recipient Movement account not exisiting yet. After funding, the balance is 1 MOVE.

# Outstanding issues
Intended to be temporary, for testnet only.

### DIFF
--- a/protocol-units/bridge/service/src/actions.rs
+++ b/protocol-units/bridge/service/src/actions.rs
@@ -82,7 +82,6 @@ where
 
 				if recipient.0.len() == 32 {
 					if let Err(e) = movement_utils::fund_recipient(&recipient).await {
-					tracing::error!("Funding failed: {:?}", e);
 					return Err(ActionExecError(action.clone(), e));
 					}
 				}
--- a/protocol-units/bridge/service/src/actions.rs
+++ b/protocol-units/bridge/service/src/actions.rs
@@ -1,3 +1,6 @@
+use anyhow::anyhow;
+use aptos_sdk::rest_client::{Client as RestClient, FaucetClient};
+use aptos_sdk::types::account_address::AccountAddress;
 use crate::chains::bridge_contracts::BridgeContract;
 use crate::chains::bridge_contracts::BridgeContractError;
 use crate::types::Amount;
@@ -10,6 +13,8 @@ use std::fmt;
 use std::future::Future;
 use std::pin::Pin;
 use thiserror::Error;
+
+use url::Url;
 
 #[derive(Error, Debug, Clone)]
 pub struct ActionExecError(TransferAction, BridgeContractError);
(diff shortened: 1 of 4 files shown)


0xmovses (changes requested):
Looking good, few improvements needed.

protocol-units/bridge/service/src/actions.rs
```diff
@@ -83,6 +88,53 @@ where
...
+
+					/// Fetch and log balance before funding
+					/// This should return an error for a new address because funding creates the account
+					match rest_client.get_account_balance(account_address).await {
+						Ok(balance) => tracing::info!("Balance before funding: {:?}", balance),
+						Err(e) => tracing::error!("Failed to retrieve balance before funding: {:?}", e),
+					}
+					
+					// Execute the funding transaction and capture the result
+					match faucet_client
```
musitdev, line 117: Instead of match, you should use `.map_err(|err| ActionExecError(action, err))` or something similar. You don't need to log anything because returned error are logged when catched.
Changing all match you make the code simpler and easier to read.
In fact, you can link all the process using .and_then() after each client process.

0xmovses: Agree, matching on Result is overly verbose, you see it in some example code but generally there are more concise way to handle result types.

0xmovses (changes requested):

protocol-units/bridge/service/src/actions.rs
andygolay: @0xmovses @musitdev I cleaned up the code and refactored the fund_recipient function. 

I tried to remove match from the logic and do this instead:

```
pub async fn fund_recipient(recipient: &BridgeAddress<Vec<u8>>) -> Result<(), BridgeContractError> {
	// Parse URLs
	let faucet_url = Url::parse(MOVEMENT_FAUCET_URL)
	.map_err(|_| BridgeContractError::InvalidUrl)?;
	let rest_url = Url::parse(MOVEMENT_RPC_URL)
	.map_err(|_| BridgeContractError::InvalidUrl)?;
	
	// Create clients
	let rest_client = RestClient::new(rest_url.clone());
	let faucet_client = FaucetClient::new(faucet_url, rest_url);

	// Convert recipient to AccountAddress
	let recipient_address: [u8; 32] = recipient.0.clone().try_into()
	.map_err(|_| BridgeContractError::SerializationError)?;
	let account_address = AccountAddress::new(recipient_address);

	// Check balance before funding
	rest_client.get_account_balance(account_address)
	.await
	.map_err(|_| BridgeContractError::AccountBalanceError)?;

	// Execute the funding transaction
	faucet_client.fund(account_address, 100_000_000)
	.await
	.map_err(|_| BridgeContractError::FundingError)?;

	// Check balance after funding
	rest_client.get_account_balance(account_address)
	.await
	.map_err(|_| BridgeContractError::AccountBalanceError)?;

	Ok(())
}
```

and I also tried mapping to `ActionExecError`. What happens in both cases is I get the error bubbling up and causing the bridge transfer to fail:
image

When I removed the balance checks (most recent commit) there's no error but I think if the fund step fails the transfer will fail. So it may be better to use the match pattern? 

Also, do you think BridgeContractError is fine, because ActionExecError contains a BridgeContractError anyway, so for this funding helper function it seems fine to just use BridgeContractError?

protocol-units/bridge/service/src/actions.rs
```diff
@@ -83,6 +79,14 @@ where
 			amount,
 		} => {
 			let future = async move {
+
+				if recipient.0.len() == 32 {
+					if let Err(e) = movement_utils::fund_recipient(&recipient).await {
+					tracing::error!("Funding failed: {:?}", e);
```
musitdev, line 85: You don't need to trace all action execution error are traced [here](https://github.com/movementlabsxyz/movement/blob/78252235e36c45f07bdf806f24809a23d0da7392/protocol-units/bridge/service/src/lib.rs#L282) 
This is here where you can define some specific process when you get and error.

0xmovses (approved):
To follow up with this https://github.com/movementlabsxyz/movement/issues/792

Thanks for this e2e passing

PR #789, opened 2024-10-30
Bridge: correct eth monitoring issue with Move token contract

musitdev:
# Summary

Change the contract used by the ETH monitoring to Move token one.
Update the deployment setup.
# Changelog

# Testing
The e2e test Eth->Mvt doesn't pass because the move token setup is incorrectly done.

# Outstanding issues


protocol-units/bridge/config/src/common/eth.rs
```diff
@@ -38,6 +39,8 @@ pub struct EthConfig {
 	pub eth_counterparty_contract: String,
 	#[serde(default = "default_eth_weth_contract")]
 	pub eth_weth_contract: String,
+	#[serde(default = "default_eth_move_token_contract")]
+	pub eth_move_token_contract: String,
```
0xmovses, line 43: This is intended for the MOVE ERC20 contract right?

0xmovses (approved):

PR #791, opened 2024-10-30
update: docker files with env val

0xmovses:
# Summary
- Adds an env var for a new config field. 

# Changelog
- Update Dockerfile

# Testing
This value is currently set in the production relayer, building and running, testing strategy inspired by @0xPrimata


protocol-units/bridge/config/src/common/eth.rs
```diff
@@ -79,7 +79,7 @@ env_default!(
 );
 env_default!(
 	default_eth_move_token_contract,
-	"ETH_MOVETOKEN_CONTRACT",
+	"ETH_MOVE_TOKEN_CONTRACT",
```
musitdev, line 82: If you change it here, you must apply the same change in the docker file.

musitdev (approved):

andyjsbell:
@0xmovses can this be merged?

0xmovses:
Yes, will fix future deployment.

Issue #775, opened 2024-10-30
Persist Relayer Transfers


0xmovses:
When the relayer is restarted the pullstate file gets wiped. We need to:

- Ensure the file persists on disk across restarts (systemd features?)
- Implement a mechanism to ensure that the first thing the relayer does is check this file and continue transfers for any pending ones.

musitdev:
Why the pullstate.store feil is wiped? The relayer didn't remove it.

0xmovses:
When the service is restarted with `systemctl restart <service>` the array 
```
INFO bridge_service: Bridge current transfer processing [ ]
```
Is always empty, no pending transfers are persisted. It could be a fix we need to make within the docker compose files.

0xmovses:
I'm doing a local test now to check its working as expected locally.

musitdev:
Could we get them from the blockchain by requesting `bridgeTransfers` fields of the Initiator contract? I should be more reliable. We've to ensure that transfer finished are removed from the list. S by getting the list we get all pending transfer.

0xmovses:
Yes, that's a good idea, the contract will return the state of each transfer and that should be our source of truth. I'll start working on this.

andygolay:
> Could we get them from the blockchain by requesting `bridgeTransfers` fields of the Initiator contract? I should be more reliable. We've to ensure that transfer finished are removed from the list. So by getting the list we get all pending transfer.

When I call `bridgeTransfers()` for example like

```
			InitiatorContract::Move(instance) => {
				for (bridge_transfer_id, transfer) in instance.bridgeTransfers()
```

It says it's expecting an arg. And the contract does seem like you're supposed to do `bridgeTransfers(<bridgeTransferId>)` to return details of one transfer. 		

See https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol#L26		
	
So it seems like I'll need to add a view function to the contract, to get all the transfers... does that seem correct or am I missing some way to view all the transfers? 

However to make a view function work, we'll need to store the bridge transfer IDs somewhere as well, because `mapping` doesn't appear to support length.

Maybe it's better to simply, on each `run_bridge` loop, cache `logs` into some file that persists? That way we don't need to do any blockchain transfers?

musitdev:
[...]

PR #794, opened 2024-10-31
Bridge Indexed DB

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

1. Introduced and indexed (relational) database covering bridge `TransferActionType` and `BridgeContractEvent` types. 
2. Adds orchestration and GraphQL API serving with Hasura. 

- [Schema](https://github.com/movementlabsxyz/movement/blob/l-monninger/bridge-graphql/protocol-units/bridge/indexer-db/migrations/2024-10-31-112521_indexed_bridge_actions/up.sql)
- [Client](https://github.com/movementlabsxyz/movement/blob/l-monninger/bridge-graphql/protocol-units/bridge/indexer-db/src/client.rs)

### Using the GraphQL Endpoint

**Get Everything By Initiator and Recipient**:
```graphql
query getBridgeTransfersByInitiatorAndRecipient($initiator: String!, $recipient: String!) {
  lock_bridge_transfers(
    where: {
      initiator: { _eq: $initiator },
      recipient: { _eq: $recipient }
    }
  ) {
    id
    bridge_transfer_id
    hash_lock
    initiator
    recipient
    amount
  }

  initiated_events(
    where: {
      initiator_address: { _eq: $initiator },
      recipient_address: { _eq: $recipient }
    }
  ) {
    id
    bridge_transfer_id
    initiator_address
    recipient_address
    hash_lock
    time_lock
    amount
    state
  }

  locked_events(
    where: {
      initiator: { _eq: $initiator },
      recipient: { _eq: $recipient }
    }
  ) {
    id
    bridge_transfer_id
    initiator
    recipient
    hash_lock
    time_lock
    amount
  }
}
```

**Get Bridge Transfer IDs by Initiator Address**
```graphql
query getBridgeTransferIdsByInitiator($initiator: String!, $limit: Int!, $offset: Int!) {
  lock_bridge_transfers(
    where: { initiator: { _eq: $initiator } }
    limit: $limit
    offset: $offset
  ) {
    bridge_transfer_id
  }
}

```

**Get everything by Bridge Transfer ID**
```graphql
query getBridgeTransferEvents($bridge_transfer_id: String!) {
  lock_bridge_transfers(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
    hash_lock
    initiator
    recipient
    amount
  }
  
  wait_and_complete_initiators(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    wait_time_secs
    pre_image
  }

  initiated_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
    initiator_address
    recipient_address
    hash_lock
    time_lock
    amount
    state
  }

  locked_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
    initiator
    recipient
    hash_lock
    time_lock
    amount
  }

  initiator_completed_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
  }

  counter_part_completed_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
    pre_image
  }

  cancelled_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
  }

  refunded_events(where: { bridge_transfer_id: { _eq: $bridge_transfer_id } }) {
    id
    bridge_transfer_id
  }
}
```

# Testing

# Outstanding issues


0xmovses:
cc @andygolay & @musitdev

protocol-units/bridge/indexer-db/migrations/2024-10-31-112521_indexed_bridge_actions/up.sql
```diff
@@ -0,0 +1,56 @@
+CREATE TABLE lock_bridge_transfers (
+    id SERIAL PRIMARY KEY,
+    bridge_transfer_id BYTEA NOT NULL,
+    hash_lock BYTEA NOT NULL,
+    initiator BYTEA NOT NULL,      -- Address stored as bytes
```
0xmovses, line 5: Might we want to limit this to 32 bytes?

protocol-units/bridge/indexer-db/src/client.rs
```diff
@@ -0,0 +1,88 @@
...
+					.execute(&self.conn)?;
+			}
+			TransferActionType::WaitAndCompleteInitiator(u64, HashLockPreImage) => {
+				diesel::insert_into(transfer_actions::table)
+					.values(NewTransferAction { action_type })
+					.execute(&self.conn)?;
+			}
+			TransferActionType::RefundInitiator => {
+				// do nothing
+			}
```
0xmovses, line 36: While we are here. I think there is value to storing Refund Actions in the DB also. Purely for internal reports.

0xmovses (changes requested):
Looks good, let's store refund and cancel actions in the DB also. We had discussed the possibility of a DB a few times within the Bridge team, it was concluded that the indexer suffice. However, I must say I prefer this more targeted implementation. Just a simple DB we can maintain for our own purposes. 

Hash this been tested? Or is the idea to add tests in a followup PR?

0xmovses:
This fixes #775 , the DB can be queried at startup and transfers not yet completed but initialized can continue. I would say @andygolay wait for this to land, then we can implement the startup-and-continue logic which should be fairly trivial.

0xmovses:
Not directly related to the PR, but freely associating here.

Note all inputs on the solidity contracts for `bytes` are limited to `bytes32`, so any possible path for an SQL injection are thwart there.

However on the `atomic_bridge.move` the input here is simply `vector<u8>` could this be exploited as it allows arbitrary length? https://github.com/movementlabsxyz/aptos-core/blob/7a0e210fe29e2c81c378568eba2b467a7ef6a56e/aptos-move/framework/aptos-framework/sources/atomic_bridge.move#L176

We should limit to 32 bytes length, this is more relevant now we have a db involved. Something like:
```
fun create_fixed_length_vector(input: vector<u8>): vector<u8> acquires SomeResource {
    assert!(Vector::length<u8>(&input) == 32, 0); // Checks if the length is exactly 32
    input
}
```

musitdev:
I think we should dissociate all this process from the Relayer. The role of the relayer is to relay transfer between two blockchains. Perhaps we are going to support other blockchain, relay using different contract on the same chain or other tokens. If we add this management inside the relayer for each evolution will need to update this part and perhaps will see some incompatibilities.
I think the role of this process is the same has the indexer for suzuka node, so we should keep the same architecture. This indexer get the event from the blockchain to update its data like the suzuka indexer. 
Relayer and indexer it's not the same domain so they should be separated.

0xmovses:
Yes @musitdev I had raised a similar idea to @l-monninger last night, about using our existing indexer infra for this. However this presents a problem, we must index eth-side separately. This means using a subgraph, this is not a trivial task, to deploy host and and run an ethereum subgraph. 

I do agree that we don't want to add load to the relayer. @l-monninger how about making this a seperate binary and using the already gRPC port to get the data from the relayer?

musitdev:
Both Indexer has different role. They can share the same Postgres instance and host, but they should be developed in an independent manner. Bridge indexer should index all blockchain that we bridge to. 
To get its data the best is to use the bridged blockchains and not the relayer. It's not its role. For example, it doesn't keep a history of all transfer. To start, the indexer needs to access the transfer history to update its data from its last sync point. I don't think the relayer should provide this function, and all blochain provide it out of the box.
So I think the bridge indexer should be an independent project where we define what to index, how to get it and how to store it.

0xmovses:
So, you're proposing that the indexer get its data _only_ from the origin and destination chains and that the relayer should not be used as a source of transfer data. 

What about hot transfers though? I suppose we could ascertain that from the contract's bridgeTransfer state field ..

protocol-units/bridge/indexer-db/src/client.rs
l-monninger: They don't have any data though. What do you want to store?

protocol-units/bridge/indexer-db/migrations/2024-10-31-112521_indexed_bridge_actions/up.sql
l-monninger: You can't restrict `BYTEA` like that. You can `CHAR(64)`. But, `BYTEA` is supposed to be better on disk and better for exact matches when querying on most engines.

protocol-units/bridge/indexer-db/src/client.rs
0xmovses: Yeah, that's yet to be completed. We update later

protocol-units/bridge/indexer-db/migrations/2024-10-31-112521_indexed_bridge_actions/up.sql
0xmovses: Ok, `BYTEA` is fine, the contract args are themselves restricted (only on eth side), we should restrict the move calls also, as an aside.

l-monninger:
@musitdev this is in many ways quite similar to how the indexer works for Aptos. Subgraph is more complex, but I don't think we need to go there.

This is a quick solution, does not generate many external dependencies, and if not used permanently can still be used for relayer observability.

0xPrimata:
My understanding is that this is ideally coupled with the relayer for observability and reducing load on infra. Ideally this serves partners as well who will want to use our bridge for 1:1 asset bridging which is inexpensive compared to liquidity bridging.

PR #810, opened 2024-11-05
Bridge: add event monotiring health check

musitdev:
# Summary

From the issue #782. Add health check for monitoring loop.
# Changelog
Add heath check in eth and movement event monitoring loop.
Update the Rest health check to get the result of monitoring loop check.
Add timeout on monitoring Rest call.

# Testing
No specific test until the e2e test is workgin. Test that the bridge start and the Rest entry point return Ok.

# Outstanding issues


andygolay (approved):
I tested that it builds, code looks good.

protocol-units/bridge/service/src/lib.rs
```diff
@@ -20,6 +20,28 @@ pub mod rest;
...
+struct HeathCheckStatus {
+	chain_one: bool,
+	chain_two: bool,
+}
+
+impl HeathCheckStatus {
+	fn new() -> Self {
+		HeathCheckStatus { chain_one: true, chain_two: true }
+	}
+	fn check(&self) -> bool {
```
andyjsbell, line 33: If this is just a logical AND why don't we just have the one bool and and the setter `update_state` simply && the current state instead?

protocol-units/bridge/service/src/main.rs
```diff
@@ -42,10 +42,13 @@ async fn main() -> Result<()> {
 
 	tracing::info!("Bridge config loaded: {bridge_config:?}");
 
-	let one_stream = EthMonitoring::build(&bridge_config.eth).await.unwrap();
+	let (eth_health_tx, eth_health_rx) = tokio::sync::mpsc::channel(10);
+	let one_stream = EthMonitoring::build(&bridge_config.eth, eth_health_rx).await.unwrap();
```
andyjsbell, line 46: It's probably me but wouldn't `stream_a` or `stream_one` be a little easier to read

andyjsbell (approved):
Some nits but as a whole looks good

protocol-units/bridge/service/src/lib.rs
musitdev: There's 2 monitoring loop. I think it's important to be able to detect which loop failed. In the health check return response, when it's ok it return OK, but when it failed, it returns the details. This check is used to know if there's any failed loop or not.

protocol-units/bridge/service/src/main.rs
musitdev: I really don't know. We start with one and two chain, so if there isn't a good reason to change, we should stay like that.

protocol-units/bridge/service/src/chains/ethereum/event_monitoring.rs
```diff
@@ -58,265 +62,352 @@ impl EthMonitoring {
...
+						}
+						Err(mpsc::error::TryRecvError::Empty) => (), //nothing
+						Err(err) => {
+							tracing::warn!("Check Eth monitoring loop health channel error: {err}");
+						}
+					}
+
+					//Get block number.
+					let block_number = match tokio::time::timeout(
+						tokio::time::Duration::from_secs(5),
```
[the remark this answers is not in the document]
musitdev: No real reason, it's a reasonable time to wait for a REST call. Longer call mostly means an issue. But we need to test it in prod if it doesn't generate false timeout.
So 5 sec is a good starting point.

protocol-units/bridge/service/src/main.rs
andygolay: It's a language convention, regarding the ordering of words. `stream_one` is more English and `one_stream` is more French.

It's similar to how in English today's date is "November 6" but in French it's "Le 6 Novembre". 

I do tend to think because this code is based on English, having the number after the name of what is being numbered (eg `stream_one` and `stream_two`), could be helpful for consistency with English language conventions. I can understand it either way, but it's something worth considering.

musitdev: Ok I thought it was to replace `one` with `a` and `two` with `b` not the ordering.
Ok yes you're right it's better, I'll change it.

protocol-units/bridge/service/src/chains/ethereum/event_monitoring.rs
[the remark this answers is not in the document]
musitdev: Ok I add it to the bridge config.

andygolay (approved):

protocol-units/bridge/service/src/chains/ethereum/event_monitoring.rs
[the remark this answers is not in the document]
musitdev: I've updated the code. You can review the new change.

PR #813, opened 2024-11-06
Fix up ansible/follower-node documentation and recipes

mzabaluev:
# Summary
- **Categories**: `scripts`.

The follower node setup provisioned with Ansible is supposed to use the follower overlay, like in the README.


mzabaluev:
It's worth checking if other parameters can be added to the template. For example, we could set the network ID to save people the trouble that Rahat recently had.

radupopa369 (approved):

PR #819, opened 2024-11-07
Avoid races in CD container builds over symbolic references in checkouts

mzabaluev:
# Summary
- **Categories**: `cicd`.

Don't supply the symbolic GitHub reference name to the `ref` parameter of the checkout action in the build-push-container-all workflow jobs.
The default behavior of actions/checkout would do the right thing, checking out the SHA of the triggering event, thereby avoiding races with concurrent pushes to the repository modifying the symbolic references.

### DIFF
--- a/.github/workflows/build-push-containers-all.yml
+++ b/.github/workflows/build-push-containers-all.yml
@@ -28,7 +28,6 @@ jobs:
       uses: actions/checkout@v4
       with: 
         submodules: true
-        ref: ${{ github.event.pull_request.head.ref || github.ref }}
     
     - name: Login to Docker Hub
       uses: docker/login-action@v3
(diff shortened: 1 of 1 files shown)


l-monninger:
@mzabaluev I believe this will cause it to check out detached heads when responding to a label event.

mzabaluev:
> @mzabaluev I believe this will cause it to check out detached heads when responding to a label event.

That may be actually the correct way to run: the symbolic branch head may have been pushed to another commit by the time the job runs, which has happened e.g. in https://github.com/movementlabsxyz/movement/actions/runs/11726835429/job/32668512421#step:2:215

I can restore the statements as well, but replace the `.ref` with `.sha`.

Issue #823, opened 2024-11-08
Bridge: Add the DB url in the config


musitdev:
The bridge database URL is defined [here](https://github.com/movementlabsxyz/movement/blob/598ade4a8f318f93ebced5ab24e1738c4b1c356d/protocol-units/bridge/indexer-db/src/client.rs#L31) using an env variable. 

To respect the way the bridge configuration is done it should be added to the Bridge config with a default value for local execution.

PR #690, opened 2024-10-11
Bridge Mvt -> Eth transfer e2e Test

musitdev:
# Summary
- **RFCs**: [Link to RFC](#./link/to/rfc), [Link to RFC](#./link/to/rfc), or $\emptyset$.
- **Categories**: any of `protocol-units`, `networks`, `scripts`, `util`, `cicd`, or `misc`.

Mvt -> Eth Transfer e2e test.
# Changelog
Update the contract deployment and add the initialize call to the counterpart contract.
Add the `test_eth_client_lock_then_complete_transfer` to test counterpart contract call.

# Testing

I create a test that recreate the issue:

RUST_BACKTRACE=1 cargo test --test client_eth_movement test_eth_client_lock_then_complete_transfer -- --nocapture --test-threads=1 
# Outstanding issues

Still an issue with LockTransfer call that don't works

PR #824, opened 2024-11-09
Create bridge config for DB URL

Icarus131:
# Summary
- **RFCs**: $\emptyset$
- **Category**: `protocol-units`

Changed the database URL configuration for the bridge indexer from environment variable to structured config as specified in #823 .

# Changelog

- Added `DatabaseConfig` struct in config crate with default PostgreSQL URL
- Integrated database config into main `Config` struct
- Removed `from_env()` method in favor of `from_bridge_config()` for consistent configuration handling

# Testing
- `bridge_service`  which depends on the `from_bridge_config()` function builds successfully
- integration tests pending


protocol-units/bridge/config/src/lib.rs
```diff
@@ -17,6 +17,10 @@ pub struct Config {
 	/// Optional testing config
 	#[serde(default)]
 	pub testing: common::testing::TestingConfig,
+
+	/// Database configuration for the bridge indexer
+	#[serde(default)]
```
l-monninger, line 22: I would call this `indexer_database` and the related module the same. This is just because there are other forms of persistence we've discussed already. But, otherwise, this is good.

l-monninger (changes requested):

protocol-units/bridge/config/src/lib.rs
Icarus131: @l-monninger , I've committed the requested changes, let me know if that works

0xmovses:
:shipit:

0xmovses (approved):

andygolay:
Code LGTM, thanks @Icarus131.

@l-monninger looks like your changes were addressed. 

One concern: With the `cicd:bridge-containers` label, `bridge-service-build` fails with 

```
Run docker/login-action@v3
Error: Username and password required
```

and `bridge-setup-build` fails with 

```
Error: The process '/usr/bin/git' failed with exit code 1
```

I tried adding `cicd:bridge` and running CI again; more checks fail as expected if `cicd:bridge-containers` fails. Any thoughts on these CI check issues @0xmovses @andyjsbell ?

0xmovses:
Yes sadly, CI is failing, we will need that to pass for this to get merged.

andygolay:
I ran the node locally against this PR's branch and I don't get the errors in https://github.com/movementlabsxyz/movement/pull/824

image

image

What I think would be a good order of operations is we merge #690 and then make sure E2E tests pass against this branch.

It seems like maybe the CI issues I mention above are more CI-related than related to issues with the branch, curious about your thoughts on it @l-monninger @andyjsbell @mzabaluev

PR #723, opened 2024-10-21
Permissioned Blobs

l-monninger:
# Summary
- **RFCs**: [Link to RFC](#./link/to/rfc), [Link to RFC](#./link/to/rfc), or $\emptyset$.
- **Categories**: any of `protocol-units`, `networks`, `scripts`, `util`, `cicd`, or `misc`.

Same as #604 , but signed. 

# Changelog

# Testing

# Outstanding issues

PR #826, opened 2024-11-10
Whitelist for Transaction Signers

l-monninger:
# Summary
- **RFCs**: $\emptyset$

Enables a transaction signers whitelist.
- Adds a whitelist set to the Maptos transaction pipe which will rejected with `TooManyTransactions` if the signer is not whitelisted.
- This restriction on mempool ingress with the DA signer permissions #723, prevents transactions from unauthorized signers from reaching the DA. Insofar as you permissioned proposers respect the same whitelist, whitelisting is consistent. 
- Reads whitelist from lines in a file specified by the `MAPTOS_INGRESS_ACCOUNT_WHITELIST` environment variable or directly in the Maptos Config. 

# Testing
1. e2e tested.

# Outstanding issues
1. There is no filtering of signers at the VM itself. It is presumed the whitelist can be appropriately applied at transaction ingress. 
2. This is in memory. It is suited for a small-ish $O(10^6)$ whitelist (0.032 GB). Scale would benefit from a database, i.e., `RocksDB` set.


protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -71,6 +75,14 @@ impl TransactionPipe {
 				mempool_config.sequence_number_ttl_ms,
 				mempool_config.gc_slot_duration_ms,
 			),
+			whitelisted_accounts: mempool_config.whitelisted_accounts()?,
```
mzabaluev, line 78: Why does this need to be fallible? Wouldn't it be better to have Config correct on construction and `whitelisted_accounts` a simple getter?

protocol-units/execution/maptos/util/src/config/mempool.rs
```diff
@@ -0,0 +1,53 @@
...
+}
+
+impl Config {
+	pub fn whitelisted_accounts(&self) -> Result<Option<HashSet<AccountAddress>>, anyhow::Error> {
+		match &self.ingress_account_whitelist {
+			Some(whitelist_path) => {
+				let mut whitelisted = HashSet::new();
+
+				// read the file from memory
+				let file_string = String::from_utf8(std::fs::read(whitelist_path)?)?;
```
mzabaluev, line 40: This should be done eagerly when config is initialized, not lazily in the getter which may be called multiple times.

mzabaluev (changes requested):

PR #839, opened 2024-11-12
Garbage Collect Inflight

### LINKED ISSUE #834: Garbage Collect Inflight Transactions Count
# Summary
Garbage collect the inflight transactions count to account for fallibility in other systems as described in #829 .

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

#834 

1. Introduces several garbage collected container types: `GcMap`, `GcSet`, `GcCounter`, and `atomic::GcCounter`.
2. Uses `atomic::GcCounter` to resolve #834 leaky-bucket without changing program concurrency model.
3. Includes the other container types as a fallback. The atomic implementation accounts for saturating subs, but makes weak guarantees about decrementing operations. If we want something less subtle in its guarantees, we should wrap one of the former in synchronization primitives. 

# Testing

1. Unit tests added cover collections and integrations: `cargo test -p movement-collections` and `cargo test -p maptos-opt-executor`
2. e2e tests continue to pass; behavior validated manually by checking inflight logs against `howzit`.

# Outstanding issues
1. Working on e2e for collection as secondary objective.

### DIFF
--- a/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
@@ -18,10 +18,9 @@ use crate::gc_account_sequence_number::UsedSequenceNumberPool;
 use futures::channel::mpsc as futures_mpsc;
 use futures::StreamExt;
 use movement_collections::garbage::counted::GcCounter;
-use std::sync::Arc;
+use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
-use tokio::sync::RwLock;
 use tracing::{debug, info, info_span, warn, Instrument};
 
 const GC_INTERVAL: Duration = Duration::from_secs(30);
@@ -121,8 +120,11 @@ impl TransactionPipe {
 			self.used_sequence_number_pool.gc(epoch_ms_now);
 
 			// garbage collect the transactions in flight
-			let mut transactions_in_flight = self.transactions_in_flight.write().await;
-			transactions_in_flight.gc(epoch_ms_now);
+			{
+				// unwrap because failure indicates poisoned lock
+				let mut transactions_in_flight = self.transactions_in_flight.write().unwrap();
+				transactions_in_flight.gc(epoch_ms_now);
+			}
 
 			// garbage collect the core mempool
 			self.core_mempool.gc();
@@ -195,7 +197,7 @@ impl TransactionPipe {
 	) -> Result<SubmissionStatus, Error> {
 		// For now, we are going to consider a transaction in flight until it exits the mempool and is sent to the DA as is indicated by WriteBatch.
 		let in_flight = {
-			let transactions_in_flight = self.transactions_in_flight.read().await;
+			let transactions_in_flight = self.transactions_in_flight.read().unwrap();
 			transactions_in_flight.get_count()
 		};
 		info!(
@@ -258,7 +260,7 @@ impl TransactionPipe {
 					.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
 				// increment transactions in flight
 				{
-					let mut transactions_in_flight = self.transactions_in_flight.write().await;
+					let mut transactions_in_flight = self.transactions_in_flight.write().unwrap();
 					transactions_in_flight.increment(now, 1);
 				}
 				self.core_mempool.commit_transaction(&sender, sequence_number);
--- a/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
@@ -17,10 +17,11 @@ use aptos_vm_validator::vm_validator::{self, TransactionValidation, VMValidator}
 use crate::gc_account_sequence_number::UsedSequenceNumberPool;
 use futures::channel::mpsc as futures_mpsc;
 use futures::StreamExt;
-use movement_collections::garbage::atomic::counted::GcCounter;
-use std::sync::{atomic::AtomicU64, Arc};
+use movement_collections::garbage::counted::GcCounter;
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
+use tokio::sync::RwLock;
 use tracing::{debug, info, info_span, warn, Instrument};
 
 const GC_INTERVAL: Duration = Duration::from_secs(30);
@@ -36,9 +37,9 @@ pub struct TransactionPipe {
 	// State of the Aptos mempool
 	core_mempool: CoreMempool,
 	// Shared reference on the counter of transactions in flight.
-	transactions_in_flight: GcCounter,
+	transactions_in_flight: Arc<RwLock<GcCounter>>,
 	// The configured limit on transactions in flight
-	in_flight_limit: u64,
+	in_flight_limit: Option<u64>,
 	// Timestamp of the last garbage collection
 	last_gc: Instant,
 	// The pool of used sequence numbers
@@ -57,8 +58,8 @@ impl TransactionPipe {
 		db_reader: Arc<dyn DbReader>,
 		node_config: &NodeConfig,
 		mempool_config: &MempoolConfig,
-		transactions_in_flight: GcCounter,
-		transactions_in_flight_limit: u64,
+		transactions_in_flight: Arc<RwLock<GcCounter>>,
+		transactions_in_flight_limit: Option<u64>,
 	) -> Self {
 		TransactionPipe {
 			mempool_client_receiver,
@@ -120,7 +121,8 @@ impl TransactionPipe {
 			self.used_sequence_number_pool.gc(epoch_ms_now);
 
 			// garbage collect the transactions in flight
-			self.transactions_in_flight.gc(epoch_ms_now);
+			let mut transactions_in_flight = self.transactions_in_flight.write().await;
+			transactions_in_flight.gc(epoch_ms_now);
 
 			// garbage collect the core mempool
 			self.core_mempool.gc();
@@ -192,19 +194,24 @@ impl TransactionPipe {
 		transaction: SignedTransaction,
 	) -> Result<SubmissionStatus, Error> {
 		// For now, we are going to consider a transaction in flight until it exits the mempool and is sent to the DA as is indicated by WriteBatch.
-		let in_flight = self.transactions_in_flight.get_count();
+		let in_flight = {
+			let transactions_in_flight = self.transactions_in_flight.read().await;
+			transactions_in_flight.get_count()
+		};
 		info!(
 			target: "movement_timing",
 			in_flight = %in_flight,
 			"transactions_in_flight"
 		);
-		if in_flight > self.in_flight_limit {
-			info!(
-				target: "movement_timing",
-				"shedding_load"
-			);
-			let status = MempoolStatus::new(MempoolStatusCode::MempoolIsFull);
-			return Ok((status, None));
+		if let Some(inflight_limit) = self.in_flight_limit {
+			if in_flight >= inflight_limit {
+				info!(
+					target: "movement_timing",
+					"shedding_load"
+				);
+				let status = MempoolStatus::new(MempoolStatusCode::MempoolIsFull);
+				return Ok((status, None));
+			}
 		}
 
 		// Pre-execute Tx to validate its content.
@@ -250,7 +257,10 @@ impl TransactionPipe {
 					.await
 					.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
 				// increment transactions in flight
-				self.transactions_in_flight.increment(now, 1);
+				{
+					let mut transactions_in_flight = self.transactions_in_flight.write().await;
+					transactions_in_flight.increment(now, 1);
+				}
 				self.core_mempool.commit_transaction(&sender, sequence_number);
 
 				// update the used sequence number pool
--- a/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
@@ -115,8 +115,16 @@ impl TransactionPipe {
 			// todo: these will be slightly off, but gc does not need to be exact
 			let now = Instant::now();
 			let epoch_ms_now = chrono::Utc::now().timestamp_millis() as u64;
+
+			// garbage collect the used sequence number pool
 			self.used_sequence_number_pool.gc(epoch_ms_now);
+
+			// garbage collect the transactions in flight
+			self.transactions_in_flight.gc(epoch_ms_now);
+
+			// garbage collect the core mempool
 			self.core_mempool.gc();
+
 			self.last_gc = now;
 		}
 
@@ -242,7 +250,7 @@ impl TransactionPipe {
 					.await
 					.map_err(|e| anyhow::anyhow!("Error sending transaction: {:?}", e))?;
 				// increment transactions in flight
-				self.transactions_in_flight.increment(now);
+				self.transactions_in_flight.increment(now, 1);
 				self.core_mempool.commit_transaction(&sender, sequence_number);
 
 				// update the used sequence number pool
--- a/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
+++ b/protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
@@ -17,6 +17,7 @@ use aptos_vm_validator::vm_validator::{self, TransactionValidation, VMValidator}
 use crate::gc_account_sequence_number::UsedSequenceNumberPool;
 use futures::channel::mpsc as futures_mpsc;
 use futures::StreamExt;
+use movement_collections::garbage::atomic::counted::GcCounter;
 use std::sync::{atomic::AtomicU64, Arc};
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
@@ -35,7 +36,7 @@ pub struct TransactionPipe {
 	// State of the Aptos mempool
 	core_mempool: CoreMempool,
 	// Shared reference on the counter of transactions in flight.
-	transactions_in_flight: Arc<AtomicU64>,
+	transactions_in_flight: GcCounter,
 	// The configured limit on transactions in flight
 	in_flight_limit: u64,
 	// Timestamp of the last garbage collection
@@ -56,7 +57,7 @@ impl TransactionPipe {
 		db_reader: Arc<dyn DbReader>,
 		node_config: &NodeConfig,
 		mempool_config: &MempoolConfig,
-		transactions_in_flight: Arc<AtomicU64>,
+		transactions_in_flight: GcCounter,
 		transactions_in_flight_limit: u64,
 	) -> Self {
 		TransactionPipe {
@@ -183,7 +184,7 @@ impl TransactionPipe {
 		transaction: SignedTransaction,
 	) -> Result<SubmissionStatus, Error> {
 		// For now, we are going to consider a transaction in flight until it exits the mempool and is sent to the DA as is indicated by WriteBatch.
-		let in_flight = self.transactions_in_flight.load(std::sync::atomic::Ordering::Relaxed);
+		let in_flight = self.transactions_in_flight.get_count();
 		info!(
 			target: "movement_timing",
 			in_flight = %in_flight,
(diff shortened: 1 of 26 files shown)


util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,84 @@
...
+	K: Eq + Hash,
+	V: Eq + Hash,
+{
+	/// Creates a new GcMap with a specified garbage collection slot duration.
+	pub fn new(value_ttl_ms: u64, gc_slot_duration_ms: u64) -> Self {
+		GcMap { value_ttl_ms, gc_slot_duration_ms, value_lifetimes: BTreeMap::new() }
+	}
+
+	/// Gets a sequence number for an account
+	pub fn get_value(&self, account: &K) -> Option<&V> {
```
andyjsbell, line 29:
```suggestion
pub fn get_value(&self, account: &K) -> Option<&V> {
    for lifetimes in self.value_lifetimes.values().rev() {
        if let Some(value) = lifetimes.get(account) {
            return Some(value);
        }
    }
    None
}
```

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,84 @@
...
+		}
+	}
+
+	/// Sets the sequence number for an account.
+	pub fn set_value(&mut self, key: K, value: V, current_time_ms: u64) {
+		// remove the old sequence number
+		self.remove_value(&key);
+
+		// compute the slot for the new lifetime and add accordingly
+		let slot = current_time_ms / self.gc_slot_duration_ms;
```
andyjsbell, line 61: div 0 !!

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,84 @@
...
+
+		// add the new sequence number
+		self.value_lifetimes.entry(slot).or_insert_with(HashMap::new).insert(key, value);
+	}
+
+	/// Garbage collects sequence numbers that have expired.
+	/// This should be called periodically.
+	pub fn gc(&mut self, current_time_ms: u64) {
+		let gc_slot = current_time_ms / self.gc_slot_duration_ms;
```
andyjsbell, line 70: div 0 !!

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,84 @@
...
+
+		// remove all slots that are too old
+		let slot_cutoff = gc_slot - self.value_ttl_ms / self.gc_slot_duration_ms;
```
andyjsbell, line 73: div 0 !! and some underflow for flavour

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,84 @@
...
+		let slots_to_remove: Vec<u64> = self
+			.value_lifetimes
+			.keys()
+			.take_while(|slot| **slot < slot_cutoff)
+			.cloned()
+			.collect();
+		for slot in slots_to_remove {
+			self.value_lifetimes.remove(&slot);
+		}
```
andyjsbell, line 82: Could we maybe use `drain` here

util/collections/src/garbage/set.rs
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	/// Garbage collects sequence numbers that have expired.
+	/// This should be called periodically.
+	pub fn gc(&mut self, current_time_ms: u64) {
+		let gc_slot = current_time_ms / self.gc_slot_duration_ms;
+
+		// remove all slots that are too old
+		let slot_cutoff = gc_slot - self.value_ttl_ms / self.gc_slot_duration_ms;
+		let slots_to_remove: Vec<u64> = self
```
andyjsbell, line 53: drain

andyjsbell (changes requested):
Let's not panic if we can help it.  Maybe some efficiency with `drain`.  Not clear what the objective is from the description nor how this is tested.

util/collections/src/garbage/set.rs
l-monninger: Okay, there you go.

util/collections/src/garbage/set.rs
```diff
@@ -0,0 +1,63 @@
...
+		}
+	}
+
+	/// Sets the sequence number for an account.
+	pub fn insert(&mut self, value: V, current_time_ms: u64) {
+		// remove the old sequence number
+		self.remove_value(&value);
+
+		// compute the slot for the new lifetime and add accordingly
+		let slot = current_time_ms / self.gc_slot_duration_ms;
```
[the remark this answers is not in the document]
l-monninger: I rewrapped the types. But, in general, I don't agree that allowing panic on div 0 is incorrect. You are usually going to have to panic one way or another if you encounter this and for cases like this the failure is fast.

util/collections/README.md
```diff
@@ -0,0 +1,30 @@
+# `buildtime`
+The `buildtime` crate provides useful functions and macros for use in `build.rs` scripts.
+
+## `workspace`
+You can use `cargo_workspace()` or `cargo_workspace!()` to get the path of the workspace root.
```
mzabaluev, line 5: This README does not seem like it belongs to this crate.

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,131 @@
...
+	K: Eq + Hash + Debug,
+	V: Eq + Hash + Debug,
+{
+	/// Creates a new GcMap with a specified garbage collection slot duration.
+	pub fn new(value_ttl: Duration, gc_slot_duration: Duration) -> Self {
+		GcMap { value_ttl, gc_slot_duration, value_lifetimes: BTreeMap::new() }
+	}
+
+	/// Gets a value for a key
+	pub fn get_value(&self, key: &K) -> Option<&V> {
```
andyjsbell, line 31: This can be simplified as
```rust
for lifetimes in self.value_lifetimes.values().rev() {
        if let Some(value) = lifetimes.get(key) {
            // Return immediately if a valid value is found
            return Some(value);
        }
    }
    None
````

File: util/collections/src/lib.rs [hunk 10]
```diff
@@ -0,0 +1 @@
+pub mod garbage;
```
mzabaluev, line 1: It would be nice to add a crate-level comment explaining what it is about.
Like, what collections could we provide here, other than garbage collections? 😁

util/collections/src/garbage/set.rs
```diff
@@ -0,0 +1,108 @@
+use crate::garbage::Duration;
+use std::collections::{BTreeMap, HashSet};
+use std::hash::Hash;
+
+pub struct GcSet<V>
```
mzabaluev, line 5: This type lacks documentation explaining its purpose.

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,131 @@
...
+				}
+				None => {}
+			}
+		}
+
+		None
+	}
+
+	/// Removes the value for an key.
+	pub fn remove_value(&mut self, key: &K) {
```
andyjsbell, line 48: could this return a `bool` instead?

util/collections/src/garbage/mod.rs
```diff
@@ -0,0 +1,20 @@
+pub mod atomic;
+pub mod counted;
+pub mod map;
+pub mod set;
+use std::num::NonZeroU64;
+
+#[derive(Debug, Clone)]
+pub struct Duration(NonZeroU64);
```
mzabaluev, line 8: Does `garbage::Duration` provide added value over a more widely used domain type like `std::time::Duration`?

andyjsbell (approved):

util/collections/src/garbage/map.rs
```diff
@@ -0,0 +1,131 @@
+use crate::garbage::Duration;
+use std::collections::{BTreeMap, HashMap};
+use std::fmt::Debug;
+use std::hash::Hash;
+
+pub struct GcMap<K, V>
```
mzabaluev, line 6: I think I get the idea, but documentation would be nice.

util/collections/src/garbage/counted.rs
```diff
@@ -0,0 +1,108 @@
+use crate::garbage::Duration;
+use std::collections::BTreeMap;
+
+pub struct GcCounter {
```
mzabaluev, line 4: Also needs documentation about the general idea of this type.

mzabaluev (changes requested):

util/collections/src/garbage/map.rs
l-monninger: Sure

util/collections/src/garbage/mod.rs
l-monninger: It is specifically to restrict the domain to non-zero while remaining a very simple `u64` wrapper.

util/collections/src/lib.rs
l-monninger: Okay

util/collections/src/garbage/map.rs
```diff
@@ -32,26 +33,24 @@ where
...
-					// check if the value is still valid
-					return Some(value);
-				}
-				None => {}
+			if let Some(value) = lifetimes.get(key) {
+				return Some(value);
 			}
 		}
 
 		None
```
mzabaluev, line 41: This can be written in functional style:

```rust
self.value_lifetimes.values().rev().find_map(|lifetimes| lifetimes.get(key))
```

mzabaluev (approved):

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -0,0 +1,159 @@
...
+		// Atomically check and set the timestamp if it doesn't match current_time
+		let active_slot = active_slot_timestamp.load(Ordering::Relaxed);
+		if active_slot == slot {
+			// Same timestamp, increment count
+			count.fetch_add(amount, Ordering::SeqCst);
+		} else {
+			// Different timestamp, reset slot for new period
+			active_slot_timestamp.store(slot_timestamp, Ordering::SeqCst);
+			count.store(amount, Ordering::SeqCst);
+		}
```
mzabaluev, line 73: I think there is a race condition: if two threads call `increment` concurrently at the same timestamp, and the slot has not been updated, they will both set each own `amount` and one of them gets overwritten.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -0,0 +1,159 @@
...
+	/// This should be called periodically.
+	pub fn gc(&self, current_time: u64) {
+		let cutoff_time = current_time - self.value_ttl.get();
+
+		for (slot_timestamp, count) in self.value_lifetimes.iter() {
+			// If the timestamp is older than the cutoff, reset the slot
+			if slot_timestamp.load(Ordering::Relaxed) <= cutoff_time {
+				slot_timestamp.store(0, Ordering::SeqCst);
+				count.store(0, Ordering::SeqCst);
+			}
```
mzabaluev, line 94: Similarly, another thread could be incrementing concurrently and this will clobber the new count set by that thread.

l-monninger: Same as the above. There's not a great way around this without a synchronization primitive.

util/collections/src/garbage/atomic/counted.rs
l-monninger: In other words, these should always set the same slot where we would have a race. Update would provide a weak guarantee in the same way that decrement is handled.

mzabaluev: > Perhaps we can check that the slot pushed is not less than the other slot?

Slots are modulo the vector length, perhaps you mean the timestamps. This is not good enough because you want to check if the value has not been changed between the load and the store as well. Maybe you need to use `compare_exchange_weak` instead.

l-monninger: `fetch_update` is more complete...

l-monninger: But, it doesn't quite work here.

l-monninger: I think I may have it.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -61,15 +61,36 @@ impl GcCounter {
...
+				if !count
+					.compare_exchange(active_amount, amount, Ordering::SeqCst, Ordering::Relaxed)
+					.is_ok()
+				{
+					// otherwise fetch add
+					count.fetch_add(amount, Ordering::SeqCst);
+				}
+			} else {
+				// If another thread updated the timestamp, the amount will also have been reset, so just go ahead and add
+				count.fetch_add(amount, Ordering::SeqCst);
```
mzabaluev, line 92: So the assumption here is that another thread has stored exactly the same `slot_timestamp` value into `active_slot_timestamp`. I guess this holds, unless a thread has been suspended for so long that the slot wrapped around. Which might happen e.g. if the machine was suspended for some reason, or someone's running a node on a laptop or similar.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -61,15 +61,36 @@ impl GcCounter {
 		let slot = slot_timestamp % self.num_slots;
 		let (active_slot_timestamp, count) = &self.value_lifetimes[slot as usize];
 
-		// Atomically check and set the timestamp if it doesn't match current_time
+		// Atomically check and set the timestamp if it doesn't match the current slot
 		let active_slot = active_slot_timestamp.load(Ordering::Relaxed);
+		let active_amount = count.load(Ordering::Relaxed);
```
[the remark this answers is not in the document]
l-monninger: What do you mean it's wasted? 

I had in mind here that correct usage would mean the slot can only be equal to or greater than the current slot by the time we store, which would mean that the `fetch_add` is at worst adding to a later slot. I think we can accept that behavior.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -61,15 +61,36 @@ impl GcCounter {
...
+			// Unless gc windows are very small.
 			count.fetch_add(amount, Ordering::SeqCst);
 		} else {
-			// Different timestamp, reset slot for new period
-			active_slot_timestamp.store(slot_timestamp, Ordering::SeqCst);
-			count.store(amount, Ordering::SeqCst);
+			// now we will zero out the slot and add the new value
+			// Use compare_exchange to safely reset the slot and count only
+			if active_slot_timestamp
+				.compare_exchange(active_slot, slot_timestamp, Ordering::SeqCst, Ordering::Relaxed)
```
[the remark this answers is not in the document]
l-monninger:
```rust
compare_exchange(active_slot, slot_timestamp, Ordering::SeqCst, Ordering::Relaxed)
```

We want total ordering on the successful operation as we are under the hood performing a store operation that can clobber the other values. We don't want to allow for any disagreement about the order of these operations. 

The relaxed ordering on the failure case is because we actual don't use the failure case [load](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicU64.html). If the comparison fails, we just decide we're going to add to the value. So, we don't need to impose any synchronization here.

util/collections/src/garbage/atomic/counted.rs
l-monninger: Yeah, so I read into this a bit more. Practically, atomic operations may involve invalidating cache lines on different cores. It's not quite such a nice use of an ISA's atomic instructions as I originally expected. 

So, we should expect synchronization to be on the order of nanoseconds. If a slot were indeed this small, we would get this wrap around. In that case, we're adding to a later slot. 

The problem would be the count could keep increasing without ever being collected under this condition. I'm going to scratch my head a little more to see if there's something clever. But, I think we may just have to say such a small slot size is not the intended usage.

l-monninger: So, I think we should either leave this as is but increase the size of the buffer.

1. No matter how we play with these primitives, we're not going to have complete control over the critical section. If contended paths lead to `fetch_add` then we can increment past what was intended. If contended paths lead to `exchange`, then we will miss counts. In the current implementation, 
2. However, currently, we would only start to `gc` when we loop back around, which is also when we would start adding values again. We would want to actually make sure, with proper usage, we are leaving a gap to garbage collect on some factor of the `slot_duration`, the minimum being `1*slot_duration`. This means our buffer should, in fact, be strictly greater than `value_ttl.get() / gc_slot_duration.get();` and we kind of need to play with the mapping. We might instead want a notion of a relative "next" slot in the ring buffer instead of an absolute mapping. Or, we could use a double-length ring buffer.

mzabaluev: What I mean is, the OS does not guarantee a thread to be scheduled with any minimal time interval between run slots. So it's conceivable that a thread would run fetch_add on line 92 at a completely different time from another thread that has overwritten the same slot.

I think a way to be bulletproof here has to do with trying `compare_exchange` in a loop until the operation overwrites an expected value.

protocol-units/execution/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -120,7 +121,8 @@ impl TransactionPipe {
 			self.used_sequence_number_pool.gc(epoch_ms_now);
 
 			// garbage collect the transactions in flight
-			self.transactions_in_flight.gc(epoch_ms_now);
+			let mut transactions_in_flight = self.transactions_in_flight.write().await;
+			transactions_in_flight.gc(epoch_ms_now);
```
mzabaluev, line 125: No need to hold this lock for the rest of the function:

```suggestion
			{
			    let mut transactions_in_flight = self.transactions_in_flight.write().await;
			    transactions_in_flight.gc(epoch_ms_now);
			}
```

protocol-units/execution/opt-executor/src/executor/mod.rs
```diff
@@ -41,15 +41,16 @@ impl Executor {
...
+		let mut transactions_in_flight = self.transactions_in_flight.write().await;
+		let current = transactions_in_flight.get_count();
 		info!(
 			target: "movement_timing",
 			count,
 			current,
 			"decrementing_transactions_in_flight",
 		);
-		self.transactions_in_flight.decrement(count);
+		transactions_in_flight.decrement(count);
```
mzabaluev, line 53: As I mentioned in Slack, there is nothing that needs to hold the mutex lock across await points. So the mutex could be `std::sync::Mutex`, which is better for performance.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -58,43 +55,36 @@ impl GcCounter {
 	/// Increments the value in a specific slot.
 	pub fn increment(&self, current_time: u64, amount: u64) {
 		let slot_timestamp = current_time / self.gc_slot_duration.get();
-		let slot = slot_timestamp % self.num_slots;
+		let pass_number = slot_timestamp / (self.num_slots / 2); // Determine if we're in the first or second pass
```
mzabaluev, line 58: If `self.num_slots` is odd, we could get a small "third pass" here, where `pass_number % 2 == 0`, but it will quickly cycle back to the first pass that will go to the same slot offset.

util/collections/src/garbage/atomic/counted.rs
```diff
@@ -58,43 +55,36 @@ impl GcCounter {
...
+		let base_slot = slot_timestamp % (self.num_slots / 2); // Calculate the base slot index within the first pass
+
+		// Calculate the actual slot index by adding the pass offset (0 or 1 times num_slots / 2)
+		let slot = base_slot + (pass_number % 2) * (self.num_slots / 2);
```
mzabaluev, line 62: I don't think this fundamentally solves the problem with `atomic::GcCounter`, though. A thread could still be scheduled off for two (or more) passes and contest the same slot with a different timestamp.

l-monninger: Yes, but correct usage would mean this should not happen in a way that continues to sum. That is, this should have been garbage collected.

protocol-units/execution/opt-executor/src/executor/mod.rs
l-monninger: Okay, I will introduce this at the risk of someone later modifying, at which point I will link this comment.

protocol-units/execution/opt-executor/src/executor/mod.rs
```diff
@@ -40,20 +40,17 @@ impl Executor {
...
-				info!(
-					target: "movement_timing",
-					count,
-					current,
-					"decrementing_transactions_in_flight",
-				);
-				Some(current.saturating_sub(count))
-			})
-			.unwrap_or_else(|_| 0);
+	pub async fn decrement_transactions_in_flight(&self, count: u64) {
```
mzabaluev, line 43: This method, and the `decrement_transaction_in_flight` method upwards of it, does not need to be async any more.

mzabaluev (commented):
The functionality that is actually used looks good, aside from unneeded asyncness.

I'm not sure about leaving `atomic::GcCounter` in with the known problems, though.
Maybe feature-gate it?

PR #845, opened 2024-11-12
Rename Atomic Bridge to Native Bridge

Primata:
# Summary

After a consensus, bridge has to be renamed to native bridge. This implements all changes required to have the protocol relay on the new naming.

# Changelog

AtomicBridge -> NativeBridge
atomicBridge -> nativeBridge
Atomic Bridge -> Native Bridge
atomic bridge -> native bridge
atomic_bridge -> native_bridge

# Testing

cd /Users/primata/movement/protocol-units/bridge/contracts
forge test

# Outstanding issues
change on aptos-core
integration might be still an issue


franck44:
Why would we rename the bridge components?

We have provided atomic bridges components (contracts) and in our setting they are **used** to implement a bridge for MOVE (instantiation that we may call native).
We may re-use the same components to build an ETH (or USDT) bridge, which is probably not a native bridge.
[BTW I'd like to know precisely what "native" is supposed to mean here.]

What is important is the _atomicity_, the fact that there are two counterparts in the contracts (one initiator, one counterparty).
 
I suggest that in the code (Solidity, Move) we keep the names as they were before, using atomic bridge rather than native.
The documentation may be updated with the term native where it is adequate.

For example, in the aptos_framework, the `Coin` or (`FA`) is not named `NativeCoin`, it is a generic template to deploy coins. 
Same applies for the atomic bridge.

franck44:
One more thing:

> After a consensus, bridge has to be renamed to native bridge. This implements all changes required to have the protocol relay on the new naming.

Who was involved in the consensus?

apenzk:
@0xPrimata i changed the changelog in the head comment. there was some inconsistency.. please double check

Issue #827, opened 2024-11-10
[Bridge] [Solidity] Use `requires` for input validation rather than `if ... revert`


franck44:
# Problem 

## Using `requires` to validate inputs
In our (bridge) Solidity contracts, we are using the pattern `if (!cond) revert Error` to validate inputs. This is not the best practice and there were several instances of bugs in the past where the contract had `if !C someCode; revert()`, ommitting the `{}` around the `if-body`, which can have unintended consequences. You can read more about the risks (and a bug in APPLE-SSL) [here](https://github.com/ethereum/solidity/issues/13312#issuecomment-1752316315).

> [!WARNING] 
> In short, if written by hand, there is no guarantee that the `if-revert` pattern is used properly.

Using `requires(cond, error)` has several advantages:
- `requires` clearly identify the condition `cond` that should be checked rather than the negation of the condition that appears in the equivalent `if-revert` pattern;
- `requires(cond, error)` is automatically translated (by the Solidity compiler)  in `if (!cond) revert error` thereby enforcing that the pattern `if-revert` is used correctly.

> [!IMPORTANT] 
> The use of `requires` is recommended to validate inputs and has **better readability and security**.

## `CustomError` to minimise gas cost

There is a caveat though: until recently, `requires(cond, error)` only supported `error` of type int or strings, and these two types are more expensive gas-wise than _custom errors_. 
This is why developers would deliberately use `if (!C) revert CustomError` (instead of the `requires`) in the code to optimise gas costs, while compromising security.

As of version 0.8.27 ([release notes](https://soliditylang.org/blog/2024/09/04/solidity-0.8.27-release-announcement/)) , the Solidity compiler supports `requires(conditional, customError)`.

> [!TIP] 
> We can now use a **safe** and **gas-efficient** pattern `requires(cond, customError)`.

# Proposed changes  

There are two Solidity contracts that are currently using the unsafe and hard-to-read `if-revert` pattern to validate inputs:

- [AtomicBridgeInitiatorMOVE.sol](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol)
- [AtomicBridgeCounterpartyMOVE.sol](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/bridge/contracts/src/AtomicBridgeCounterpartyMOVE.sol)

The **proposal is to use `requires` for input validation**. 

[...]

PR #852, opened 2024-11-13
Bridge contracts: implement require(cond, error) pattern and update to solc 0.8.27

andygolay:
# Summary
- Addresses #827 
- **Categories**: `bridge` , `contracts`

# Changelog

Update `AtomicBridgeInitiatorMOVE.sol` and AtomicBridgeCounterparty.sol` to use the require -> custom error pattern.

# Testing

Solidity tests pass with `forge test --fork-url https://ethereum-sepolia-rpc.publicnode.com -vv` in `protocol-units/bridge/contracts`

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
+++ b/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
@@ -69,7 +69,9 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
         require(moveAmount > 0, ZeroAmount());
 
         // Transfer the MOVE tokens from the user to the contract
-        require(moveToken.transferFrom(originator, address(this), moveAmount), MOVETransferFailed());
+        if (!moveToken.transferFrom(originator, address(this), moveAmount)) {
+            revert MOVETransferFailed();
+        }
 
         // Update the pool balance
         poolBalance += moveAmount;
@@ -110,7 +112,7 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
         
         // Decrease pool balance and transfer MOVE tokens back to the originator
         poolBalance -= bridgeTransfer.amount;
-        require(moveToken.transfer(bridgeTransfer.originator, bridgeTransfer.amount), MOVETransferFailed());
+        if (!moveToken.transfer(bridgeTransfer.originator, bridgeTransfer.amount)) revert MOVETransferFailed();
 
         emit BridgeTransferRefunded(bridgeTransferId);
     }
@@ -120,7 +122,7 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
         require(poolBalance >= amount, InsufficientMOVEBalance());
 
         poolBalance -= amount;
-        require(moveToken.transfer(recipient, amount), MOVETransferFailed());
+        if (!moveToken.transfer(recipient, amount)) revert MOVETransferFailed();
     }
 }
 
--- a/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
+++ b/protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity ^0.8.22;
+pragma solidity ^0.8.27;
 
 import {IAtomicBridgeInitiatorMOVE} from "./IAtomicBridgeInitiatorMOVE.sol";
 import {MockMOVEToken} from "./MockMOVEToken.sol";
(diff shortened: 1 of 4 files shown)


protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
```diff
@@ -67,14 +66,10 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
         address originator = msg.sender;
 
         // Ensure there is a valid amount
-        if (moveAmount == 0) {
-            revert ZeroAmount();
-        }
+        require(moveAmount > 0, ZeroAmount());
 
         // Transfer the MOVE tokens from the user to the contract
-        if (!moveToken.transferFrom(originator, address(this), moveAmount)) {
```
franck44, line 75: @andygolay As I explained in the issue, the scope of this change is _input validation_.
This `revert()` is not related to input validation, so it may not be changed to a `require`.

protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
```diff
@@ -97,32 +92,35 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
...
-        if (bridgeTransfer.state != MessageState.INITIALIZED) revert BridgeTransferStateNotInitialized();
-        if (block.timestamp < bridgeTransfer.timeLock) revert TimeLockNotExpired();
+        require(bridgeTransfer.state == MessageState.INITIALIZED, BridgeTransferStateNotInitialized());
+        require(block.timestamp >= bridgeTransfer.timeLock, TimeLockNotExpired());
+
         bridgeTransfer.state = MessageState.REFUNDED;
         
         // Decrease pool balance and transfer MOVE tokens back to the originator
         poolBalance -= bridgeTransfer.amount;
-        if (!moveToken.transfer(bridgeTransfer.originator, bridgeTransfer.amount)) revert MOVETransferFailed();
```
franck44, line 116: This is not related to input validation, so may not be modified to use a `requires`.

protocol-units/bridge/contracts/src/AtomicBridgeInitiatorMOVE.sol
```diff
@@ -97,32 +92,35 @@ contract AtomicBridgeInitiatorMOVE is IAtomicBridgeInitiatorMOVE, OwnableUpgrade
...
     }
 
     function withdrawMOVE(address recipient, uint256 amount) external {
-        if (msg.sender != counterpartyAddress) revert Unauthorized();
-        if (poolBalance < amount) revert InsufficientMOVEBalance();
+        require(msg.sender == counterpartyAddress, Unauthorized());
+        require(poolBalance >= amount, InsufficientMOVEBalance());
+
         poolBalance -= amount;
-        if (!moveToken.transfer(recipient, amount)) revert MOVETransferFailed();
```
franck44, line 125: This is not input validation, so may bot be changed to a `requires`.

franck44 (changes requested):
The scope of the issue is tio fix _input validation_, and use the `requires` pattern for input validation.

Some of the proposed changes impact code that is not _input validation_ and should not be changed to use `requires`.

andygolay:
> The scope of the issue is to fix _input validation_, and use the `requires` pattern for input validation.
> 
> Some of the proposed changes impact code that is not _input validation_ and should not be changed to use `requires`.

Thank you @franck44, I made the requested changes.

franck44 (approved):
Thanks @andygolay !

0xPrimata:
@franck44 I believe that if it's not a conditional logic flow, require should be used for consistency.
I think @andygolay could go back to using `require` on transfer checks.

andygolay:
> @franck44 I believe that if it's not a conditional logic flow, require should be used for consistency. I think @andygolay could go back to using `require` on transfer checks.

Whatever you guys decide is okay with me; I did look it up and it seems like require is technically fine to use throughout, but it's conventionally used for input values. I wonder if that's because require used to be less gas efficient? And maybe now conventions can shift to using require in general? It does strike me as more normalized syntax.

The first commit is with all require, the second is with require on input values only. I'll defer to you Solidity gurus, as I don't have a strong opinion about.

franck44:
@0xPrimata 

> I believe that if it's not a conditional logic flow, require should be used for consistency.
> I think @andygolay could go back to using require on transfer checks.

No, that's not best practice. I provided lots of documentation in [issue 827](https://github.com/movementlabsxyz/movement/issues/827), so we are going to follow best practice.

`requires` to validate inputs, and `reverts` for more complex errors.

We can merge this PR and close the issue.

franck44:
@andygolay 

I forgot to mention that there is another reason why `requires` are used for input validation but not for other revert conditions: `requires` act as _pre-conditions_ i.e. conditions under which the function can be called. The `requires` are used by [SMTChecker](https://docs.soliditylang.org/en/latest/smtchecker.html) in the Solidity compiler as specifications to _verify_ some properties of the functions.

Issue #847, opened 2024-11-13
Remove "M1" References on mainnet/testnet codebase


rolandoesparza:
**Background**
M1 has been used previously to describe our Shared Sequencer as well as our legacy Avalanche subnet. We should remove references to M1 in the code for our testnets/mainnet.

mzabaluev:
What should be the new name for m1-da-light-node?

Issue #417, opened 2024-08-21
Reorganize Suzuka e2e tests


mzabaluev:
Discussion: #395

We should have a dedicated, clearly named crate for the E2E tests. Going with the current structure of end projects for Suzuka hosted, under `./networks/suzuka`, the E2E tests for Suzuka can be a crate under `./networks/suzuka/e2e-tests`, named `suzuka-e2e-tests`.

Renaming the `suzuka-client` crate can be a good start, since this crate currently provides no other functionality, other than re-exported names from `aptos-sdk`.

The tests should be organized under a uniform structure plugging into the `cargo test` harness: either integration-style test binaries under `./tests`, or modules under `./src/`, but no hodgepodge of non-enumerable binaries and unit tests.

The load and soak test cases should be grouped so as to be easily invoked as the respective group. Alternatives:
* Top-level modules in the library crate, giving `load::*` and `soak::*`.
* Dedicated test binaries: `./tests/load-tests` and `./tests/soak-tests`.

mzabaluev:
@l-monninger Do you think this needs to be further broken down to more specific test targets?

PR #863, opened 2024-11-18
Change Suzuka naming to Movement, m1 to movement-celestia

### LINKED ISSUE #848: Rename suzuka-full-node
Background:
We want to remove reference to "suzuka" on suzuka-full-node as we have testnets with different names and infra partners are getting confused when deploying on Bardock testnet.

Options:
1) Rename suzuka-full-node to testnet-full-node
2) Rename suzuka-full-node to full-node
2) Make a copy of suzuka-full-node and rename to bardock-full-node

Priority: Urgent (P0)

mzabaluev:
Resolves #848, #847.

# Summary
- **Categories**: `networks`, `scripts`, `util`, `cicd`, `misc`.

Rename most of suzuka branding in sources, documentation, scripts, manifests and elsewhere.
The name of the network is Movement.
In particular, rename suzuka-client to movement-tests because everything in that crate is about testing.
This goes towards resolving #417.

# Changelog

Renamed `suzuka-*` binaries and crates to `movement-*`.
Renamed `m1-da-light-node*` binaries and crates to `movement-celestia-da-light-node*`.

# Testing

Existing CI tests should pass (with trigger labels renamed acccordingly).

# Outstanding issues

While we are at it, it would to be good to rename the DA light node as per #847.


CODEOWNERS
```diff
@@ -5,7 +5,7 @@
 protocol-units/execution/* @l-monninger @0xmovses @musitdev @mzabaluev
 
 # m1-da
-protocol-units/da/m1/* @l-monninger @mzabaluev
+protocol-units/da/celestia/* @l-monninger @mzabaluev
```
l-monninger, line 8: Ah, I wouldn't do this. Keep it named as the generic abstraction it is. If you need another name from business, ask @rolandoesparza . I would suggest `movement-da` for simplicity.

l-monninger (changes requested):

CODEOWNERS
mzabaluev: There's now `da/grpc` that covers the gRPC related components and this is not specific to a DA layer.
I thought of `da/celestia` as the specific implementation of the DA light node service that works with Celestia. Is there something wrong with this layout and naming?

l-monninger: The abstractions in those protocol units are not purely Celestia though. If anything, I would argue that `/celestia` should be a subdir of those protocol units.

l-monninger:
I think this looks good. So, we'll just wait on #873 I guess.

CODEOWNERS
l-monninger: Sorry, I misread https://github.com/movementlabsxyz/movement/pull/863/commits/320d74b86076b66271e1cac14e96f8974f735b16

I would go for `da/movement`. If you want to have subdirs for just the `grpc` units then `da/movement/grpc`, but I would say `grpc` is providing the abstracted protocol. 

For Celestia, I would either leave it as is, i.e., just being distinguished by structs and modules within the crates, or I would entertain a refactor to move the Celestia implementations of various traits out into their own crates. In this latter case, `da/movement/celestia` would make sense to me.

l-monninger (changes requested):
Still needs fixes.

mzabaluev:
> Still needs fixes.

Pushed the fix in [0f2422b](https://github.com/movementlabsxyz/movement/pull/863/commits/0f2422b6fc7f1b7d6fee61477e5baa8f729d4769)

CODEOWNERS
mzabaluev: Ah, I see clearer now.

Do we consider Celestia as _the_ DA selected for Movement (the network as in `networks/movement`), or could there be other DAs that we migrate to in the future without hard a hard fork?

Is it conceivable that we could reuse the Celestia DA protocol units for a hypothetical future network that is not named Movement? Then I'd say `da/celestia` in the file tree and`movement-celestia-da*` in crate names (these are global, so we can't just publish `celestia-da*` crates without people getting confused) make more sense.

l-monninger: Yes, there could be other DAs. Yes, Celestia DA logic itself might be reused with other components besides this initial `movement` DA.

mzabaluev: `da/grpc` might be a poor choice for the subdir I mean to contain abstract protocol glue. Maybe `da/protocol`?

l-monninger: This DA is specifically the "Movement DA" as the before was the "M1 DA." SO I would again prefer, `da/movement/protcol` and similar.

mzabaluev: Got you, will rearrange accordingly. So I think there should be `da/movement/celestia/*` crates and `da/movement/protocol/light-node`, but then `movement` is the only thing under `da` so it looks like we are future-proofing.

mzabaluev: Done in [23d33dc](https://github.com/movementlabsxyz/movement/pull/863/commits/23d33dca6570624b8fe11ddc99c1e49bfa44cfeb)

l-monninger (approved):

PR #886, opened 2024-11-20
update state logic with the new bridge protocol

musitdev:
Update of the relayer state change adapted to the new bridge protocol.


protocol-units/bridge/state_logic.md
```diff
@@ -1,67 +1,34 @@
...
-
-  GE{"Event L2 completeBridgeTransfer Failed"}-->FF
-  HE{"Event refundBridgeTransfer Failed"}-->FF
-  IE{"Event BridgeTransferCancelledEvent Failed"}-->FF
+  DE{"Event BridgeTransferCompleted Tx fail"}-->DD("Aborted")
+  DD -->DA("Action wait -> completeBridgeTransfer")
 
+  EE{"Event BridgeTransferCompleted Event fail to read"}-->EEE("Error No change")
+  EEE -->EA("Action Log error")
+  FE{"Event Event BridgeTransferInitiated Event fail to read"}-->EEE("Error No change")
```
andygolay, line 32:
```suggestion
  FE{"Event BridgeTransferInitiated Event fail to read"}-->EEE("Error No change")
```

protocol-units/bridge/state_logic.md
```diff
@@ -1,67 +1,34 @@
...
- 
+
   BE{"Event BridgeTransferInitiated"}-->BB("Initialized")
-  BB -->BA{"Action lock_bridge_transfer"}
+  BB -->BA{"Action completeBridgeTransfer"}
 
-  CE{"Event BridgeTransferLockedEvent"}-->CC("Locked")
-  CC -->CA("Action None")
+  CE{"Event BridgeTransferCompletedEvent"}-->CC("Completed")
+  CC -->CA("Action Remove state")
```
andygolay, line 26: I'm not sure that it's in the scope of this file to answer, but I'm wondering how the state is removed. Is it assigned a value of "None"?

andygolay (approved):
LGTM, just one nit and one question.

protocol-units/bridge/state_logic.md
musitdev: It means that we don't manage the state in the relayer anymore. As it's an action, I found that naming it `Remove state` is better than done.

0xmovses (approved):

0xmovses:
A check is failing, I think a rebase is needed.

protocol-units/bridge/state_logic.md
```diff
@@ -1,67 +1,34 @@
...
     L1{"L1"} --Event BridgeTransferInitiated--> B("Initialized")
-    B --Action lock_bridge_transfer--> L2("L2")
-    L2 --Event BridgeTransferLockedEvent--> C("Locked")
-    L2 --Event BridgeTransferCompletedEvent--> D("KnowSecret")
-    D -- Action BridgeTransferCompletedEvent --> L1
-    L1 -- Event BridgeTransferCompleted --> E("Done(Completed)")
-    L2 -- Event lock_bridge_transfer Failed --> F("NeedRefund")
+    B --Action complete_bridge_transfer--> L2("L2")
+    L2 --Event BridgeTransferCompletedEvent--> C("Completed")
+    L2 -- Event BridgeTransferCompleted Tx Failed --> D("Aborted")
```
Primata, line 10: Why D("Aborted")?

protocol-units/bridge/state_logic.md
Primata: Can we stick to done? I think it's better than remove

Primata (changes requested):

protocol-units/bridge/state_logic.md
musitdev: It just says that the next action is to remove the state from the internal storage. When a transfer is completed, an action is generated to remove the state from the relayer state storage.
I can name it Done or None, but it's not what it means. 
As I say it's not a state of the transfer but an action to be executed when the transfer change it's state to completed.

protocol-units/bridge/state_logic.md
musitdev: Aborted is a state. The line mean that a transfer has this state when a complete Tx fail.
An aborted transfer is replayed.

PR #888, opened 2024-11-20
Adds Id to Hash and Verification of IR Blobs for Light Node

### LINKED ISSUE #877: (OTS) Blob ID Is Not Included in Signed Message
# Summary
[Blob ID Is Not Included in Signed Message](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/util/src/ir_blob.rs#L44) And Could Allow Forging a New Valid Blob with Invalid ID: None of the blob verification steps check that the id matches H(blob + timestamp), which means the following is possible:
An attacker can take a valid blob from Celestia, modify the id, and post it on Celestia again. Such a blob will pass all the verification steps in m1_da and end up in the full-node execution task where blob.id is [used for de-duplicating the execution](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/networks/suzuka/suzuka-full-node/src/tasks/execute_settle.rs#L124). Since the ID is now forged and new, it will be incorrectly re-executed.

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Adds id to hash and verification of IR blobs for light node. Resolves #877.

# Testing

1. Unit test added to ensure signing and verification respect id.

# Outstanding issues
None


protocol-units/da/m1/util/src/ir_blob.rs
```diff
@@ -24,6 +24,20 @@ impl InnerSignedBlobV1Data {
 		Self { blob, timestamp }
 	}
 
+	pub fn compute_id<C>(&self) -> Vec<u8>
+	where
+		C: PrimeCurve + CurveArithmetic + DigestPrimitive + PointCompression,
+		Scalar<C>: Invert<Output = CtOption<Scalar<C>>> + SignPrimitive<C>,
+		SignatureSize<C>: ArrayLength<u8>,
+		AffinePoint<C>: FromEncodedPoint<C> + ToEncodedPoint<C> + VerifyPrimitive<C>,
+		FieldBytesSize<C>: ModulusSize,
```
mzabaluev, line 33: Pity we weren't able to anything about this rust-crypto spam...

protocol-units/da/m1/util/src/ir_blob.rs
```diff
@@ -24,6 +24,20 @@ impl InnerSignedBlobV1Data {
 		Self { blob, timestamp }
 	}
 
+	pub fn compute_id<C>(&self) -> Vec<u8>
```
mzabaluev, line 27: There should be a doc comment explaining the purpose of the method.

mzabaluev: If we assume that the IR blob format bakes in the hash algorithm for these IDs, the ID could be a newtype wrapping a fixed-size array and exposing the array length as an associated item.

I haven't tried to reconcile const generics with `ArrayLength`, so it might be more difficult than it sounds. But please for the love of everything, do not use `generic-array` in APIs more than it already is.

mzabaluev (approved):

protocol-units/da/m1/util/src/ir_blob.rs
l-monninger: Will add and fix.

protocol-units/da/movement/celestia/util/src/ir_blob.rs
```diff
@@ -19,12 +19,33 @@ pub struct InnerSignedBlobV1Data {
 	pub timestamp: u64,
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Id(Vec<u8>);
```
mzabaluev, line 23: The dedicated newtype is better, but I hoped it'd be a non-allocating fixed size array as well, so that it implements `Copy`.

I guess this will have to wait until someone figures out a reasonable system of bounds to abstract the hash algorithm. The hash is always of fixed size, if I understand the currently used bounds correctly.

protocol-units/da/movement/celestia/util/src/ir_blob.rs
```diff
@@ -19,12 +19,33 @@ pub struct InnerSignedBlobV1Data {
...
+
+/// The id for an Ir Blob
+impl Id {
+	pub fn as_slice(&self) -> &[u8] {
+		self.0.as_slice()
+	}
+
+	pub fn to_vec(self) -> Vec<u8> {
```
mzabaluev, line 31: Should be `into_vec` by convention.

mzabaluev (changes requested):

protocol-units/da/movement/celestia/util/src/ir_blob.rs
l-monninger: Yes, but yes extracting the size is the trick. I had done this once somewhere else. Not going to worry about it here. Feel free to put in an issue.

mzabaluev (approved):

PR #889, opened 2024-11-20
Daily Rate Limiter

Primata:
# Summary

Using  https://github.com/bokkypoobah/BokkyPooBahsDateTimeLibrary?tab=readme-ov-file#timestamptodate to rate limit on a daily basis.

# Changelog

rate limit bidirectionally on Ethereum only.

# Testing

to be tested if accepted.

# Outstanding issues
verify


protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -120,7 +131,23 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
         emit BridgeTransferCompleted(bridgeTransferId, initiator, recipient, amount, nonce);
     }
 
+    function setRateLimits(uint256 _outboundRateLimit, uint256 _incomingRateLimit) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
0xmovses, line 134: Wouldn't we want to set both volume and duration? Here we just set volume.

0xmovses: Ok, I think this is fine to be restricted to a daily limit.

Primata: so, we could check if the time constraint has passed since the last transaction has been done, I can test both scenarios if you'd like. It would be nice to test which one is less expensive.

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -101,7 +116,7 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
         uint256 amount,
         uint256 nonce
     ) internal {
-        // _l2l1RateLimit(amount);
+        _incomingRateLimit(amount);
```
l-monninger, line 119: Instead of "outbound" and "incoming," use "outbound" and "inbound."

protocol-units/bridge/contracts/src/NativeBridge.sol
l-monninger: I would not call this correct w.r.t. to [MIP-56](https://github.com/movementlabsxyz/MIP/blob/mip/rate-limiter/MIP/MIP-56/README.md).

The amount insured is actually fixed by whatever account. The parameters which can be set are the `risk_period` and the `weight`.

l-monninger: This is also still does not use an actual balance to compute the rate limit as was suggested elsewhere.

l-monninger (changes requested):

l-monninger:
@0xmovses @Primata this whole PR should target main s.t.--if your MIP-58 work does not come through--this is still available.

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -19,7 +21,15 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
 
     mapping(uint256 nonce => OutgoingTransfer) public noncesToOutgoingTransfers;
     mapping(bytes32 bridgeTransferId => uint256 nonce) public idsToIncomingNonces;
+    // mapping(uint256 year => mapping(uint256 month => (uint256 day => uint256 amount))) public outboundRateLimitBudget;
+    // mapping(uint256 year => mapping(uint256 month => (uint256 day => uint256 amount))) public incomingRateLimitBudget;
 
+    uint256 public lastReset;
+    uint256 public outboundRateLimitBudget;
```
l-monninger, line 28: Budget is not the right term here; these are totals because we are counting. 

These would also need to be bucketed for a sliding window.

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -120,7 +135,41 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
...
+        _resetTime();
+        require(outboundRateLimitBudget += amount <= outboundRateLimit, OutboundRateLimitExceeded());
+    }
+
+    _rateLimitIncoming(uint256 amount) internal {
+        _resetTime();
+        require(incomingRateLimitBudget += amount <= outboundRateLimit, IncomingRateLimitExceeded());
+    }
+
+    _resetTime() internal {
```
l-monninger, line 168: This is too naive a way to do this. You would want a sliding window. 

Consider the case where you have a large volume of transfers immediately before one day since the last reset at $t-1$. You reset at $t$. Then, you have not deducted these from your budget at $t+1$. 

As the granularity of your tracking approaches the duration risk window, the value of the rate limit loses meaning. 

However, actuality, you don't need to reset. The rate limit is de facto provided by the balance of the insurance fund. The notion of the risk period or a daily rate limit is only for our operational perspective which is where the $\text{rateLimit} = \frac{\text{securityFund}}{2 * \text{riskPeriod}}$ comes from. As long as you are computing your rate limit affecting an individual transaction at any point in time according to this, you are safe under the operational framework of MIP-56. 

In other words, the budget will always be whatever we provide. 

What elements of the time boundary you would include here would be for fairness amongst users. So, if you have

l-monninger (changes requested):

protocol-units/bridge/contracts/src/NativeBridge.sol
apenzk: yes. this has been bugging me before as well and i thought that it introduces potential for unsafe situations. a sliding window would improve the situation

l-monninger: Yeah, I agree that if you keep that calculation in, it's safer.

What I would worry about though is now you're using terms that I thought were supposed to account for wanting the L2 to be covered and for availability of insurance funds for future risks periods just to account for safety in one window.

Sliding window is, ofc, just about how you are actually tracking inflight w.r.t to the above. The safest sliding window is naturally one bucketing on each block. This could actually not be that gas expensive to compute.

When you do not subject a transfer to the rate limit, you would do something like...
```
windows[block.timestamp] += transferAmount
```

In an ideal world, when you call a reset, you would just do something like...
```
gasToDeduct = windows[block.timestamp - 1 days]
totalUsed -= gasToDeduct
```

However, you're not likely to have timestamps line up like that. So, you will want to look back over a few timestamps. This is where using a larger window, i.e. slot, and mapping multiple blocks into a slot could help.

l-monninger: You also have to account for potentially not having a transfer in each window, and thus a need to gc--going back further.

l-monninger: For this reason, it may be better for MVP to use relatively large slots to divide the risk period into $n$ fractions and simply sum over $n$ totals.

l-monninger: Basically, I would encourage you to come to something safer via a combination of a heuristics treatment of the terms in MIP-56 and more granular tracking of totals.

l-monninger: For example, if you wanted to leave this entirely heuristic, computing  $\text{rateLimit} = \frac{\text{securityFund}}{4 * \text{riskPeriod}}$ would be basically the same as having two windows per risk period.

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -135,45 +128,25 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
...
     function togglePause() external onlyRole(DEFAULT_ADMIN_ROLE) {
         paused() ? _pause() : _unpause();
+        emit PauseToggled(paused());
     }
 
     function _rateLimitOutbound(uint256 amount) internal {
         (uint256 year, uint256 month, uint256 day) = block.timestamp.timestampToDate();
         outboundRateLimitBudget[year][month][day] += amount;
-        require(outboundRateLimitBudget[year][month][day] <= outboundRateLimit, OutboundRateLimitExceeded());
+        require(outboundRateLimitBudget[year][month][day] < moveToken.balanceOf(insuranceFund) / 4, OutboundRateLimitExceeded());
```
l-monninger, line 144: Yeah, this is fine.

Once I run the tests I'll approve.

PR #891, opened 2024-11-21
Bridge: Update relayer utils crate

musitdev:
# Summary

First PR for the new Bridge protocol transition. Remove the old type and define a new one.
It's WIP. The PR breaks the bridge code.
# Changelog

Update the utils crate for the new bridge protocol:
Remove all preimage, hashlock, and timelock struct.
Cut the BridgeContract trait into 2 traits: BridgeClientContract and BridgeRelayerContract to reflect the 2 roles between USer and Relayer.
Update the state and transition WIP

# Testing

# Outstanding issues

By merging this PR it will break all the relayer code.


0xmovses (changes requested):
I like the addition of the trait, it's good to define the boundaries and behaviour in this way. Nut I think the PR should at least build before merging into feature branch. It can break tests for now.

musitdev:
If we want to work in // on the code we need to merge fast even if it doesn't compile.
The other way is I do everything then we merge.

0xmovses:
Let's comment out the code that doesn't compile then, that will be useful to see at a glance also.

musitdev:
If we do that we're going to comment a lot of code and in the end, it will be difficult to go back. If we let the compiler show us where there are issues it's easier to update instead of un commenting code to see if it builds.
I think it's the role of the feature branch to not build at the beginning.

0xmovses:
Alright, so is this now ready to go into feature. I checked the code looks fine.

PR #896, opened 2024-11-22
Sponsorship Fee

l-monninger:
# Summary
- **RFCs**:  $\emptyset$

Adds a sponsorship fee on the L1 as an optional but recommended safety feature for https://github.com/movementlabsxyz/aptos-core/pull/104

# Testing

1. Some tests updated for assertions. Generally, however, there is a lot that's unchecked in the current tests. 

# Outstanding issues


andygolay (approved):
LGTM.

If users have to pay for the sponsorship upfront, they'll appreciate it more.

And they'll only need to do it once, due to the check in `is_sponsorship_eligible` in https://github.com/movementlabsxyz/aptos-core/pull/104/

PR #912, opened 2024-12-03
Fix Race Condition in Block Building

### LINKED ISSUE #882: (OTS) Race Condition in Parent Block Update
# Summary
[The parent_block update code here](https://github.com/movementlabsxyz/movement/blob/f8e6e303c5257af5b1bca4913031311bb2fdb2a7/protocol-units/sequencing/memseq/sequencer/src/lib.rs#L108-L114) has a race condition if two threads both read the same parent_block together and only then both update the parent_block. The issue is caused by the read lock being released before the write, creating a small window where other thread might jump-in. This is currently impossible though since the sequencer only runs on a single thread, but it needs to be fixed if the single thread assumption might ever change.

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Resolves #882 

# Testing
1. Adds units test to check that the parent block is properly updated. 

# Outstanding issues
1. This is fairly trivial. But, it would be nice to have this race condition be checked by tests or verification someday.

### DIFF
--- a/protocol-units/sequencing/memseq/sequencer/src/lib.rs
+++ b/protocol-units/sequencing/memseq/sequencer/src/lib.rs
@@ -16,9 +16,11 @@ use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 
 #[derive(Clone)]
 pub struct Memseq<T: MempoolTransactionOperations> {
+	/// The mempool to get transactions from.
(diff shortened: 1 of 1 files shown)


mzabaluev (approved):
Looks good.

Consider using `loom` or something similar to test possible race conditions. But it's not required for this PR.

PR #913, opened 2024-12-03
Fixes Block Splitting Implementation Used in Heuristic for Celestia Submission

### LINKED ISSUE #878: (OTS) Sequencer Heuristic Does Not Split Blocks
# Summary
The sequencer heuristics try to generate optimal blocks by performing different heuristic passes on the passed-in mempool blocks. However, the final blob bytes that are submitted to Celestia are only generated at the start and are never changed based on the heuristics. For example, if B1[tx1, tx2] is the original block and ends up being split, the two new WrappedBlocks will still store the same Celestia blob bytes containing both transactions. This seems incorrect. Here’s the relevant code:
- WrappedBlock created [here](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/light-node/src/v1/sequencer.rs#L170) with signed blob of the original block
- WrappedBlock’s Split [implementation](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/light-node/src/v1/sequencer.rs#L519) here simply clones the underlying blob in the two new WrappedBlocks
- Finally, the cloned blob is submitted to Celestia [here](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/light-node/src/v1/sequencer.rs#L146) (called from [here](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/light-node/src/v1/sequencer.rs#L200))

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Resolves #878 

# Testing

1. Adds unit test to check the splitting implementation results in the correct split blocks. 

# Outstanding issues
None

### DIFF
--- a/protocol-units/da/movement/celestia/light-node/src/v1/sequencer.rs
+++ b/protocol-units/da/movement/celestia/light-node/src/v1/sequencer.rs
@@ -484,7 +484,7 @@ pub mod block {
 	use movement_types::block::Block;
 
 	/// A wrapped block that can be used with the binpacking heuristic
-	#[derive(Debug)]
+	#[derive(Debug, Clone, PartialEq, Eq)]
 	pub struct WrappedBlock {
 		pub block: Block,
 		pub blob: Blob,
@@ -513,10 +513,11 @@ pub mod block {
 
 	impl Splitable for WrappedBlock {
 		fn split(self, factor: usize) -> Result<Vec<Self>, anyhow::Error> {
+			let namespace = self.blob.namespace;
 			let split_blocks = self.block.split(factor)?;
 			let mut wrapped_blocks = Vec::new();
 			for block in split_blocks {
-				let wrapped_block = WrappedBlock { block, blob: self.blob.clone() };
+				let wrapped_block = WrappedBlock::try_new(block, namespace)?;
 				wrapped_blocks.push(wrapped_block);
 			}
 			Ok(wrapped_blocks)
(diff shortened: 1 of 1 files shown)


protocol-units/da/movement/celestia/light-node/src/v1/sequencer.rs
```diff
@@ -513,10 +513,11 @@ pub mod block {
 
 	impl Splitable for WrappedBlock {
 		fn split(self, factor: usize) -> Result<Vec<Self>, anyhow::Error> {
+			let namespace = self.blob.namespace;
 			let split_blocks = self.block.split(factor)?;
 			let mut wrapped_blocks = Vec::new();
 			for block in split_blocks {
-				let wrapped_block = WrappedBlock { block, blob: self.blob.clone() };
+				let wrapped_block = WrappedBlock::try_new(block, namespace)?;
```
mzabaluev, line 520: I wonder if anything in block creation is actually fallible.
The block to wrap is supposed to be correct by construction, and it would be unexpected for the BCS encoder to fail. Or are there size restrictions on encoded bytes, for example?

mzabaluev (approved):
LGTM with some passing remarks.

protocol-units/da/movement/celestia/light-node/src/v1/sequencer.rs
l-monninger: I don't think `bcs::to_bytes` is actually fallible. Not sure about `zstd`.

PR #917, opened 2024-12-04
Rate Limiter Parameterization

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Adds parameterization to rate limit contracts and tests. 
- Parameterization allows setting rate limits less than $\frac{1}{4}$
- Parameterization allows shutting down the bridge without having to transfer away from the insurance fund. 

# Testing
1. Unite tests added demonstrating usage. 

# Outstanding issues
None

### DIFF
--- a/protocol-units/bridge/contracts/src/RateLimiter.sol
+++ b/protocol-units/bridge/contracts/src/RateLimiter.sol
@@ -49,7 +49,7 @@ contract RateLimiter is AccessControlUpgradeable {
 
     function setRateLimiterCoefficients(uint256 numerator, uint256 denominator) external onlyRole(RATE_LIMITER_ADMIN) {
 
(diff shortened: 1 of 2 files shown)


protocol-units/bridge/contracts/src/RateLimiter.sol
```diff
@@ -6,14 +6,23 @@ import {IERC20} from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 contract RateLimiter is AccessControlUpgradeable {
     bytes32 public constant ATOMIC_BRIDGE = keccak256("ATOMIC_BRIDGE");
+    bytes32 public constant RATE_LIMITER_ADMIN = keccak256("RATE_LIMITER_ADMIN");
 
     mapping(uint256 day => uint256 amount) public outboundRateLimitBudget;
```
Primata, line 11: It should be version to day to amount mapping with versioning then

Primata (changes requested):

PR #923, opened 2024-12-04
Bridge: Relayer logic unit test

musitdev:
# Summary

Add Mock client and monitor to test relayer logic.
Put the test for happy and unhappy path.
Comment the unhappy path because the relayer logic doesn't implement the replay complete Tx in case of failure.
Need to update this part before activate the test.
# Changelog

# Testing

The test run alone with the command:
`RUST_BACKTRACE=1 cargo test --test relayer test_relayer_logic -- --nocapture --test-threads=1`

# Outstanding issues

The relayer doesn't replay failed complete Tx. The unhappy path test failed that why it has been commented until the loic is corrected on the relayer.

### DIFF
--- a/protocol-units/bridge/service/tests/relayer.rs
+++ b/protocol-units/bridge/service/tests/relayer.rs
@@ -1,0 +1,267 @@
+use bridge_service::types::Amount;
+use bridge_util::chains::bridge_contracts::BridgeContractError;
+use bridge_util::chains::bridge_contracts::BridgeContractResult;
+use bridge_util::chains::bridge_contracts::BridgeTransferCompletedDetails;
+use bridge_util::chains::bridge_contracts::BridgeTransferInitiatedDetails;
+use bridge_util::types::AddressError;
+use bridge_util::types::BridgeAddress;
+use bridge_util::types::Nonce;
+use bridge_util::BridgeContractEvent;
+use bridge_util::BridgeContractMonitoring;
+use bridge_util::BridgeRelayerContract;
+use bridge_util::BridgeTransferId;
+use futures::SinkExt;
+use futures::{
+	channel::mpsc::{UnboundedReceiver, UnboundedSender},
+	Stream, StreamExt,
+};
+use std::{pin::Pin, task::Poll};
+use tiny_keccak::{Hasher, Keccak};
+
+#[derive(Debug, PartialEq, Eq, Hash, Clone)]
+pub struct MockAddress(pub Vec<u8>);
+
+impl From<MockAddress> for Vec<u8> {
+	fn from(address: MockAddress) -> Self {
+		address.0
+	}
+}
+
+impl From<BridgeAddress<MockAddress>> for MockAddress {
+	fn from(address: BridgeAddress<MockAddress>) -> Self {
+		address.0
+	}
+}
+
+impl std::ops::Deref for MockAddress {
+	type Target = Vec<u8>;
+
+	fn deref(&self) -> &Self::Target {
+		&self.0
+	}
+}
+
+impl TryFrom<Vec<u8>> for MockAddress {
+	type Error = AddressError;
+
+	fn try_from(vec: Vec<u8>) -> Result<Self, Self::Error> {
+		Ok(MockAddress(vec))
+	}
+}
+
+fn calculated_transfer_bridfe_id(
+	initiator: &[u8],
+	recipient: &[u8],
+	amount: Amount,
+	nonce: Nonce,
+) -> BridgeTransferId {
+	let mut hasher = Keccak::v256();
+	hasher.update(initiator);
+	hasher.update(recipient);
+	hasher.update(&amount.to_le_bytes());
+	hasher.update(&nonce.0.to_le_bytes());
+	let mut output = [0u8; 32];
+	hasher.finalize(&mut output);
+	BridgeTransferId(output)
+}
+
+#[derive(Clone)]
+pub struct RelayerMockClient {
+	sender: UnboundedSender<BridgeContractResult<BridgeContractEvent<MockAddress>>>,
+	complete_notifier:
+		tokio::sync::mpsc::Sender<BridgeContractResult<BridgeContractEvent<MockAddress>>>,
+}
+
+impl RelayerMockClient {
+	pub fn build(
+		sender: UnboundedSender<BridgeContractResult<BridgeContractEvent<MockAddress>>>,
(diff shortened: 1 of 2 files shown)


protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+use bridge_util::types::Nonce;
+use bridge_util::BridgeContractEvent;
+use bridge_util::BridgeContractMonitoring;
+use bridge_util::BridgeRelayerContract;
+use bridge_util::BridgeTransferId;
+use futures::SinkExt;
+use futures::{
+	channel::mpsc::{UnboundedReceiver, UnboundedSender},
+	Stream, StreamExt,
+};
```
0xmovses, line 17: Can we group the imports to avoid re importing the same crate?

protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+impl TryFrom<Vec<u8>> for MockAddress {
+	type Error = AddressError;
+
+	fn try_from(vec: Vec<u8>) -> Result<Self, Self::Error> {
+		Ok(MockAddress(vec))
+	}
+}
+
+fn calculated_transfer_bridfe_id(
```
0xmovses, line 52:
```suggestion
fn calculated_transfer_bridge_id(
```

protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+
+	// //get ok transfer completed
+	// let event = tokio::time::timeout(std::time::Duration::from_secs(5), l2_notifier.recv())
+	// 	.await
+	// 	.expect("L2 commplete not call by the relayer.");
+	// assert!(event.is_some());
+	// let event = event.unwrap();
+	// assert!(event.is_ok());
+	// let event = event.unwrap();
+	// assert_eq!(event.bridge_transfer_id(), l1_transfer_id);
```
0xmovses, line 273: Let's remove the commented out lines.

protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+	// assert!(event.is_err());
+	// //get ok transfer completed
+	// let event = tokio::time::timeout(std::time::Duration::from_secs(5), l1_notifier.recv())
+	// 	.await
+	// 	.expect("L1 commplete not call by the relayer.");
+	// assert!(event.is_some());
+	// let event = event.unwrap();
+	// assert!(event.is_ok());
+	// let event = event.unwrap();
+	// assert_eq!(event.bridge_transfer_id(), l2_transfer_id);
```
0xmovses, line 325: Same again remove the commented out lines

protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+	let (mut l2_sender, l2_listener) = futures::channel::mpsc::unbounded::<
+		BridgeContractResult<BridgeContractEvent<MockAddress>>,
+	>();
+	let (l2_relayer_client, mut l2_notifier) = RelayerMockClient::build(0, l2_sender.clone());
+	let l2_monitor = MockMonitoring::build(l2_listener);
+
+	let (l1_health_tx, _l1_health_rx) = tokio::sync::mpsc::channel(10);
+	let (l2_health_tx, _l2_health_rx) = tokio::sync::mpsc::channel(10);
+	let (_health_tx, health_rx) = tokio::sync::mpsc::channel(10);
+	let _loop_jh = tokio::spawn(async move {
```
0xmovses, line 229: Can we name this var to something else, not sure what jh refers to. Or just use `_` as its unused anyway

protocol-units/bridge/service/tests/relayer.rs
```diff
@@ -0,0 +1,347 @@
...
+		amount,
+	};
+	let event = BridgeContractEvent::Initiated(details);
+
+	sender.send(Ok(event)).await.unwrap();
+	bridge_transfer_id
+}
+
+#[tokio::test]
+async fn test_relayer_logic() -> Result<(), anyhow::Error> {
```
0xmovses, line 200:
```suggestion
async fn test_l1_to_l2_relayer_logic() -> Result<(), anyhow::Error> {
```

0xmovses (changes requested):
In general I think it's a good idea to start writing more unit tests, as the integration tests are cumbersome and expensive. 

In this test all you are asserting is that the event is received. Perhaps the intention of this PR is just to setup the mocks and initial tooling for the unit testing, are you intending to write more cases and assertions in future PRs?

protocol-units/bridge/service/tests/relayer.rs
musitdev: I put jh for joinhandle. At first I thought I need to use it, but in fact no. I'll rename it with _.

protocol-units/bridge/service/tests/relayer.rs
musitdev: I test both direct in the test.

protocol-units/bridge/service/tests/relayer.rs
musitdev: Personalty I prefer to have separated import, it's easier when you've to move or copy code for example. Reimporting is impossible because of the compiler.
The last my tooling doesn't do it automatically.
There's a rule that we must group import?

protocol-units/bridge/service/tests/relayer.rs
musitdev: As I said in the description, I write the test for the unhappy path, but they failed because the logic in not implemented in the relayer code. I keep the test commented to avoid redeveloping it when the relayer will be corrected.
The other possibility is to keep them and accept to have some test that fail. I choose to have the test passing so I comment them.
I work on the relayer logic update.

musitdev:
In the test, I assert that the test send the expected event. The listener is not the relayer listener but the mock listener. I use it to validate that the relayer call the complete function of the client.
It tests the relayer logic. So I send an initiate event and I want to be sure the relayer call the complete function of the client. 
In the unhappy path, I generate an error in the complete transfer call and I want to be sure the call is retry.
So the test goal is to test all the relayer logic that why I mock the client part and add some notification (listener event) in the mock to detect the right client function are called.

protocol-units/bridge/service/tests/relayer.rs
musitdev: done

0xmovses:
This can be closed now as it's part of #942 .

PR #924, opened 2024-12-04
Fix mvmt client test 

0xmovses:
# Summary
Updates the `Nonce` value to 1. Note Auto-formatting was applied, not sure why it was formatting to other settings, double check your IDE to make sure the fmt settings we have in the workspace are being applied. 

We can't generate it in the same way as the Eth client test as the nonce it is asserting against is fetched from the emitted event. 

A developer has the node running and repeatedly runs the same integration test over and over.  

I wasn't actually able to run the test locally due the integration test randomly hanging forever. But this was the what the assertion was looking for so it should pass. Let's get CI up to remove this burden of having to run all this locally cc @andygolay 

# Changelog
- Update the Nonce

# Testing
1. 
```
CELESTIA_LOG_LEVEL=FATAL CARGO_PROFILE=release CARGO_PROFILE_FLAGS=--release nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash -c "just bridge native build.setup.eth-local.celestia-local.bridge.bridge-indexer --keep-tui"
```
2. `RUST_BACKTRACE=1 DOT_MOVEMENT_PATH="$(pwd)/.movement" cargo test --test client_mvt_tests test_movement_client_initiate_transfer -- --nocapture --test-threads=1`

### DIFF
--- a/protocol-units/bridge/integration-tests/tests/client_mvt_tests.rs
+++ b/protocol-units/bridge/integration-tests/tests/client_mvt_tests.rs
@@ -14,8 +14,8 @@ use bridge_util::BridgeClientContract;
 use bridge_util::BridgeContractEvent;
 use bridge_util::BridgeRelayerContract;
 use futures::StreamExt;
-use tokio::{self};
 use rand::Rng;
+use tokio::{self};
 
 #[tokio::test]
 async fn test_movement_client_initiate_transfer() -> Result<(), anyhow::Error> {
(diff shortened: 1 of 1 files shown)


andygolay (changes requested):
Simply setting the nonce to 1 isn't enough because the nonce updates in the Move module each time the test is run. So it will be 2 on the 2nd time, etc. Seems to me we need to be able to first get the current nonce from the module and then after calling the function, check that the nonce has been incremented. Or we could remove the assertion about the nonce, but probably best to check all the details from the Initiated event.

0xmovses:
But an integration such as this should always be tearing down and starting state (or the node) on each run. We should be adjusting values to compensate for state changing on consecutive runs. This can cause different kinds of issues. 

The node start and stop commands should be happening within the tests themselves (different PR). Tests should not be run over and over and again without restarting the node. 

So Nonce should always be 1

0xmovses:
Also in the ETH initiate tests we never even assert on the Nonce so its not really a must-have. Let's keep it at 1 for now.

andygolay (approved):

PR #925, opened 2024-12-04
Bridge: Add the address convertion trait to the relayer type.

musitdev:
# Summary

Initially it was done to remove the hex encoding of the initiator address from the relayer logic since the PR #916.
@andygolay is working on the serialization to remove the encoding, but I think it's a good thing to have this convert trait in the relayer logic in case we need to add specific processing (like the hex encoding) during the relayer address conversion without putting it in its logic.
I put in draft until the hex encoding is fixed.
# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/service/src/actions.rs
+++ b/protocol-units/bridge/service/src/actions.rs
@@ -31,7 +31,7 @@ where
 					.complete_bridge_transfer(
 						bridge_transfer_id,
 						initiator,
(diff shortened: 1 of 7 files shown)


0xmovses:
I assume this is not ready for review yet?

musitdev:
In fact it was but Andy has done some changed so I need to adapt a little. I switch it to draf.

0xmovses:
We can close this now as it's no longer needed.

PR #932, opened 2024-12-06
Bridge: Refactor bridge logic and add the retry complete transfer Tx.

musitdev:
# Summary

This PR contains an important change in the relayer logic. To avoid all the duplicated code, the relayer loop has been changed to process only one direction from source to target blockchain.
Update the relayer logic to manage failed Complete transfer Tx and retry 5 times.

All the relayer logic is implemented.

# Changelog
 * one direction relayer.
 * Failed Tx retry.
 * End transfer on Complete Tx event.
 * Add a method called child to get a monitoring stream copy that get chain events without adding a new pull task. The initial monitor pulling task is shared between all child monitors.
 * Add unhappy path test.
 * Remove relayer code from lib.rs and define specific files.
 * Do some cleaning
 

# Testing

To run the happy and unhappy path unit test.

` RUST_BACKTRACE=1 cargo test --test relayer -- --nocapture --test-threads=1`

# Outstanding issues

The indexer Db logic hasn't been integrated. This part still need to be done.


0xmovses:
This is part of #942 now so can close/

PR #937, opened 2024-12-10
Zstd bomb fix

### LINKED ISSUE #876: (OTS) Zstd Bomb Blob
# Summary
A malicious user may post a zstd bomb (a small compressed blob which decompresses to a huge buffer) as a blob on the movement Celestia namespace. Even though blob data is checked to be signed, zstd decompression happens before the signature checks. This is done with the [zstd::decode_all(blob.data)](https://github.com/movementlabsxyz/movement/blob/0fe8c12c7fd2bf2d0b39ce5356e4ae539886099e/protocol-units/da/m1/util/src/ir_blob.rs#L171) function, which has no limits on the decompressed size of the blob. Even with Celestia’s 2 MB blob size limit, we were able to produce a PoC that uses approximately 100 GB of RAM on decompression.

mzabaluev:
Fixes #876 (but see outstanding issues below), #1008.

# Dependencies

https://github.com/movementlabsxyz/aptos-core/pull/111

# Summary

Use streamed decompression API to avoid potentially huge allocations
for blob data decompressed from zstd. The bcs decoder enforces the
length limit of 2^31 - 1 on byte arrays.

# Changelog

- Prevent unlimited memory allocations in decoding compressed blobs from the DA.

# Testing

Added unit tests to movement-celestia-da-util to exercise compression and decompression,
with supercritical and valid lengths for compressed payload and the data field. The test creating legit data structures with super-long blobs is ignored in default runs because of the memory and processing requirements.

# Outstanding issues

As commented in https://github.com/movementlabsxyz/movement/issues/876#issuecomment-2530058294, there are four byte array fields in the blob data structure that can each be bloated up to 2 GiB.


protocol-units/da/movement/celestia/util/src/ir_blob.rs
```diff
@@ -212,13 +212,9 @@ pub mod celestia {
...
-			// decompress blob.data with zstd
-			let decompressed =
-				zstd::decode_all(blob.data.as_slice()).context("failed to decompress blob")?;
-
-			// deserialize the decompressed data with bcs
-			let blob =
-				bcs::from_bytes(decompressed.as_slice()).context("failed to deserialize blob")?;
+			// decompress the blob and deserialize the data with bcs
+			let decoder = zstd::Decoder::new(blob.data.as_slice())?;
+			let blob = bcs::from_reader(decoder).context("failed to deserialize blob")?;
```
l-monninger, line 217: Do you know how [`fill_buf`](https://github.com/gyscos/zstd-rs/blob/bfe1e34f593c2427e9dc35f1b98d06788174711f/src/stream/zio/reader.rs#L123) would be called here? I wondering if this in fact going to raise an `Error` or whether it might just panic if the buffer size is restricted.

mzabaluev: I don't think I fully understand this question, and we didn't restrict the size of the internal buffer.
But you did bring to my attention that the slice also supports `BufRead` and we can take advantage of it.

mzabaluev: Yes, bcs would return an `Error`.

I have implemented a customizable limit on length of any variable-length fields in https://github.com/zefchain/bcs/pull/12.

l-monninger:
It looks like there was a simple container conflict.

I still have same reservation that we can effectively still be `zstd` bombed because the block processing time would exceed 10 seconds from the cases we've talked about offline. But, I will approve for now and we can revisit later.

l-monninger (approved):

PR #952, opened 2024-12-12
Only Inbound Rate Limiting 

Primata:
# Summary

Because we will be switching to Inbound Rate Limiting only, moving the Outbound Rate Limiting to the Move side of the bridge, we needed to remove the outbound rate limiter locally.
This PR also addresses Rate Limiting concerns raised on feature/trusted-relayer

# Changelog

- Removal of OutboundRateLimiter
- setRiskDenominator

# Testing

- Rate Limiting Tests including change of denominator 

# Outstanding issues


0xmovses (approved):
Thanks for this.

PR #964, opened 2024-12-18
Hsm/signapi

musitdev:
# Summary

First draft of the signing API as a lib.
# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/util/signing/src/lib.rs
+++ b/util/signing/src/lib.rs
@@ -19,11 +19,26 @@ pub struct KeyVersion(pub String);
 #[derive(Debug, Clone)]
 pub struct KeyId(pub String);
 
+/// Errors thrown by SigningService.
+#[derive(Debug, thiserror::Error)]
+pub enum SignerError {
+	#[error("Error during signing : {0}")]
+	Sign(String),
+	#[error("Error during public key retrieval : {0}")]
+	GetPublicKey(String),
+	#[error("Error can't decode provided hex data : {0}")]
+	Hex(String),
+	#[error("Signature not found.")]
+	SignatureNotFound,
+	#[error("public key not found.")]
+	PublicKeyNotFound,
+}
+
 pub struct SigningService;
 
 impl SigningService {
 	/// Create the service with environment variable.
-	pub fn try_from_env() -> Result<Self, anyhow::Error> {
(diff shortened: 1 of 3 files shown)


util/signing/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+#[derive(Debug, Clone, Default)]
+pub struct KeyVersion(pub String);
+
+/// Id that identify a Key.
+#[derive(Debug, Clone)]
+pub struct KeyId(pub String);
+
+pub struct SigningService;
+
+impl SigningService {
```
mzabaluev, line 24: Should this rather be a trait? I think we need implementations of this for the different backends.

util/signing/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+	/// Create the service with environment variable.
+	pub fn try_from_env() -> Result<Self, anyhow::Error> {
```
mzabaluev, line 26: If `SigningService` (or `Hsm` in the earlier demo PRs) is to be a trait, this does not belong in the abstraction. We potentially want the dynamic dispatch to be feasible, and the constructor APIs should belong with the specific backends.

l-monninger:
Right now, this looks more less like what was already used in #961 and proposed in the original demo application. 

I recommend you do a spec API similar to #961 to make it easier to generate modules. Check the following lines for reference:
- `AwsKms` spec: https://github.com/movementlabsxyz/movement/blob/l-monninger/hsm-demo/demo/hsm/src/cryptography/aws_kms.rs
- `AwsKms` spec usage: https://github.com/movementlabsxyz/movement/blob/df28163913abe8d88d9510fd2099eeeba50bdc0d/demo/hsm/src/hsm/aws_kms.rs#L81
- `AwsKms` HSM usage with spec: https://github.com/movementlabsxyz/movement/blob/2093ac1e3530fffbcffa8b3386067929155c5146/demo/hsm/src/cli/server/secp256k1/aws_kms.rs#L14

util/signing/src/lib.rs
musitdev: The idea was to be backend agnostic so that the application don't have to know how keys are organized.

PR #969, opened 2024-12-19
Simplified signing API, Aptos integration

mzabaluev:
# Summary
- **Categories**: `util`.

Building on #968, simplify the signing API and use fixed-size types for crypto material.

Add integration with aptos types, enabling use of async remote signers
to sign Aptos transactions.

# Changelog

* movement-signer: an asynchronous remote cryptographic signing API
* movement-signing-aptos: integration of the signer API for Aptos

# Testing

To be added

# Outstanding issues

* Consider not adding a dependency on ring-compat, as the verification implementations are available in crates already depended on elsewhere, or have small pure no_std Rust implementations.


util/signing/integrations/aptos/src/lib.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
+pub enum Error {
+	#[error("failed to produce signing material")]
+	CryptoMaterial(#[from] CryptoMaterialError),
+	#[error(transparent)]
+	Signer(#[from] SignerError),
+}
+
+pub trait TransactionSigner: Sync {
```
l-monninger, line 17: This looks about right.

util/signing/interface/src/cryptography/ed25519.rs
```diff
@@ -0,0 +1,36 @@
+use crate::cryptography::Curve;
+use crate::{Verify, VerifyError};
+use anyhow::Context;
+use ring_compat::signature::{ed25519, Verifier};
+
+/// The Ed25519 curve.
+#[derive(Debug, Clone)]
+pub struct Ed25519;
+
+fixed_size!(pub struct PublicKey([u8; 32]));
```
l-monninger, line 10: How does `fixed_sized` help here?

util/signing/interface/src/cryptography/ed25519.rs
```diff
@@ -0,0 +1,36 @@
...
+fixed_size!(pub struct Signature([u8; 64]));
+
+impl Curve for Ed25519 {
```
l-monninger, line 13: Yeah, this is how I would do the associated types.

util/signing/interface/src/cryptography/mod.rs
```diff
@@ -0,0 +1,26 @@
+macro_rules! fixed_size {
```
l-monninger, line 1: Oh, I see. This isn't from the `fixed_size` crate. It's custom.

util/signing/interface/src/cryptography/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+			}
+		}
+	};
+}
+
+pub mod ed25519;
+pub mod secp256k1;
+
+/// A designator for an elliptic curve.
+///
```
l-monninger, line 20: Yeah, good.

util/signing/interface/src/cryptography/ed25519.rs
mzabaluev: It's a macro I defined in `cryptography/mod.rs`, to have one place to write the boilerplate for these fixed-sized types. The types are specific to their elliptic curves to provide type safety against mixing different types of keys or signatures, so they have each their own definition, with no type aliasing or const generics to reuse.

musitdev (approved):

PR #970, opened 2024-12-19
Alloy signing integration

musitdev:
# Summary

First version of Alloy crate integration for signing with the API. Inspired from the Alloy AWS implementation `alloy_signer_aws::AwsSigner`
# Changelog

# Testing
Provide an integration test to test AWS KMS integration. The test can't be used in CI because it needs AWS KMS auth env variable to set to run (`AWS_ACCESS_KEY` and `AWS_SECRET_KEY`).
The test start a local Anvil eth node and execute a transfer Tx using an AWS KMS key defined with the `AWS_KEY_ID` env var.

`AWS_KEY_ID=<>AW key id> AWS_ACCESS_KEY=<access key> AWS_SECRET_KEY=<secret key> cargo test -p movement-signing-alloy test_aws_kms_send_tx -- --nocapture
`
# Outstanding issues


util/signing/integrations/alloy/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+        f.debug_struct("HsmSigner")
+            .field("chain_id", &self.chain_id)
+            .field("pubkey", &hex::encode(self.pubkey.to_sec1_bytes()))
+            .field("address", &self.address)
+            .finish()
+    }
+}
+
+#[async_trait::async_trait]
+impl alloy_network::TxSigner<AlloySignature> for HsmSigner {
```
l-monninger, line 27: Have you already worked on the code for how this gets integrated? I think it's still probably easier for us to change the usage s.t. we are producing a raw transaction and signing that rather than tightly coupling with a target API.

util/signing/integrations/alloy/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+    fn set_chain_id(&mut self, chain_id: Option<ChainId>) {
+        self.chain_id = chain_id;
+    }
+}
+
+impl HsmSigner {
+    /// Instantiate a new signer from an existing `Client` and key ID.
+    ///
+    /// Retrieves the public key from HMS and calculates the Ethereum address.
+    pub async fn new(
```
l-monninger, line 66: Because this returns a result, I would call it `try_new`

l-monninger (changes requested):

util/signing/integrations/alloy/src/lib.rs
musitdev: Changed

musitdev:
Integrated in another PR.

PR #973, opened 2024-12-20
Musitdev/signapi take2

musitdev:
# Summary

From the discussion we add, the implementation I've started this is a new proposition that integrate the signing part and the key management part.
# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/util/signing/signer/src/lib.rs
+++ b/util/signing/signer/src/lib.rs
@@ -1,31 +1,12 @@
 pub mod cryptography;
 
-/// A collection of bytes.
-#[derive(Debug, Clone)]
-pub struct Bytes(pub Vec<u8>);
-
-/// A signature.
-#[derive(Debug, Clone)]
-pub struct Signature(pub Bytes);
-
-/// A public key.
-#[derive(Debug, Clone)]
-pub struct PublicKey(pub Bytes);
-
-/// Version of a key.
-/// Default mean the current key.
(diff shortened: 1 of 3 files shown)


util/signing/signer/src/lib.rs
```diff
@@ -89,36 +66,60 @@ pub trait VerifierOperations<C: cryptography::Curve> {
...
 
-/// Verifier wraps an implementation of [VerifierOperations] and provides a simple API for verifying signatures.
-impl<O, C> Verifier<O, C>
-where
-	O: VerifierOperations<C>,
-	C: cryptography::Curve,
-{
-	pub fn new(operations: O) -> Self {
-		Self { operations, _curve_marker: std::marker::PhantomData }
+impl KeyManager {
```
l-monninger, line 89: `KeyManager` with a single `KeyId` type will likely not be particularly useful as each cloud already kind of has their own format.

musitdev: The idea was to hide the cloud provider/real key and define more an application key name. For example, the relayer key id is "relayer" and in the key manager config, the relayer key is associated to the AWS cloud and a AWS key id. 
One of the requirement was that the application shouldn't know which provider manages the key so that the effective key can be changed without rebuilding the application.
So the role of the Key manager is to define this mapping between application key_id and the real key.
Perhaps I should rename to keyname.

PR #974, opened 2024-12-20
Test signer

mzabaluev:
# Summary
- **Categories**: `util`.

Add a movement-signer-test crate, providing a local `TestSigner`, initially for the Ed25519 curve.

# Changelog

* movement-signer-test: local test support for the signing API defined in movement-signing.

# Testing

To be added


util/signing/interface/src/cryptography/mod.rs
```diff
@@ -29,15 +33,18 @@ macro_rules! fixed_size {
...
 pub trait Curve {
-	type PublicKey: TryFromBytes;
+	type PublicKey;
 	type Signature;
 }
+
+/// Errors that occur when parsing signature or key material from byte sequences.
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct CryptoMaterialError(Box<dyn Error + Send + Sync>);
```
musitdev, line 50: I think it will be easier to have one error type for the crate. It avoids adding map_err every time when you need to extract the publickey from an u8 array, for example, and it eases the error management for external crate.
There's a reason you didn't add a new variant to the SignerError. Perhaps we need to rename it in a more generic name?

mzabaluev: I don't like using the "fat" main error type for the much slimmer use case of `TryFrom` conversions.

PR #975, opened 2024-12-20
DEVOP-53 automatically deploy main branch to devNet

mcmillennick:
# Summary
- **Categories**: `cicd`

Adds workflow jobs for triggering new deployment to k8s pipeline.

# Changelog

Modified the build-push-containers to trigger and check a workflow in the movement-infra org that deploys the movement-network helm chart to the kubernetes clusters in the devNet environment. This happens after the container checks so that it's guaranteed the containers are built and tested.

# Testing

I've done some pretty extensive testing with triggering the workflow both in this repo and the infrastructure repo. I added error handling in the pipeline to make debugging failed deployments a easier.

But once there's a commit on main or a branch is tagged to build the containers we should see this pipeline trigger the devNet deploy.

# Outstanding issues
The helm chart sometimes errors out but this shouldn't interfere with the deployment pipeline but will inform in the build-push-containers workflow that triggers the deploy.

### DIFF
--- a/.github/workflows/build-push-containers-all.yml
+++ b/.github/workflows/build-push-containers-all.yml
@@ -1023,7 +1023,7 @@ jobs:
         RUN_ID: ${{ steps.get_run_id.outputs.run_id }}
         TOKEN: ${{ secrets.INFRA_GH_TOKEN }}
       run: |
-        API_URL="https://api.github.com/repos/${OWNER}/${REPO}/actions/runs/${run_id}"
+        API_URL="https://api.github.com/repos/${OWNER}/${REPO}/actions/runs/${RUN_ID}"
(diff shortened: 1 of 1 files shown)


radupopa369 (approved):

0xmovses:
This might sound silly, as it may have already been decided. Do we always want main to be devnet latest? Will there be a seperate branch then for mainnet (presumably a fair few versions behind devnet / main)

0xmovses:
Also 
>or a branch is tagged to build the containers we should see this pipeline trigger the devNet deploy

I don't think we want this functionality, often branches use this tag for CI testing purposes, we don't want to automically enable devnet deploys from a container build job.

PR #988, opened 2025-01-08
Convert DER signature to raw signature for AWS KMS signing

andygolay:
# Summary
- The goal of this PR is to be able to fetch the key Id from AWS Parameter Store, by submitting the canonical key (e.g. `/movement/prod/full_node/mcr_settlement/signer/weth-transfer-sign/replica-1`).
- When I run `cargo build --release` in `demo/hsm` I get 

```
error[E0308]: mismatched types
     --> util/signing/providers/aws-kms/src/hsm/mod.rs:39:49
      |
39    |                 let ssm_client = SsmClient::new(&shared_config);
      |                                  -------------- ^^^^^^^^^^^^^^ expected `&Config`, found `&SdkConfig`
      |                                  |
      |                                  arguments to this function are incorrect
      |
      = note: expected reference `&aws_types::config::Config`
                 found reference `&SdkConfig`
```

It seemed likely to be an incompatibility between versions of the crates `aws_sdk_ssm` and `aws_sdk_kms`. However I can't seem to pin the right versions; on this latest build it throws

```
error: rustc 1.79.0 is not supported by the following packages:
  aws-sdk-kms@1.54.0 requires rustc 1.81.0
  aws-sdk-ssm@1.59.0 requires rustc 1.81.0
  aws-sdk-sso@1.53.0 requires rustc 1.81.0
  aws-sdk-ssooidc@1.54.0 requires rustc 1.81.0
  aws-sdk-sts@1.54.0 requires rustc 1.81.0
  home@0.5.11 requires rustc 1.81
Either upgrade rustc or select compatible dependency versions with
`cargo update <name>@<current-ver> --precise <compatible-ver>`
where `<compatible-ver>` is the latest version supporting rustc 1.79.0
```

But I believe I'm only using `aws-sdk-ssm` 1.0.0 installed, per my `Cargo.toml` in the `movement-signer-aws-kms` crate. I ran `cargo clean`, removed `Cargo.lock` and `cargo update`.                  

I'm looking for a way to make the versions of each AWS dependency compatible so that the above "mismatched types" error is resolved.

# Changelog

# Testing

# Outstanding issues


util/signing/providers/aws-kms/Cargo.toml
```diff
@@ -10,10 +10,13 @@ publish = { workspace = true }
 rust-version = { workspace = true }
 
 [dependencies]
-movement-signer = { workspace = true }
+anyhow = { workspace = true }
 aws-sdk-kms = { workspace = true }
+aws-sdk-ssm = "1.0.0"
```
musitdev, line 15: Why you use a different version from `aws-sdk-kms`. In workspace `Cargo.toml ``aws-sdk-kms` use the version `1.51.0`. Did you try to use the same (` aws-sdk-ssm = "1.51.0"`) ?
Did you do a Cargo clean to test? Save the Cargo.lock before because you can have some issue with the crate and compiler compatibilities.

andygolay: If I try setting in the workspace `Cargo.toml` 

```
[workspace.dependencies]
aws-sdk-kms = "1.51.0"
aws-sdk-ssm = "1.51.0"
aws-config = "1.0.0"
aws-types = "1.0.0"
```

and then I delete the `Cargo.lock` and run `cargo clean` then `cargo build` from the workspace root, for some reason the `Cargo.lock` still shows up with for example

```
[[package]]
name = "aws-sdk-kms"
version = "1.54.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "a6cf16c0e5853312995505557b876dd3f9fb9941e96d031383528ccef14ace57"
dependencies = [
 "aws-credential-types",
 "aws-runtime",
 "aws-smithy-async",
 "aws-smithy-http",
 "aws-smithy-json",
 "aws-smithy-runtime",
 "aws-smithy-runtime-api",
 "aws-smithy-types",
 "aws-types",
 "bytes 1.9.0",
 "http 0.2.12",
 "once_cell",
 "regex-lite",
 "tracing",
]

[[package]]
name = "aws-sdk-ssm"
version = "1.59.0"
source = "registry+https://github.com/rust-lang/crates.io-index"
checksum = "935f454e645bcb5d6dfdda0bdf1f2c65cac06c57e876038a2ccf584cb5ef403c"
dependencies = [
 "aws-credential-types",
 "aws-runtime",
 "aws-smithy-async",
 "aws-smithy-http",
 "aws-smithy-json",
 "aws-smithy-runtime",
 "aws-smithy-runtime-api",
 "aws-smithy-types",
 "aws-types",
 "bytes 1.9.0",
 "fastrand 2.3.0",
 "http 0.2.12",
 "once_cell",
 "regex-lite",
 "tracing",
]
```

I don't see why `Cargo.lock` would show a different version than the `Cargo.toml`.

musitdev: This is the config of my Cargo.toml:
```
aws-types = "1.3.3"
aws-config = { version = "1.1.7", features = ["behavior-version-latest"] }
aws-sdk-s3 = "1.42.0"
aws-sdk-kms = "1.51.0"
```
Perhaps you should force everything with the later version.

l-monninger: @andygolay Have you given this a try? This sort of thing should indeed be why your build is failing. What Philippe has above is from the docs: https://docs.rs/aws-sdk-ssm/1.51.0/aws_sdk_ssm/index.html#getting-started, but if you ever don't have that you can always check the version of the crate being used in the actual `cargo tree` or the [implementation](https://github.com/awslabs/aws-sdk-rust).

util/signing/providers/aws-kms/src/hsm/mod.rs
```diff
@@ -1,109 +1,143 @@
...
+                self.key_id = key_id;
+        }
+
+        /// Tries to create a new AWS KMS HSM from the environment
+        pub async fn try_from_env() -> Result<Self, anyhow::Error> {
+                // Load shared AWS configuration
+                let shared_config: SdkConfig = aws_config::load_from_env().await;
+
+                // Fetch the canonical key ID from Parameter Store
+                let canonical_key = "/movement/prod/full_node/mcr_settlement/signer/weth-transfer-sign/replica-1";
```
l-monninger: Yes, this works, by why are you hardcoding this in? You should wrap in up to the CLI even for our testing so that others can participate.

l-monninger (changes requested):

util/signing/providers/aws-kms/src/hsm/mod.rs
```diff
@@ -1,96 +1,74 @@
...
-                println!(
-                        "Signing request: key_id={}, algorithm={}, message={:?}",
-                        &self.key_id,
-                        C::signing_algorithm_spec(),
-                        message
-                );
+                println!("Preparing to sign message. Message bytes: {:?}", message);
 
                 let blob = Blob::new(message);
+                let key_id = "arn:aws:kms:us-east-1:052940064682:key/9f228b96-d9f0-4597-bd7f-9f6dce6a71d9"; // Replace with dynamic key retrieval if needed
```
l-monninger, line 66: Please avoid hardcoding even as you are building. This key_id should be baked into the struct, so you should be updating the builder or the constructors.

l-monninger (changes requested):

util/signing/providers/aws-kms/src/hsm/mod.rs
andygolay: Ah I missed this comment, okay, will redo.

util/signing/providers/aws-kms/src/hsm/mod.rs
andygolay: Removed it. I'll use this and @musitdev's comments as guidance when we get to integrating Parameter Store.

PR #992, opened 2025-01-09
Implement-74

Primata:
# Summary

Implements MIP-74 specs

# Changelog

adds outbound rate limit
adds PAUSER_ROLE

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/bridge/contracts/src/NativeBridge.sol
+++ b/protocol-units/bridge/contracts/src/NativeBridge.sol
@@ -21,7 +21,9 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
     mapping(uint256 day => uint256 amount) public inboundRateLimitBudget;
 
     bytes32 public constant RELAYER_ROLE = keccak256(abi.encodePacked("RELAYER_ROLE"));
-    uint256 public constant MINIMUM_RISK_DENOMINATOR = 3;
+
+    // Risk denominator must be above 3
+    uint256 public constant RISK_DENOMINATOR_LOWER_BOUND = 3;
     IERC20 public moveToken;
     address public insuranceFund;
     uint256 public riskDenominator;
@@ -54,7 +56,7 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
 
         // Set insurance fund
         insuranceFund = _insuranceFund;
-        riskDenominator = MINIMUM_RISK_DENOMINATOR + 1;
+        riskDenominator = RISK_DENOMINATOR_LOWER_BOUND + 1;
 
         // Maintainer is optional
         _grantRole(RELAYER_ROLE, _maintainer);
@@ -72,7 +74,7 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
         whenNotPaused
         returns (bytes32 bridgeTransferId)
     {
-        // Ensure there is a valid amount
+        // Ensure there is a valid amount`
         require(amount > 0, ZeroAmount());
         address initiator = msg.sender;
 
@@ -177,7 +179,7 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
      */
     function setRiskDenominator(uint256 _riskDenominator) external onlyRole(DEFAULT_ADMIN_ROLE) {
         // risk denominator must be at least 4
-        require(_riskDenominator > MINIMUM_RISK_DENOMINATOR, InvalidRiskDenominator());
+        require(_riskDenominator > RISK_DENOMINATOR_LOWER_BOUND, InvalidRiskDenominator());
         riskDenominator = _riskDenominator;
         emit RiskDenominatorUpdated(_riskDenominator);
     }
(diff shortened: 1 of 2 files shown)


protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -19,9 +19,13 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
     mapping(uint256 nonce => OutboundTransfer) public noncesToOutboundTransfers;
```
apenzk, line 19:
```suggestion
    // Inbound: Transfers from the bridge to the user (e.g., tokens released to the recipient).
    // Outbound: Transfers from the user to the bridge (e.g., tokens locked for bridging).
    mapping(uint256 nonce => OutboundTransfer) public noncesToOutboundTransfers;
```

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -19,9 +19,13 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
...
     mapping(bytes32 bridgeTransferId => uint256 nonce) public idsToInboundNonces;
     mapping(uint256 day => uint256 amount) public inboundRateLimitBudget;
+    mapping(uint256 day => uint256 amount) public outboundRateLimitBudget;
 
     bytes32 public constant RELAYER_ROLE = keccak256(abi.encodePacked("RELAYER_ROLE"));
-    uint256 public constant MINIMUM_RISK_DENOMINATOR = 3;
+    bytes32 public constant PAUSER_ROLE = keccak256(abi.encodePacked("PAUSER_ROLE"));
+
+    // Risk denominator must be above 3
+    uint256 public constant RISK_DENOMINATOR_LOWER_BOUND = 3;
```
apenzk, line 28: say what means risk denominator or what it represents

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -19,9 +19,13 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
     mapping(uint256 nonce => OutboundTransfer) public noncesToOutboundTransfers;
     mapping(bytes32 bridgeTransferId => uint256 nonce) public idsToInboundNonces;
     mapping(uint256 day => uint256 amount) public inboundRateLimitBudget;
+    mapping(uint256 day => uint256 amount) public outboundRateLimitBudget;
 
     bytes32 public constant RELAYER_ROLE = keccak256(abi.encodePacked("RELAYER_ROLE"));
-    uint256 public constant MINIMUM_RISK_DENOMINATOR = 3;
+    bytes32 public constant PAUSER_ROLE = keccak256(abi.encodePacked("PAUSER_ROLE"));
+
+    // Risk denominator must be above 3
```
apenzk, line 27: equal or above?

protocol-units/bridge/contracts/src/NativeBridge.sol
apenzk: lower bound implies 3 is included, but you state "must be above 3"

protocol-units/bridge/contracts/src/NativeBridge.sol
```diff
@@ -54,10 +59,12 @@ contract NativeBridge is AccessControlUpgradeable, PausableUpgradeable, INativeB
 
         // Set insurance fund
         insuranceFund = _insuranceFund;
-        riskDenominator = MINIMUM_RISK_DENOMINATOR + 1;
+        riskDenominator = RISK_DENOMINATOR_LOWER_BOUND + 1;
```
apenzk, line 62:
```suggestion
        riskDenominator = RISK_DENOMINATOR_LOWER_BOUND;
```

protocol-units/bridge/contracts/src/NativeBridge.sol
apenzk, line 27:
```suggestion
    // Risk denominator must be equal or above 3
```

protocol-units/bridge/contracts/src/NativeBridge.sol
apenzk, line 28:
```suggestion
    uint256 public constant RISK_DENOMINATOR_LOWER_BOUND = 4;
```

apenzk (changes requested):
looks good for an initial approach apart from 
-  in the sol contract the risk_denominator cannot be updated. I am ok with this but it is inconsistent with the move contract equivalent
- requesting a few minor changes for acceptance.

long term, with more complex approaches like sliding window, we should consider the length of the window, but for now setting the window length to 1day is a good minimal approach.

protocol-units/bridge/contracts/src/NativeBridge.sol
apenzk: here in sol it is fixed, whereas in move there is a function update_risk_denominator. 

This is inconsistent.

protocol-units/bridge/contracts/src/NativeBridge.sol
Primata: not good since the lowerbound determines the minimum excluding bound. It has to be 3 because the check has to be >, not >= (saves some computation).

protocol-units/bridge/contracts/src/NativeBridge.sol
Primata: we got setRiskDenominator here

Primata: how can we change the name but not change the code logic?

0xmovses:
It's failing a few checks. Will review once those are passing.

apenzk (changes requested):
Requesting some minor changes. Otherwise looks good

Also there is mention of AtomicBridge and NativeBridge. Could update the name also with this PR or separately.

PR #1014, opened 2025-01-22
Key rotation with WAL file

andygolay:
# Summary

# Changelog

# Testing

# Outstanding issues


util/signing/interface/src/key/mod.rs
```diff
@@ -55,7 +55,7 @@ impl TryFromCanonicalString for Environment {
 	fn try_from_canonical_string(s: &str) -> Result<Self, String> {
 		match s {
 			"prod" => Ok(Environment::Prod),
-			"dev" => Ok(Environment::Dev),
+			"devNet" => Ok(Environment::Dev),
```
l-monninger, line 58: Keep this as dev.

util/signing/providers/aws-kms/src/hsm/mod.rs
```diff
@@ -31,8 +33,7 @@ where
 
 	/// Tries to create a new AWS KMS HSM from the environment
 	pub async fn try_from_env() -> Result<Self, anyhow::Error> {
-		let key_id = std::env::var("AWS_KMS_KEY_ID").context("AWS_KMS_KEY_ID not set")?;
-
+		let key_id = std::env::var("AWS_KMS_KEY_ID").unwrap_or("0x1".to_string());
```
l-monninger, line 36: You're going to want to pull changes from the `e2e-integration` branch at some point.

util/signing/signing-admin/src/cli/wal.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::fs::{self, OpenOptions};
+use std::io::{BufReader, BufWriter};
+use std::path::Path;
+
+pub const WAL_FILE: &str = "wal.json";
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub struct WalEntry {
```
l-monninger, line 10: For our `Wal` file, we're going to want to want to describe which messages have been passed back and forth and move them from an `prepared` queue to a `committed` queue as we know they succeed. This will allow us to replay. 

Thus, I would recommend the `WalEntry` have those two fields.

You will also likely want to generated a UID.

util/signing/signing-admin/src/cli/wal.rs
```diff
@@ -0,0 +1,66 @@
...
+}
+
+pub fn append_to_wal(entry: WalEntry) -> Result<()> {
+                let mut entries = read_wal()?;
+                entries.push(entry);
+                write_wal(&entries)
+}
+
+pub fn update_wal_status(canonical_string: &str, new_status: &str) -> Result<()> {
+                let mut entries = read_wal()?;
```
l-monninger, line 46: What your doing here is not a WAL. It's a key-value serialization of the key state. This can also be useful, but it does not describe how to recover from failed procedures.

util/signing/signing-admin/src/cli/wal.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
+use serde::{Deserialize, Serialize};
+use std::fs::{self, OpenOptions};
+use std::io::{BufReader, BufWriter};
+use std::path::Path;
+
+pub const WAL_FILE: &str = "wal.json";
```
l-monninger, line 7: Don't use a hard-coded constant. This should eventually be brought up as a CLI arg.

util/signing/signing-admin/src/cli/wal.rs
```diff
@@ -0,0 +1,66 @@
...
+                let mut entries = read_wal()?;
+                for entry in &mut entries {
+                                if entry.canonical_string == canonical_string {
+                                                entry.status = new_status.to_string();
+                                }
+                }
+                write_wal(&entries)
+}
+
+pub fn update_wal_entry<F>(canonical_string: &str, update_fn: F) -> Result<()>
```
l-monninger, line 55: There's no implementation here for actually continuing from the WAL file.

util/signing/signing-admin/src/cli/wal.rs
```diff
@@ -0,0 +1,66 @@
+use anyhow::Result;
```
l-monninger, line 1: Basically, a WAL file should be a queue of "transactions" with our system. When one completes successfully, you should pop it off. 

For simplicity, we will only process on transaction at a time. Furthermore, the transaction should always complete unless manual intervention takes place. 

So, consider the following as an example of valid usage via the CLI.

1. Successful rotation writes to WAL but removes entry when complete
- **Command:**
```
signing-admin rotate-key --alias always --identifier new --url https://app --wal wal.file
# exits with code 0
```

- **WAL:**
```
# empty
```

2. Unsuccessful rotation leaves log of where it left off. 
- **Command:**
```
signing-admin rotate-key --alias always --identifier new --url https://app --wal wal.file
# exits with code 1
```
- **WAL:**
```
{
    "log" : [
      {
         "id" : "unique-id"
         "messages" : {
              "committed" : [
                    "START_ROTATION",
                    "SEND_APP_ROTATE",
                    "RECV_APP_ROTATE",
                    "SEND_HSM_ROTATE"
               ], 
              "prepared" : [
                    "RECV_HSM_ROTATE"
               ]
         }
      }
    ]
}
```

3. Someone tries to run a over the unfinished log and it prevents
- **Command:**
```
signing-admin rotate-key --alias always --identifier new --url https://app --wal wal.file
# exits with code 1; reports unfinished WAL
```
- **WAL:**
```
{
    "log" : [
      {
         "id" : "unique-id"
         "messages" : {
              "committed" : [
                    "START_ROTATION",
                    "SEND_APP_ROTATE",
                    "RECV_APP_ROTATE",
                    "SEND_HSM_ROTATE"
               ], 
              "prepared" : [
                    "RECV_HSM_ROTATE"
               ]
         }
      }
    ]
}
```

4. Operator comes along to correct and complete
- **Command:**
```
signing-admin recover
# reads through WAL file for remaining messages
```
- **WAL:**
```
# empty because the messages were complete
```

l-monninger (changes requested):

PR #1019, opened 2025-01-27
SignerIdentifier::try_raw_private_key

mzabaluev:
Add a method to extract private key bytes from
`SignerIdentifier`, to use in tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10684,6 +10684,7 @@ version = "0.0.2"
 dependencies = [
  "anyhow",
  "async-trait",
+ "hex",
  "movement-signer",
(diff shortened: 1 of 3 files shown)


mzabaluev:
The test failure is also seen in #993.

l-monninger:
This should basically already be merged in by #1024 . Not exactly sure why the diff, but merging.

Cargo.lock
```diff
@@ -10684,6 +10684,7 @@ version = "0.0.2"
 dependencies = [
  "anyhow",
  "async-trait",
+ "hex",
  "movement-signer",
```
Farristom, line 10688: I'm signing

PR #1020, opened 2025-01-27
Cargo.toml to use `branch = movement`

0xmovses:
# Summary
Self explanatory

# Changelog
Updates all aptos deps, this method should always be used going forward, to get latest changes from `aptos-core/movement` 

# Testing
CI check

# Outstanding issues


mzabaluev (approved):
Good idea, as long as we also keep `Cargo.lock` under version control and so the revision is locked there.

This will not solve merge conflicts over `Cargo.lock`, but those are easily solved by checking the file out cleanly from one of the bases and running a rebuild with cargo.

0xmovses:
> This will not solve merge conflicts over Cargo.lock, but those are easily solved by checking the file out cleanly from one of the bases and running a rebuild with cargo.

This doesn't always work out, as some manual cargo commands require to pin some deps to a certain version. But yes, better. 

I manually resolved about 50 conflicts in the Cargo.lock recently, (nuking the Cargo.lock was breaking the build for me). Yikes.

0xmovses:
As mentioned by @l-monninger Cargo doesn't detect upstream changes automatically with the branch method, I thought it did for some reason. So, we stick to revision commit hashes.

PR #1038, opened 2025-02-06
Celestia mainnet config

### LINKED ISSUE #1030: Add direct support for Celestia Mainnet Service Compositions
# Summary
We have thusfar relied on manual reconfiguration to support using Celestia Mainnet as DA. This is, however, easy to break with our current delivery pipeline. To resolve this, we need to make changes to both the application and our pipeline. 

This issue details the changes needed on the application side. 

## Requirements
1. The [`celestia-light`](https://github.com/movementlabsxyz/movement/tree/main/protocol-units/da/movement/celestia/runners/src/celestia_light) runner MUST support a [mainnet](https://docs.celestia.org/how-to-guides/light-node#initialize-the-light-node) subcommand. 
2. The [`celestia-light`](https://github.com/movementlabsxyz/movement/tree/main/protocol-units/da/movement/celestia/runners/src/celestia_light) runner MUST load a Celestia private key from disk (by passing its path as an arg to the Celestia light command). The path to this key shall be be provided in the config. 
3. The [`celestia-light`](https://github.com/movementlabsxyz/movement/tree/main/protocol-units/da/movement/celestia/runners/src/celestia_light) runner MUST idempotently initialize the light node. This can also be built into the setup script, but such is more complicate.
4. The process-compose `movement-full-node` runners MUST be configured with a [`celestia-mainnet`](https://github.com/movementlabsxyz/movement/tree/main/process-compose/movement-full-node) target. This runner MUST remove unnecessary setup and runtime processes, such as the bridge node. 
5. The docker-compose `movement-full-node` MUST be configured with a [`celestia-mainnet`](https://github.com/movementlabsxyz/movement/tree/main/docker/compose/movement-full-node) target. This runner MUST remove unnecessary setup and runtime processes, such as the bridge node.  
6. Either the setup or `celestia-light` runner SHOULD use a [trusted hash](https://docs.celestia.org/how-to-guides/celestia-node-trusted-hash) to reduce sync time.

mzabaluev:
Resolves #1030

# Summary
- **Categories**: `protocol-units`, `scripts`, `util`.

Add "mainnet" configuration and setup infrastructure for Celestia DA.

Redo the celestia DA configuration: transpose the network enum and the parameter struct, so that network is a field. This makes for a lot less code duplication.


l-monninger:
# Next Steps

## Testing e2e
Test against Celestia mainnet. Use an arbitrary namespace. You may request TIA from myself. 

## Key Info
The keys used should be available into the `celestia-appd/keys` under `validator.info` and `<matching-address>.address`. Per #1030 , I would like this to be populated from some value in our config containing serialized or unserialized key material. This makes it easier than brittle file redirection for the moment. 

## Keyring Naming
It is fine to leave as test, but you can also rename this as you need. I do not believe using a keyring with a name other than `"test"` actually changes the encoding used.

## Providers
- consensus.lunaroasis.net
- public-celestia-rpc.numia.xyz
- rpc.celestia.pops.one

mzabaluev:
> ## Key Info
> The keys used should be available into the `celestia-appd/keys` under `validator.info` and `<matching-address>.address`. Per #1030 , I would like this to be populated from some value in our config containing serialized or unserialized key material. This makes it easier than brittle file redirection for the moment.

The way it can work with the `cel-key` utility is, we export the key in PGP ASCII armor, protected by a random passphrase generated by setup. Then the `cel-key import` command can import the key to be consumable by celestia node(s).

> ## Keyring Naming
> It is fine to leave as test, but you can also rename this as you need. I do not believe using a keyring with a name other than `"test"` actually changes the encoding used.

There are no named keyrings, but you can select backends. I think the only backend available on Linux is `test` (there is also `memory`, but it's for in-process use).

mzabaluev:
Relevant configuration via environment variables given to the setup binary that creates config for movement-celestia-da-light-node and movement-celestia-light runner:

* `CELESTIA_NETWORK`, with values `mainnet`, `mocha`, `arabica`, or local (the default).
* `CELESTIA_NAMESPACE`: hexadecimal namespace suffix for a v0 ID (up to 10 bytes) if the value starts with "0x", otherwise decoded as the base64 encoding of the full 29-byte namespace ID.
  * **WARNING:** the `=` padding character of base64 is known to be misinterpreted when used in the environment specification in process-compose; use an `.env` file.
  * If you are testing with public Celestia networks, be sure to pick a randomly generated namespace.
* `CELESTIA_LIGHT_KEYNAME`: name of the key in the `test` cosmos-sdk keyring, that can be provisioned with [cel-key](https://docs.celestia.org/tutorials/celestia-node-key). The default key name is "movement_celestia_light".

The command to run the celestia light node (unless using the Movement runner) can be obtained from the [runner source code](https://github.com/movementlabsxyz/movement/tree/c586b804eb83e3a1acdade53bea36dc776a9472b/protocol-units/da/movement/protocol/celestia-runners/src/celestia_light) for the respective network.

For Docker, see [this guide](https://docs.celestia.org/tutorials/celestia-node-key#running-your-node), including the section on mounting the keyring.

I may have time to add this to the docker-compose files in the repository.

musitdev (approved):
I have a practical question. For local execution is it still the same as before?

mzabaluev:
> I have a practical question. For local execution is it still the same as before?

Yes, it should work the same way. I have tested this while working on the branch, but not on the latest commits.

musitdev (approved):

PR #996, opened 2025-01-13
Fixes Issues w/ Stream Crashing Over Large Replay

l-monninger:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `protocol-units`
- **Issues**:  #1003

Adds fixes and features for debugging issues with Movement DA Light Node Stream.

- **Problem:** over large replays, the stream from Celestia was occasionally dropped owing to (a) overfetching against the Celestia API which (b) could result in errors that were incorrectly propagated and closed the stream. 
- **Solution:** this PR addresses streaming issues by softening error propagation. It also adds debugging tools. 
- **Other Common Problems:** when debugging in a production context, transport issues may occasionally arise. HTTP/2 used with gRPC streams does not always play nicely with CloudFlare and AWS ALB in particular. The updated production setup avoids routing HTTP/2 traffic through either of these, instead preferring a direct connection. 

**Review the change log to understand useful debugging features!!!*

# Change Log
- **Adds [`movement-full-node da stream-blocks`](https://github.com/movementlabsxyz/movement/blob/l-monninger/stream-size-fix/networks/movement/movement-full-node/src/da/stream_blocks/mod.rs):** this allows for direct monitoring of DA Stream to an arbitrary connection. 
```
Usage: movement-full-node da stream-blocks --movement-path <MOVEMENT_PATH> <LIGHT_NODE_URL> <FROM_HEIGHT>
```
- **Refactors Underlying DA to Better Manage Non-fatal Errors for Stream:** this adds a [generalized stream trait and default implementations](https://github.com/movementlabsxyz/movement/blob/l-monninger/stream-size-fix/protocol-units/da/movement/protocol/da/src/lib.rs) as well as a [mock](https://github.com/movementlabsxyz/movement/blob/l-monninger/stream-size-fix/protocol-units/da/movement/protocol/da/src/mock/mod.rs) for debugging general issues.
> [!TIP]
> As a rule of thumb, in so far as tests against the generalized version of the stream with the mock pass, issues with stream disconnects are either (a) transport issues or (b) implementation issues stemming from a case where an error in the stream should be non-fatal. 
- **Introduces [Digest Store](https://github.com/movementlabsxyz/movement/blob/l-monninger/stream-size-fix/protocol-units/da/movement/providers/digest-store/src/da/mod.rs):** this digests blobs before sending them to the DA, e.g. Celestia, and looks them up when rebuilding the stream. Importantly, this removes the need for the blob submission heuristics, as all blobs are now a fixed size. 
[...]

PR #1064, opened 2025-02-22
[DA] Update light client protocol and heartbeat.

musitdev:
# Summary

* Add heartbeat to the DA light node protocol. Use by follower node stay synced.
* Add debug trait to DA blob to facilitate logging.
* Add source error to error messages.
# Changelog

# Testing

# Outstanding issues


musitdev:
The test `test_execute_block_state_get_api` never end, it's not related to this PR.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
```diff
@@ -0,0 +1,99 @@
+syntax = "proto3";
+package movementlabs.protocol_units.da.light_node.v1beta2;
```
mzabaluev, line 2: Do we need an entirely new version of the protocol?

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
```diff
@@ -0,0 +1,99 @@
...
+    uint64 timestamp = 5;
+    bytes signer = 6;
+}
+
+message BlobResponse {
+    oneof blob_type {
+      Blob passed_through_blob = 1;
+      Blob sequenced_blob_intent = 2;
+      Blob sequenced_blob_block = 3;
+      bool heartbeat_blob = 4;
```
mzabaluev, line 20: Adding a new field should be a backward-compatible change in protobuf, so this change alone does not justify the `.v1beta2` package.

musitdev: It has been discussed here: https://github.com/movementlabsxyz/movement/pull/996#pullrequestreview-2559597210
And we go to a new version of the protocol.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
musitdev: It has been discussed here: https://github.com/movementlabsxyz/movement/pull/996#pullrequestreview-2559597210
And we go to a new version of the protocol.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
musitdev: If you send a heartbeat with an old client it's not compatible. I think it's more theto define a revision in the protocol.

networks/movement/movement-full-node/src/da/stream_blocks/mod.rs
```diff
@@ -38,13 +38,30 @@ impl StreamBlocks {
...
 					anyhow::bail!("Invalid blob type in response")
 				}
 			};
-			info!("{} {}  {}", hex::encode(block_id), block_timestamp, da_height);
+			// pretty print (with labels) the block_id, block_timestamp, and da_height
+			tracing::info!(
+				"Block ID: {}, Block Timestamp: {}, DA Height: {}",
+				hex::encode(block_id),
+				// unix date string from the block timestamp which is in microseconds
+				chrono::DateTime::from_timestamp((block_timestamp / 1_000_000) as i64, 0)
```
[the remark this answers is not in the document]
musitdev: I copy past from another code and didn't think about it. I do the change.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
mzabaluev, line 20: I agree with @l-monninger in https://github.com/movementlabsxyz/movement/pull/996#discussion_r1920533395, this should not be named "something_blob".

```suggestion
      bool heartbeat = 4;
```

mzabaluev: In general, why multiplex this in repeated blob responses?
Wouldn't it be better to just create a heartbeat method?

protocol-units/da/movement/protocol/da/src/lib.rs
```diff
@@ -54,7 +54,7 @@ pub enum DaError {
 /// Trait for DA operations.
 pub trait DaOperations<C>: Send + Sync
 where
-	C: Curve + Send + Sync + 'static,
+	C: Curve + Send + Sync + 'static + std::fmt::Debug,
```
mzabaluev, line 57: I don't see any code that requires the `Debug` bound on the trait itself.

protocol-units/da/movement/protocol/da/src/mock/mod.rs
```diff
@@ -64,7 +64,7 @@ where
 
 impl<C> DaOperations<C> for Mock<C>
 where
-	C: Curve + Send + Sync + 'static,
+	C: Curve + Send + Sync + 'static + std::fmt::Debug,
```
mzabaluev, line 67: Why was this bound added here, with no new code to require it?

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
```diff
@@ -35,7 +35,8 @@ where
 		+ Serialize
 		+ for<'de> Deserialize<'de>
 		+ Clone
-		+ 'static,
+		+ 'static
+		+ std::fmt::Debug,
```
mzabaluev, line 39: What's the cause of this `Debug` bound pollution everywhere?

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
```diff
@@ -135,31 +138,51 @@ where
...
+										continue;
+									}
+								}
+							}
+							Some(Err(err)) => {
+								tracing::warn!("Stream blob return an error, exit stream :{err}");
+								return;
+							},
+							None => {
+								info!("Stream blob closed , exit stream.");
```
mzabaluev, line 170: Can we do this at `debug` level, to be enabled when needed? The debug spam in regular logs is getting annoying.

protocol-units/da/movement/protocol/proto/build.rs
```diff
@@ -1 +1,4 @@
-buildtime::proto_build_main!("movementlabs/protocol_units/da/light_node/v1beta1.proto");
+buildtime::proto_build_main!(
+	"movementlabs/protocol_units/da/light_node/v1beta1.proto",
```
mzabaluev, line 2: `.v1beta1` is just dead proto source now and will decay.

mzabaluev:
> Add heartbeat to the DA light node protocol. Use by follower node stay synced.

Sorry, this is an insufficient explanation as to why we need it as a new type of "nothing blob", to the extent that we have to have a protocol break (which is not actually warranted by the proto-level changes introduced, as I already pointed out). Were HTTP/2 keepalives attempted to be used?

protocol-units/da/movement/protocol/da/src/mock/mod.rs
musitdev: Yes it has been removed because block debug for testing has been removed but as it can be useful to be able to debug the Da I keep it.

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
musitdev: To debug DA block content, we need to add the constraint on C everywhere. DaBlob<C> depend on C.

musitdev: For test and event to debug to be able to see DA block is important.

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
musitdev: I think it's not an info but a warning. This shouldn't arrive, and in this case the connection must be restarted. It a dev that we need to do, manage DA connection issue to keep the node alive.

musitdev:
We spend some time to make the grpc works. In prod the grpc stream connection was timouted and what ever the parameter we use it doesn't work. I think it's mostly because the connection between. The follower that get block and the leader use a http2 to a ngnix proxy then grpc. So grpc heartbeat doesn't maintain the http2 connection.
We add other strange issue that need to detect that the connection is still alive, and we don't miss a block. So the heartbeat on the block stream was the easiest and best solution.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
musitdev: I don't see the @l-monninger about the blob name the only one I see is
> I would also not make this a Blob type.
Which mean that the `heartbeat_blob` field should not be of type blob but bool.
The names has been defined by him.

protocol-units/da/movement/protocol/da/src/mock/mod.rs
mzabaluev: Well, it does not do anything in this particular impl, so it may be better to remove it unless it's forced by the other bounds on `C` that were added elsewhere.

I suspect it's just some lazy `#[derive(Debug)]` that requires the phantom curve parameter to have `Debug` as well, even though it's not needed.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
musitdev: The heart beat must be in the stream. We get idle connections, the pull heartbeat request can work and the stream be idle.

protocol-units/da/movement/protocol/da/src/mock/mod.rs
musitdev: All the debug bound I've added has been requested by the compiler when I add some debut code to understand the DA issues.
I keep them because it can be useful to sort DA issue.

networks/movement/movement-full-node/src/da/stream_blocks/mod.rs
```diff
@@ -38,13 +38,30 @@ impl StreamBlocks {
 				.ok_or(anyhow::anyhow!("No blob type in response"))?
```
mzabaluev, line 38: It's better to handle `None` in a graceful manner here, which would allow future changes in the protocol without having to make a package name break again.

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
```diff
@@ -167,7 +167,7 @@ where
 								return;
 							},
 							None => {
-								info!("Stream blob closed , exit stream.");
+								tracing::warn!("Stream blob closed , exit stream.");
```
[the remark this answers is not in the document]
musitdev: Yes, it mostly mean that the light node closed the stream which is abnormal.

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
l-monninger: I don't care too much about field naming. My comment was about the structure. But, of course, if you are just making it a bool, it should not be a `*_blob`.

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
l-monninger: Warn is probably better.

l-monninger:
> We spend some time to make the grpc works. In prod the grpc stream connection was timouted and what ever the parameter we use it doesn't work. I think it's mostly because the connection between. The follower that get block and the leader use a http2 to a ngnix proxy then grpc. So grpc heartbeat doesn't maintain the http2 connection.
> We add other strange issue that need to detect that the connection is still alive, and we don't miss a block. So the heartbeat on the block stream was the easiest and best solution.

@mzabaluev When your proxy gRPC http2 streams through Nginx, Nginx will drop streams where application messages are not being exchanged.

protocol-units/da/movement/protocol/light-node/src/passthrough.rs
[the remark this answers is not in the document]
musitdev: It must be present in the log and visible. For example, I got an idle connection (no block are received anymore). In debug mode, I miss the info and in info, it's hard to identify the cause of the idleness. So warn is more appropriate.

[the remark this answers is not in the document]
mzabaluev: OK. The levels are cumulative though, so with the debug fliter you won't miss any INFO logs. Or is it just the amount of messages that make it difficult?

proto/movementlabs/protocol_units/da/light_node/v1beta2.proto
musitdev: Done

mzabaluev (approved):

radupopa369 (approved):

PR #1069, opened 2025-02-26
test: fix unit tests

mzabaluev:
# Summary

Reverting the LocalAccount initialization to the way it was done before remote signing (#993) was merged makes the tests pass, though I don't yet understand why.

# Testing

```
nix develop --extra-experimental-features "nix-command flakes" --command bash -c "cargo test -p maptos-opt-executor"
```

# Outstanding issues

Why does use of `aptos_test_root_address()` make the transactions be accepted by the executor, while the address and the signing key don't actually match?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12219,6 +12219,7 @@ dependencies = [
  "async-trait",
  "chrono",
  "ed25519-dalek 2.1.1",
+ "hex",
  "maptos-dof-execution",
  "maptos-execution-util",
  "movement-signer",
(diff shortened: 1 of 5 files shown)


musitdev (approved):
This solved the test error? So it's not related to signing API?

l-monninger:
> Why does use of aptos_test_root_address() make the transactions be accepted by the executor, while the address and the signing key don't actually match?

The way Aptos core addresses work is an additional mapping or indirection atop the cryptographically derived address. In fact, since they introduced rotation, this is available to all accounts.

The address when you are signing root transactions is fixed to `0x1` or `0xa550c16`

musitdev (approved):

PR #1072, opened 2025-02-27
fix: add back labeled builds

radupopa369:
# Summary

build containers when a label is applied and not for all commits

## Changes

1. Add back the labeled builds `cicd:movement-full-node`
2. Add back labeled builds `cicd:bridge`
3. Do not run workflows twice, once for a pull request and once for a git push for every branch.
4. Increase the runners size

## Proposal of optimisation:  Trigger workflows only for Pull Requests events

If one opens a PR, one single push commit triggers for 2 events, one for a Pull request  and one for a push on a branch.

image

```yaml
on:
  pull_request:
    types:
      - labeled
      - opened
      - synchronize
  push:
    branches:
      - '**' 
```

# Outstanding issues
[build containers when a label is applied and not for all commits](https://github.com/movementlabsxyz/movement/issues/1073)

## Building only un-checked containers when there are no labels
image


.github/workflows/build-push-containers-all.yml
mcmillennick, line 1: I think we want to run the builds on the 16 cpu variant

cc. @l-monninger

.github/workflows/hsm-demo-containers.yml
mcmillennick, line 1: @andygolay @l-monninger Do we still need to be building the HSM demo container?

.github/workflows/checks-all.yml
```diff
@@ -7,25 +7,38 @@ on:
       - opened
       - synchronize
   push:
-    branches:
-      - '**'
```
mzabaluev, line 11: Excluding development branches means the CI will only run checks on auto-merged checkouts when commits are pushed to the PR. So effectively the workflow will test the code that would result if this PR is merged to the base branch at this point. This is a workflow choice that needs to be articulated.

.github/workflows/build-push-containers-all.yml
```diff
@@ -8,13 +8,25 @@ on:
       - opened
       - synchronize
   push:
-    branches:
-      - '**'
```
mzabaluev, line 12: The same applies to building containers: if you use the checkout action in a workflow triggered by `pull_request`, you need to specify the SHA of the revision on the branch, not the merge result, otherwise it may not be the same code that was pushed.

.github/workflows/hsm-demo-containers.yml
radupopa369: maybe let's keep it until we implement secure signing ...

.github/workflows/build-push-containers-all.yml
radupopa369: We discussed this detail in the slack thread. I shared the builds from event from pull request and from push. Both had the same sha because now we build containers using a new shell script and the tagging is done like this

[commit_hash=$(git rev-parse HEAD | cut -c1-7)
](https://github.com/movementlabsxyz/movement/blob/main/scripts/docker/build-push-container#L47)

Here a links to workflow runs that trigger for pull request and for git push

https://github.com/movementlabsxyz/movement/actions/runs/13574706447
https://github.com/movementlabsxyz/movement/actions/runs/13574706013

and if you search for a building of an image 
https://github.com/movementlabsxyz/movement/actions/runs/13574706013/job/37948003789#step:7:2427
https://github.com/movementlabsxyz/movement/actions/runs/13574706447/job/37948007268#step:7:2432
both are pushing the same tag `pushing manifest for ghcr.io/movementlabsxyz/movement-celestia-da-light-node:f65479f`

.github/workflows/checks-all.yml
radupopa369: I am not sure I understand what you mean here :).  Can you please provide more details?
If a PR is labeled, the checks will run. Is this not enough?
```yaml
  pull_request:
    types:
      - labeled
      - opened
      - synchronize

```

When I tested, checks run all the time, even after I pushed all the changes 
https://github.com/movementlabsxyz/movement/actions?query=branch%3Aradu-fix-build-labeled-only

mzabaluev: The checks will run, but the checkout will fetch the merge branch auto-created from the tip of the PR branch: https://github.com/movementlabsxyz/movement/actions/runs/13704740025/job/38327172716#step:2:53

If the base branch or the PR has advanced from the latest shared revision, this will not be the same source tree as what was pushed to trigger the workflow.

mzabaluev (approved):

.github/workflows/hsm-demo-containers.yml
radupopa369: ok then I will remove this file

andygolay:
Awesome, was waiting for resolution of Mikhail's issues to re-review, looks great though.

PR #1078, opened 2025-03-02
fix(docker): celestia-light-node fixes

mzabaluev:
# Summary
- **Categories**: `scripts`.

* Point to the consensus node in the infra.
* Specify the listen address for the RPC.
* Pass the `NODE_STORE` environment variable so that the endpoint calls init on the store.
* Improve the health check to check the service port with nc.


musitdev (approved):

cshein45 (approved):
[MOVE](https://movementnetwork.xyz)

radupopa369 (approved):

PR #1079, opened 2025-03-03
[Follower] Correct DA client http2 connection

musitdev:
# Summary

The DA client htp2 connection code wasn't merge in main. Put it back.

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/da/movement/protocol/client/src/http2.rs
+++ b/protocol-units/da/movement/protocol/client/src/http2.rs
@@ -1,4 +1,6 @@
 use movement_da_light_node_proto::light_node_service_client::LightNodeServiceClient;
+use std::time::Duration;
+use tonic::transport::{Channel, ClientTlsConfig};
 
 #[derive(Debug, Clone)]
 pub struct Http2 {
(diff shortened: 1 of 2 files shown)


radupopa369 (approved):

protocol-units/da/movement/protocol/client/src/http2.rs
```diff
@@ -8,7 +10,20 @@ pub struct Http2 {
 impl Http2 {
 	/// Connects to a light node service using the given connection string.
 	pub async fn connect(connection_string: &str) -> Result<Self, anyhow::Error> {
-		let client = LightNodeServiceClient::connect(connection_string.to_string()).await?;
+		let endpoint = Channel::from_shared(connection_string.to_string())?;
+
+		// Dynamically configure TLS based on the scheme (http or https)
+		let endpoint = if connection_string.starts_with("https://") {
```
mzabaluev, line 16: It would be nice to start using the `Url` domain type `tonic::transport`, as it has parsing implemented via `FromStr`, so you don't have to do naive string matching here.

mzabaluev (approved):

protocol-units/da/movement/protocol/client/src/http2.rs
```diff
@@ -8,7 +10,20 @@ pub struct Http2 {
...
+			endpoint
+				.tls_config(ClientTlsConfig::new().with_enabled_roots())?
```
mzabaluev, line 18: Also I think, TLS config can be applied unconditionally, the client will discriminate when it's needed.

protocol-units/da/movement/protocol/client/src/http2.rs
musitdev: Do we do it now, or you can open an issue and work on i later?

protocol-units/da/movement/protocol/client/src/http2.rs
musitdev: Do we do it now, or you can open an issue and work on i later?

radupopa369: let's do it later,

PR #1080, opened 2025-03-03
L moninger/mempool simplification

l-monninger:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
 [[package]]
 name = "abstract-domain-derive"
 version = "0.1.0"
-source = "git+https://github.com/movementlabsxyz/aptos-core?rev=109737d10424a7e0f04fbda329058cc581334faf#109737d10424a7e0f04fbda329058cc581334faf"
+source = "git+https://github.com/movementlabsxyz/aptos-core?rev=7908e34f0e9ffae2bd5fdb3702d4cabfa7f2b8b6#7908e34f0e9ffae2bd5fdb3702d4cabfa7f2b8b6"
 dependencies = [
  "proc-macro2",
  "quote",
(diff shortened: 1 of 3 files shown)


l-monninger:
The concern is essentially that, without the tolerance, I think users can submit way out into the future, so far that they might never realize gas fees.

This reintroduces the attack the tolerance amended.

And, I don't see anything directly in the mempool which actually prevents this.

If anything it's a side-effect of something in the VM

l-monninger:
This solves https://github.com/movementlabsxyz/movement/issues/490 insofar as we it doesn't allow the submission of duplicate transactions by sequence number. But, it's not yet clear to me that a user cannot submit transactions roughly arbitrarily out into the future that would abort on the prologue and not pay gas.

l-monninger:
@musitdev @0xmovses do you guys understand this PR?

0xmovses:
I haven't had the time to dig into it. I've perused over it a few times but I need to spend more time on it. After burn cap upgrade test for me.

0xmovses:
Immediate question is why we didn't see this in previous testnets or bardock. Or did we? Was it reported? If it wasn't it would suggest some regression entered the system that was not caught.

musitdev:
Yes I think. I've done some test and reproduce some issues.
For the first one, I'm not sure of the expected behavior.
The current behavior is:
If one account send 2 tx at the same time with 2 consecutive seq number 2 cases can arrive:
 * the Tx  are executed in order correctly.
 * The tx arrive in wrong order and one is rejected with SEQUENCE_NUMBER_TOO_NEW error. Still not sure with one is rejected; I've to add more logs.
 Currently, this error is not reported to the sequencer mempool so the test fail with error` timing out waiting for the transaction ` . I can solve it by reintroducing a synchronization primitive like you say.
 My question:
If several Tx from an account is received, they are not ordered, so Tx in the wrong order are rejected like the current behavior or do they need to be executed in the right order?

l-monninger:
> The tx arrive in wrong order and one is rejected with SEQUENCE_NUMBER_TOO_NEW error. Still not sure with one is rejected; I've to add more logs.
Currently, this error is not reported to the sequencer mempool so the test fail with errortiming out waiting for the transaction . I can solve it by reintroducing a synchronization primitive like you say.

It's unclear to me how this would happen and it hasn't showed up in runnings of the tests yet. Are you sure this isn't...

> when a Tx fail during excution for example Seq too new it's not notified to the mempool so you wait and you get 60 sec error. Try the sequence_number_ooo test, sometime it works and sometime not depending on which order the Tx are processed.

...as mentioned in the other chat? 

I would expect the current implementation to hit the following edge case:

If `transaction_2` arrives on the end of a mempool period tick, it might immediately get sent through. This actually similar to https://github.com/movementlabsxyz/movement/issues/321#issuecomment-2304190664.

PR #1105, opened 2025-03-12
chore: remove unused jupyter notebook

0xmovses:
## Changelog
- Removes unused jupyter notebook files. 

These files were are genned files that were used when researching and investigating old concurrency issues. They have not been used for some time. 

# Testing
Build should pass

# Outstanding issues
Have confirmed with other team mates that this is no longer in use.


mzabaluev:
The build job still fails in the aftermath of #1100, but it will be addressed in another PR.

PR #1111, opened 2025-03-14
Memseq Degradation

l-monninger:
# Summary
- **RFCs**: $\emptyset$.

Submits Celestia blobs in parallel. This resolves backpressuring and works in the context where you rely on Celestia for ordering. When Celestia is no longer used for ordering, e.g., WAL-Celestia, you will need to return to some kind of sequential submission to make sure these match. 

Also adds `submit_collapsed_block` to replace submission in parallel with submission of single batched block. This works under both ordering and non-ordering regimes of Celestia given submissions over the in-process channel are read and processed in order. 

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/da/movement/protocol/light-node/src/sequencer.rs
+++ b/protocol-units/da/movement/protocol/light-node/src/sequencer.rs
@@ -1,3 +1,4 @@
+use futures::future::join_all;
 use movement_da_light_node_da::DaOperations;
 use movement_da_light_node_prevalidator::{aptos::whitelist::Validator, PrevalidatorOperations};
 use movement_signer::cryptography::secp256k1::Secp256k1;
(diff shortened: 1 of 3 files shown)


util/movement-types/src/block.rs
```diff
@@ -106,6 +106,23 @@ impl Block {
 	pub fn add_transaction(&mut self, transaction: Transaction) {
 		self.transactions.insert(transaction);
 	}
+
+	pub fn collapse(blocks: Vec<Block>) -> Block {
+		let mut transactions = BTreeSet::new();
+		let parent = if let Some(first_block) = blocks.first() {
```
mzabaluev, line 112: The field is essentially bogus since we don't initialize the chain parent correctly on DA sequencer restart. But within movement-types this is the right thing to do.

util/movement-types/src/block.rs
```diff
@@ -106,6 +106,23 @@ impl Block {
...
+			first_block.parent
+		} else {
+			Id::genesis_block()
```
mzabaluev, line 115: Should we rather handle the empty case specially and return `Option<Block>` to avoid producing something out of nothing?

protocol-units/da/movement/protocol/light-node/src/sequencer.rs
```diff
@@ -163,13 +164,29 @@ where
...
+		});
+
+		join_all(futures).await.into_iter().collect::<Result<(), _>>()?;
+		Ok(())
+	}
 
+	/// Collapses the back-pressured blocks into a single block and submits it.
+	///
+	/// todo: mark not pub once this is actually used
+	pub async fn submit_collapsed_blocks(&self, blocks: Vec<Block>) -> Result<(), anyhow::Error> {
```
mzabaluev, line 185: Will we need this in the future?

musitdev (approved):

util/movement-types/src/block.rs
musitdev: All this function is as an example to explain the idea of batch aggregation.

util/movement-types/src/block.rs
musitdev: All this function is as an example to explain the idea of batch aggregation.

protocol-units/da/movement/protocol/light-node/src/sequencer.rs
mzabaluev: I understand the fix, this is not part of it though. What I'm asking is, if we'll likely need this currently unused method in the future, or is it another piece of dead code that will incur cognitive costs on anyone who needs to work with the code base.

mzabaluev (approved):

protocol-units/da/movement/protocol/light-node/src/sequencer.rs
l-monninger, line 185: @sausagee This is a collapse implementation that can be used before submitting to Celestia.

@musitdev has also mentioned doing this by aggregated blocks after Celestia. But, this should be the same thing because we only have one proposer at the moment. Further, this implementation is easier.

l-monninger: Parallel submission didn't work in this PR because of the way Celestia light nodes proxy requests. There's no internal buffer so sequence numbers can be violated.

PR #1112, opened 2025-03-14
fix(da-light-node): graceful return on unimlemented

mzabaluev:
Replace unimplemented panics on gRPC endpoints
with gRPC status returns.


0xmovses (approved):

mzabaluev:
Unit tests are still broken in main as well.

PR #1116, opened 2025-03-16
feature: Update Load and Soak test to pre execute code before test so that test can be run with one local faucet.

musitdev:
# Summary

Add a prepare function to test scenario that is executed synchronously before to run a scenario.

Allow to run more test with one faucet. The funding part is done in the prepare function serially.

Add a load soak test that reproduce the Cellestia issue when several blob are send to it.

# Changelog
* add prepare function to scenario,
* update existing test
* add soak Alice and Bob test that fail with current version.

# Testing

To run the test that fails:
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.eth-local.celestia-local.soak --keep-tui"
```

# Outstanding issues

### DIFF
--- a/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
+++ b/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
@@ -133,7 +133,7 @@ impl Scenario for BasicScenario {
 		let bob = self.bob.as_mut().unwrap();
 
 		for _ in 0..2 {
-			// Have Bod send Alice some coins.
+			// Have Bob send Alice some coins.
 			let txn_hash = coin_client
 				.transfer(bob, alice.address(), 10, None)
 				.await
--- a/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
+++ b/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
@@ -91,7 +91,7 @@ impl Scenario for BasicScenario {
 	async fn prepare(&mut self) -> Result<(), anyhow::Error> {
 		let rest_client = Client::new(NODE_URL.clone());
 		let faucet_client = FaucetClient::new(FAUCET_URL.clone(), NODE_URL.clone());
-		let coin_client = CoinClient::new(&rest_client); // Print initial balances.
+		let coin_client = CoinClient::new(&rest_client);
 
 		// Create two accounts locally, Alice and Bob.
 		let mut alice = LocalAccount::generate(&mut rand::rngs::OsRng);
--- a/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
+++ b/networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
@@ -89,10 +89,12 @@ static FAUCET_URL: Lazy<Url> = Lazy::new(|| {
 #[async_trait::async_trait]
 impl Scenario for BasicScenario {
 	async fn prepare(&mut self) -> Result<(), anyhow::Error> {
+		let rest_client = Client::new(NODE_URL.clone());
 		let faucet_client = FaucetClient::new(FAUCET_URL.clone(), NODE_URL.clone());
+		let coin_client = CoinClient::new(&rest_client); // Print initial balances.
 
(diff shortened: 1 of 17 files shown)


networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
```diff
@@ -86,107 +88,71 @@ static FAUCET_URL: Lazy<Url> = Lazy::new(|| {
 
 #[async_trait::async_trait]
 impl Scenario for BasicScenario {
-	async fn run(self: Box<Self>) -> Result<()> {
+	async fn prepare(&mut self) -> Result<(), anyhow::Error> {
 		let rest_client = Client::new(NODE_URL.clone());
 		let faucet_client = FaucetClient::new(FAUCET_URL.clone(), NODE_URL.clone());
-		let coin_client = CoinClient::new(&rest_client);
+		let coin_client = CoinClient::new(&rest_client); // Print initial balances.
```
mzabaluev, line 94: Does this comment relate to the line? It seems spuriously added.

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
```diff
@@ -86,107 +88,71 @@ static FAUCET_URL: Lazy<Url> = Lazy::new(|| {
...
+		let alice = self.alice.as_mut().unwrap();
+		let bob = self.bob.as_mut().unwrap();
 
-		// Print intermediate balances.
-		tracing::info!(
-			"Scenario:{}\n=== Intermediate Balances ===\n Alice: {:?}\n Bob: {:?}",
-			self.id,
-			coin_client
-				.get_account_balance(&alice.address())
+		for _ in 0..2 {
```
mzabaluev, line 135: Why is this done two times in one test run?

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
```diff
@@ -86,107 +88,71 @@ static FAUCET_URL: Lazy<Url> = Lazy::new(|| {
...
+			// Have Bod send Alice some coins.
```
mzabaluev, line 136:
```suggestion
			// Have Bob send Alice some coins.
```

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
musitdev: Yes it's an old comment not related to the line. Should be removed.

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
musitdev: It's to add more Tx in the scenario. Depending on the load we want to create, we can increase or decrease the number of loop. 2 is enough to get the issues we've in bardock. When they will be solved, we can try with more Tx.

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
musitdev, line 94:
```suggestion
		let coin_client = CoinClient::new(&rest_client);
```

musitdev: Update done with suggestion.

networks/movement/movement-client/src/bin/e2e/load_alice_bob.rs
musitdev: In this scenario, to scale the load we use Tx. Adding more scenario add more Tx, but each scenario needs to create and fund 2 accounts. So it takes time and can cause some issue with the faucet. By increasing the number of Tx in a scenario, we allow scaling without increasing the account creation.
I keep it because depending on the test we want to do we can need to tweak a little these parameters.
2 is a good compromise between not adding too much TX and using the number of scenario created as variable parameters and keeping the loop when someone need it.

mzabaluev (approved):

PR #1118, opened 2025-03-17
Fix: change batch creation loop timing paramter.

musitdev:
# Summary

- add a specific config parameter for batch production loop
- decrease the Da light node block creation timing loop.
- - remove 0.3.3 Celestia //  task patch.

Set the rapport between of full node batch creation and Da light node block creation to 4 by default.
# Changelog

To deploy this version on Testnet or mainnet the new config param will need to be added. This section should be update as follow, the batch_production_time should be added to the section load_shedding like this example:
```
    "load_shedding": {
      "max_transactions_in_flight": null,
      "batch_production_time": 2000
    },
```
And the `memseq_build_time` should be set to 500 and not 1000.
# Testing

```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.eth-local.celestia-local.load --keep-tui"
```
# Outstanding issues

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11992,6 +11992,7 @@ dependencies = [
  "hyper 1.5.0",
  "k256",
  "maptos-dof-execution",
+ "maptos-execution-util",
  "maptos-opt-executor",
  "mcr-settlement-client",
  "mcr-settlement-config",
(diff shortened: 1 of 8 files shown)


0xmovses:
Nothing is required from Infra team right? As it will just use the `env_default`

0xmovses (approved):

Issue #1137, opened 2025-03-20
Remove old DA Code that is not anymore used


0xmovses:
## Summary

Remove old DA code that isn't used anymore.

Issue #1136, opened 2025-03-20
da-sequencer: Save batch and define storage access 


0xmovses:
## Summary 

Provide storage for pending tx aggregated in batch and produced blocks.

The data are persisted in RocksDB collections.

## Requirements
Data stored:
 * pending Tx: like current implementation Tx are saved until they are aggregated in a saved block. After they are removed from the DB.
 * block: store all block produced from genesis.
 * some utility data like current block height or current da height.

Save batch algo:
 * save batch with its Tx in the pending Tx Db
 * each tx is associated with the batch timestamp and its order in the batch to be used to order the Tx later.

Produce bloc algo
 * get all pending Tx/batch from the DB
 * order the Tx by batch timestamp / tx order in batch
 * Add Tx to an aggregated Vec in order.
 *  For each added Tx calculate the Tx size
 *  Add the calculated size to the Tx vec size
 *  Add Tx until the size reach the max block size or there's no new Tx
 * Generate the new block height. Increment by one from last height.
 * Get parent block Id (digest)
 * Create the new block with the parent id, Tx aggregated vec, new height
 * Save the block
 * Remove from pending Tx the Tx that has been added to the block.
 * return the block + height.

The main loop guaranty that the produce_block function is only called once at a time. ie: No 2 blocks will be produced in //
Some new batch can be added during the produce_block process. This process doesn't interfere with the bock production. If not added, they will be added at the next call.

DB data access:
 * save pending Tx
 * get pending Tx
 * remove pending Tx 
 * save block and height
 * save current height
 * save current Celestia height
 * save block sent on celestia (optional)
 * save block committed in Celestia (optional)
 * get current block height
 * get current block by height / by block Digest
 * get current Celestia height

## Implementation details

RocksDb is a sync API. All function of the module will be sync. The async -> sync link will be done in the main loop.

The pending Tx db code is partially done [here](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/mempool/move-rocks/src/lib.rs) and use [here](https://github.com/movementlabsxyz/movement/blob/3a9e255e1bb3ccad59be0c15e606d23879c580d4/protocol-units/sequencing/memseq/sequencer/src/lib.rs#L91).

[...]

Issue #1135, opened 2025-03-20
Da Node: Submit Batch, verify and store


0xmovses:
## Summary
Contains:

 * Grpc update
 * Batch and Tx verification
 * Send batch to the Tx storage.

## Implementation Details 
Use existing gRPC `submit_batch` call. Remove all unnecessary calls.
Needed calls:

- rpc StreamReadFromHeight (StreamReadFromHeightRequest) returns (stream StreamReadFromHeightResponse); return one blob by height.
- rpc ReadAtHeight (ReadAtHeightRequest) returns (ReadAtHeightResponse); Change return one or none.
* rpc BatchWrite (BatchWriteRequest) returns (BatchWriteResponse); a batch is a vec of Tx.

_to be validated, depending on the openess of the netwrok we want_
As the grpc entry point is public, batch are signed by the fullnode and verified by the Da node.Use Ed25519 keys. 

Fullnode public keys are stored in a separate file in a list of 0x. The file is read every minute so that they can be updated without a restart of the node. The file path is defined in the .movement/config.json under celestia_da_light_node_config/da_light_node section.

Verify and validate each Tx as today.

Confirm if the current whitelist validation is still required.

Save the batch tx in memseq to persist it with a RocksDB. Each Tx are saved individually like today.
Each Tx saved in the storage are called pending tx because they will be removed when added to a block.

## Testing 
The gRPC rust crate auto gens unit tests I believe. Maybe we want a light e2e that tests the connection between services. Developer can leave comments to discuss.
Not sure if we need to test the grpc part because it will be tested in e2e test.
Unit test:
 * Test signed, badly signed batch, non-present public key in list. To verify if the Tx are added correctly to the tx storage, use a mock.

Issue #1134, opened 2025-03-20
[DA-node] Build the whole DA node execution loop.


0xmovses:
## Summary 

Develop the main loop that run all the logic. The pattern use to manage all concurrent task is to use one loop that manage all task start and result and the main logic:
-  process incoming request
-  link module like storage and Celestia
-  manage errors and react

The other role is to control the resource access to avoid too much sync contention.

## Requirements

The main managed process are:
 * Incoming grpc request like submit batch, get batch
 * Celestia client notification
 * Batch processing result and storage calls.
 * Error from submodule and logic to process these errors.

## Testing
This loop will be tested using the e2e integration test using the DA Sequencer gRPC client.
Error case can be simulated using mock.
Load and soak test should be applied too to see how the whole multi task processing works.

PR #1138, opened 2025-03-20
feat: da-sequencer

musitdev:
## Summary

See parent issue #1117 And its sub-issues which must all be merged into this feat branch:
Ticked issues have been merged into this feature branch. 

- [x] https://github.com/movementlabsxyz/movement/issues/1129
- [x] https://github.com/movementlabsxyz/movement/issues/1157
- [x] https://github.com/movementlabsxyz/movement/issues/1131
- [x] https://github.com/movementlabsxyz/movement/issues/1132
- [ ] https://github.com/movementlabsxyz/movement/issues/1133
- [x] https://github.com/movementlabsxyz/movement/issues/1134
- [x] https://github.com/movementlabsxyz/movement/issues/1135
- [x] https://github.com/movementlabsxyz/movement/issues/1136
- [ ] https://github.com/movementlabsxyz/movement/issues/1137
- [x] https://github.com/movementlabsxyz/movement/issues/1164
- [ ] https://github.com/movementlabsxyz/movement/issues/1158
- [x] https://github.com/movementlabsxyz/movement/issues/1165
- [x] https://github.com/movementlabsxyz/movement/issues/1186

## Testing
To run the server:
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup --keep-tui"
```
Load test:
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.load --keep-tui"
```

Soak Test:
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.soak --keep-tui"
```


protocol-units/da-sequencer/node/src/error.rs
```diff
@@ -0,0 +1,6 @@
+/// Errors thrown by Da Sequencer.
+#[derive(Debug, thiserror::Error)]
+pub enum DaSequencerError {
```
andygolay, line 3:
```suggestion
pub enum DASequencerError {
```

protocol-units/da-sequencer/node/src/error.rs
```diff
@@ -0,0 +1,6 @@
+/// Errors thrown by Da Sequencer.
```
andygolay, line 1:
```suggestion
/// Errors thrown by DA Sequencer.
```

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -0,0 +1,24 @@
+use crate::error::DaSequencerError;
```
andygolay, line 1:
```suggestion
use crate::error::DASequencerError;
```

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -0,0 +1,24 @@
...
+pub struct DaBatch<Data> {
+	data: Data,
+	signature: Ed25519Signature,
+	signer: Ed25519PublicKey,
+}
+
+/// Batch write blobs.
+fn validate_batch(
+	new_batch: DaBatch<RawData>,
+) -> std::result::Result<DaBatch<FullnodeTx>, DaSequencerError> {
```
andygolay, line 22:
```suggestion
) -> std::result::Result<DABatch<FullnodeTx>, DASequencerError> {
```

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -0,0 +1,24 @@
...
+#[derive(Debug)]
+pub struct DaBatch<Data> {
+	data: Data,
+	signature: Ed25519Signature,
+	signer: Ed25519PublicKey,
+}
+
+/// Batch write blobs.
+fn validate_batch(
+	new_batch: DaBatch<RawData>,
```
andygolay, line 21:
```suggestion
	new_batch: DABatch<RawData>,
```

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+
+mod batch;
+mod block;
+mod celestia;
+mod error;
+mod storage;
+
+/// Run Da sequencing loop.
+/// This function only return in case of error that indicate a crash of the node.
+pub fn run(config: DaSequencerConfig) -> std::result::Result<(), DaSequencerError> {
```
andygolay, line 12:
```suggestion
pub fn run(config: DASequencerConfig) -> std::result::Result<(), DaSequencerError> {
```

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+mod block;
+mod celestia;
+pub mod error;
+pub mod server;
+mod storage;
+#[cfg(test)]
+mod tests;
+
+/// Run Da sequencing loop.
+/// This function only return in case of error that indicate a crash of the node.
```
0xmovses, line 22:
```suggestion
/// Runs the DA sequencing loop.
///
/// This function only returns if a critical error occurs, indicating a node crash or unrecoverable failure.
```

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+			//propagate the new block.
+			res = &mut produce_block_jh => {
+				produce_block_jh = get_pending_future();
+				produce_block = false;
+				match res {
+					Ok(Ok(Some(block))) => {
+						let block_digest = block.get_block_digest();
+						// Send the block to all registered follower
+						// For now send to the main loop because there are very few followers (<100).
+						tracing::info!(sender_len = %connected_grpc_sender.len(), block_height= %block.height.0, "New block produced, send to fullnodes.");
```
0xmovses, line 118:
```suggestion
						tracing::info!(sender_len = %connected_grpc_sender.len(), block_height= %block.height.0, "New block produced, sent to fullnodes.");
```

protocol-units/da-sequencer/node/src/main.rs
```diff
@@ -0,0 +1,41 @@
...
+
+	let (request_tx, request_rx) = mpsc::channel(100);
+	// Start gprc server
+	let grpc_address = da_sequencer_config.grpc_listen_address;
+	let grpc_jh = tokio::spawn(async move {
+		movement_da_sequencer_node::server::run_server(grpc_address, request_tx).await
+	});
+
+	//Start the main loop
+	todo!();
```
0xmovses, line 38: This loop can be called and tested with a straightforward e2e test after the all the todo!() are complete in the Celestia client.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,233 @@
...
+use tokio_stream::Stream;
+use tonic::transport::Server;
+
+/// Runs the server
+pub async fn run_server(
+	address: SocketAddr,
+	request_tx: mpsc::Sender<GrpcRequests>,
+) -> Result<(), anyhow::Error> {
+	tracing::info!("Server listening on: {}", address);
+	let reflection = tonic_reflection::server::Builder::configure()
```
0xmovses, line 26:
```suggestion
// Enables gRPC introspection, a feature that allows gRPC clients 
// (like grpcurl, Postman, or IDEs) to query the server for:
// - Available services and RPC methods
// - Input/output message types
// - Field types, enums, and nested structures 
// - Metadata like documentation or options (if included)
let reflection = tonic_reflection::server::Builder::configure()
```

protocol-units/da-sequencer/node/src/tests/mock.rs
```diff
@@ -0,0 +1,189 @@
...
+use movement_da_sequencer_proto::blob_response::BlobType;
+use movement_da_sequencer_proto::BatchWriteRequest;
+use movement_types::block::Block;
+use movement_types::block::BlockMetadata;
+use movement_types::block::Id;
+use movement_types::transaction::Transaction;
+use std::collections::BTreeSet;
+use std::future::Future;
+use std::sync::Arc;
+use std::sync::Mutex;
```
0xmovses, line 24:
```suggestion
use std::{
	collections::BTreeSet,
	future::Future,
	sync::{Arc, Mutex},
};

use futures::StreamExt;
use ed25519_dalek::{SigningKey, VerifyingKey};

use movement_da_sequencer_client::{sign_batch, DaSequencerClient};
use movement_da_sequencer_proto::{blob_response::BlobType, BatchWriteRequest};
use movement_types::{
	block::{Block, BlockMetadata, Id},
	transaction::Transaction,
};

use crate::{
	batch::{DaBatch, FullNodeTxs, serialize_full_node_batch},
	block::{BlockHeight, SequencerBlockDigest},
	celestia::{blob::Blob, CelestiaHeight},
	DaSequencerError, DaSequencerExternalDa, DaSequencerStorage, SequencerBlock,
};
```

protocol-units/da-sequencer/setup/src/lib.rs
```diff
@@ -0,0 +1,14 @@
...
+#[cfg(test)]
+mod tests {
+	use super::*;
+
+	#[test]
+	fn it_works() {
+		let result = add(2, 2);
+		assert_eq!(result, 4);
+	}
+}
```
0xmovses, line 14: This file seems redundant.

0xmovses (changes requested):

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+	let mut connected_grpc_sender = vec![];
+
+	loop {
+		tokio::select! {
+			// Manage grpc request.
+			Some(grpc_request) = request_rx.recv() => {
+				match grpc_request {
+					GrpcRequests::StartBlockStream(produced_tx, curent_height_callback) => {
+						connected_grpc_sender.push(produced_tx);
+
```
l-monninger, line 52: I think given your implementation in `StreamReadFromHeight` your basic approach is to start the stream immediately and then back fetch heights up to first height in the stream. 

Can you write this down? Where are back fetches being buffered to ensure they are emitted in order?

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+					Ok(Ok(Some(block))) => {
+						let block_id = block.id();
+						// Send the block to all registered follower
+						// For now send to the main loop because there are very few followers (<100).
+						tracing::info!(sender_len = %connected_grpc_sender.len(), block_height= %block.height().0, "New block produced, send to fullnodes.");
+						stream_block_to_sender(&mut connected_grpc_sender, Some(block)).await;
+
+						//send the block to Celestia.
+						let celestia_send_jh = tokio::spawn({
+							let celestia = celestia.clone();
```
l-monninger, line 124: Where is the WAL which ensures we would eventually write to Celestia if this process is live?

l-monninger: You should not return blocks back over the stream until you've written to this WAL. 

Can be the DB itself, but it's not clear where the continued attempts to write and achieve confirmation from Celestia before popping off of the WAL are.

l-monninger (changes requested):

proto/movementlabs/protocol_units/da/sequencer/v1.proto
```diff
@@ -0,0 +1,54 @@
+syntax = "proto3";
```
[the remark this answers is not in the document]
l-monninger: Is the intent to make this distinct because the underlying semantics are distinct?

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: Remember, the basic idea behind similar async writes to Celestia that other platforms do is eventual consistency. Without this kind of WAL construct, it's hard to communicate how you would have that.

networks/movement/movement-client/src/bin/e2e/soak_alice_bob.rs
```diff
@@ -1,30 +1,34 @@
...
 fn main() {
 	// Define the Test config. Use the default parameters.
 	let mut config = ExecutionConfig::default();
 
 	// Set some params for the load test, drive some requests
 	config.kind = TestKind::Soak {
-		min_scenarios: 60,
-		max_scenarios: 60,
-		duration: std::time::Duration::from_secs(6000), // 10 minutes
+		min_scenarios: 200,
```
l-monninger, line 24: Does hardcoding make sense? Since this is a bin, might we be better off with some clap args?

l-monninger (changes requested):

demo/hsm/src/cli/server/ed25519/hashi_corp_vault.rs
```diff
@@ -14,27 +14,27 @@ use tokio::sync::Mutex;
 #[derive(Debug, Parser, Clone)]
 #[clap(rename_all = "kebab-case", about = "Runs signing app for ed25519 against HashiCorp Vault")]
 pub struct HashiCorpVault {
-        canonical_key: String,
-        #[arg(long)]
-        create_key: bool,
+	canonical_key: String,
```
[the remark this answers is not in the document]
musitdev: At one moment, yes. Should have been corrected. I'll reedit them.

networks/movement/movement-client/src/bin/e2e/soak_alice_bob.rs
musitdev: I was build at each run, so it was easier to change the value in the code than adding an env var and modifying the command.

proto/movementlabs/protocol_units/da/sequencer/v1.proto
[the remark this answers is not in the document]
musitdev: The underlying process is different and it was not sure the other service fit with it. For example, it was not done but the batch data (bach, signature, pubkey) should be exposed and not hidden behind a byte array.

proto/movementlabs/protocol_units/da/sequencer/v1.proto
```diff
@@ -0,0 +1,54 @@
...
+package movementlabs.protocol_units.da.sequencer.v1;
+
+
+service DaSequencerNodeService {
+  rpc StreamReadFromHeight (StreamReadFromHeightRequest) returns (stream StreamReadFromHeightResponse);
```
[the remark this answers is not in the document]
musitdev: It's implemented in the da-sequencer server.rs and the even loop provide the produced block for all connected client.

protocol-units/da-sequencer/client/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+/// Verifies a batch signature using the given verifying key.
+pub fn verify_batch_signature(
+	batch_data: &[u8],
+	signature: &ed25519_dalek::Signature,
+	verifying_key: &VerifyingKey,
+) -> Result<(), anyhow::Error> {
+	Ok(verifying_key.verify(batch_data, signature)?)
+}
+
+/// A DaSequencerClient implementation that no nothing. Can be used to mock the DA.
```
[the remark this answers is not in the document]
musitdev: Yes, I agree, I try to move it in the test::mock module.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -0,0 +1,91 @@
+use movement_types::block::Block;
+use serde::{Deserialize, Serialize};
+
+use crate::error::DaSequencerError;
+
+// TODO: use a sensible value for the max sequencer block size
+pub const MAX_SEQUENCER_BLOCK_SIZE: u64 = 1_000_000; // 1 MB
+
```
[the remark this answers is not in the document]
musitdev: I put 100Mb, it was 1gb before but I thought creating 1gb block can generate dissemination latency issue.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -0,0 +1,82 @@
...
+impl TryFrom<&[u8]> for SequencerBlock {
+	type Error = DaSequencerError;
+
+	fn try_from(bytes: &[u8]) -> Result<Self, Self::Error> {
+		bcs::from_bytes(bytes).map_err(|e| DaSequencerError::Deserialization(e.to_string()))
+	}
+}
+
+impl TryInto<Vec<u8>> for SequencerBlock {
```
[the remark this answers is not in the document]
musitdev: It contains block data used by the da-sequencer. It uses the utils::Block and add specific data needed by the da-equencer.

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+	));
+	let mut spawn_result_futures = FuturesUnordered::new();
+	let mut produce_block = false;
+	let mut produce_block_jh = get_pending_future();
+	let mut connected_grpc_sender = vec![];
+
+	loop {
+		tokio::select! {
+			// Manage grpc request.
+			Some(grpc_request) = request_rx.recv() => {
```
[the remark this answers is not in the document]
musitdev: It’s a kind of supervision pattern: the main loop coordinates all cross‐domain interactions. Since Celestia isn’t deployed yet, this pattern isn’t critical—but its purpose is to handle errors and inform other components so they can react appropriately. For example, to cope with Celestia back-pressure, the Celestia module could send a message to the loop, which would then slow down block production. We haven’t implemented that yet—because the best approach is still unclear—but this pattern makes that cross-domain logic much easier to manage.

protocol-units/da-sequencer/node/src/lib.rs
musitdev: No — if the requested height is behind, the older block is fetched first and then the newly created one. The logic is in the server.rs that implements the grpc service. It preserves the correct ordering, assuming that blocks sent earlier arrive before later ones.
This request start the steam of new block, but they are buffered in the service entry point until old block are sent.

protocol-units/da-sequencer/node/src/lib.rs
musitdev: The WAL role is done by the DB. We don't wait the Celestia confirmation. Celestia is only used to store the block ordering. The ordering is forced because we send only one blob at a time to celestia.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+	/// Stream blocks from a specified height or from the latest height.
+	async fn stream_read_from_height(
+		&self,
+		request: tonic::Request<StreamReadFromHeightRequest>,
+	) -> Result<tonic::Response<Self::StreamReadFromHeightStream>, tonic::Status> {
+		tracing::info!(request = ?request, "Stream read from height request");
+
+		// Register the new produced block channel to get lastest block.
+		// Use `unbounded_channel` to avoid filling the channel during the fetching of an old block.
+		let (produced_tx, mut produced_rx) = mpsc::unbounded_channel();
```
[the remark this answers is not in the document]
musitdev: Similar da-stream test has been implemented in the crate test module.

0xmovses:
Conflicts needed to be resolved before the re-review 🙏

networks/movement/movement-client/src/bin/e2e/soak_alice_bob.rs
l-monninger: But, now we need to think a bit about this being run on multiple devices, including CI to test regression no? Is there a case for merging this as it is?

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: Okay, but how? I don't see a background task that's continually trying to read off from some point in the DB. It just looks like values are added to the DB and then sometimes also to Celestia if everything goes right.

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: Yes, but there's a fundamental problem here when you're switching from fetching existing blocks to live streaming: how do you switch over to the live stream s.t. you don't miss any blocks?

Take this example:

```
BACKFETCH height 1
CHECK_LATEST_HEIGHT # more to back fetch

BACKFETCH height 2
CHECK_LATEST_HEIGHT # looks like we're caught up when we make this call, so start streaming

BROADCAST height 3
# OH NO! Another block arrived while we were starting to listen to the stream
# We miss the broadcast for height 3

START_STREAM
BROADCAST height 4
RECV_STREAM # receive 4
BROADCAST height 5
RECV_STREAM # receive 5
```

protocol-units/da-sequencer/node/src/server.rs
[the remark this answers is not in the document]
l-monninger: Are you talking about in this dir?: https://github.com/movementlabsxyz/movement/pull/1138/files/7ee072081d35723b33bd7e9b2fb5ce66d13c32ac#diff-4b3d5e38255afa453461daa7290617af3f01aa70b441d1efa4cd1f14873f57f3

networks/movement/movement-client/src/bin/e2e/soak_alice_bob.rs
musitdev: I was like that before. Load and Soak test aren't automated in the CI. When we'll decide to automate it, we'll see the best choice.

protocol-units/da-sequencer/node/src/lib.rs
musitdev: I think I didn't understand the role of the WAL. For me, it was to store the blob / block until they are written to Celestia then removed. As we don't write blob, but only the block digest, the current db is used as a block DA. So block are never removed.

protocol-units/da-sequencer/node/src/lib.rs
musitdev: It's done by the channel. When the request is sent to the supervisor loop, the next first produced block height send in the channel is return. During the old block fetch, new produced block are sent in the channel and wait until the fetch loop reach the first produced block height, then it starts to pull the channel. This way it can't miss a block because the switch height is defined at the beginning.

protocol-units/da-sequencer/node/src/server.rs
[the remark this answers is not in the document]
musitdev: The client.rs one: https://github.com/movementlabsxyz/movement/blob/feature/da-sequencer-node/protocol-units/da-sequencer/node/src/tests/client.rs
It implements da client block stream scenarios.

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: I mean a WAL is a WAL; you write what you're going to later confirm. If the goal isn't to have consistency with Celestia writes anymore, just lmk and I'll mark this as resolved.

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: Let me review the implementation to make sure this holds.

protocol-units/da-sequencer/node/src/lib.rs
```diff
@@ -0,0 +1,177 @@
...
+				match grpc_request {
+					GrpcRequests::StartBlockStream(produced_tx, curent_height_callback) => {
```
l-monninger, line 134: I think this is supposed to be `proposed_block`.

l-monninger (changes requested):

protocol-units/da-sequencer/node/src/server.rs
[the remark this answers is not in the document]
l-monninger: But, it's an integration test. It's not reasoning about the fundamental types; it's testing behavior for a specific implementation. This makes it less potent in checking the correctness of the algorithm.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+		.add_service(reflection)
+		.serve(address)
+		.await?;
+
+	Ok(())
+}
+
+#[derive(Debug)]
+pub enum GrpcRequests {
+	StartBlockStream(mpsc::UnboundedSender<Option<SequencerBlock>>, oneshot::Sender<BlockHeight>),
```
l-monninger: Should comment that this is going to send back blocks and heights and why one is `oneshot` and the other is an unbounded channel. 

Additionally, why should `SequencerBlock` be an `Option`?

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+
+		let mut current_produced_height = match current_height_rx.await {
+			Ok(h) => h,
+			Err(err) => {
+				tracing::warn!("start stream channel closed: {err}");
+				return Err(tonic::Status::internal("Internal error. Retry later"));
+			}
+		};
+		let mut current_block_height = request.into_inner().height;
+		//The genesis block can't be retrieved.
```
l-monninger, line 110: Why?

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+		if let Err(err) = self
+			.request_tx
+			.send(GrpcRequests::StartBlockStream(produced_tx, current_height_tx))
+			.await
+		{
+			tracing::warn!(error = %err, "Internal grpc request channel closed, can't stream blocks");
+			return Err(tonic::Status::internal("Internal error. Retry later"));
+		}
+
+		let mut current_produced_height = match current_height_rx.await {
```
l-monninger, line 102: Leave a comment here that indicates this is where we start reading from the stream s.t. we have guarantees about availability.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+					let block = match get_height_rx.await {
+						Ok(b) => b,
+						Err(err) => {
+							tracing::warn!(error = %err, "Stream block: oneshot channel closed");
+							return;
+						}
+					};
+					current_block_height +=1;
+
+					let blockv1 = match block.as_ref().map(|block| block.try_into()) {
```
l-monninger: Should be `block_v1`.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -0,0 +1,254 @@
...
+							tracing::warn!(error = %err, "Streamed block serialization failed.");
+							return;
+
+						}
+					};
+					BlockResponse { block_type: Some(BlockType::BlockV1(blockv1)) }
+
+				} else {
+					//send block in produced channel
+					let received_content = match produced_rx.recv().await {
```
l-monninger, line 149: See, I think you should actually call this instead of `current_height_rx`. It's more directly what the stream you later join with actually is and thus less susceptible to regression by an unwitting contributor.

protocol-units/da-sequencer/node/src/lib.rs
l-monninger: Okay, so think this holds under: https://github.com/movementlabsxyz/movement/pull/1138/files/7ee072081d35723b33bd7e9b2fb5ce66d13c32ac#r2088752882

But, I think this is better: https://github.com/movementlabsxyz/movement/pull/1138/files/7ee072081d35723b33bd7e9b2fb5ce66d13c32ac#r2088757548

0xmovses:
@musitdev also all of my review comments haven't been addressed (some our outdated but many are still relevant), for eg also commented on this 
https://github.com/movementlabsxyz/movement/pull/1138#discussion_r2088753813

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -0,0 +1,183 @@
...
+	signer: VerifyingKey,
+	signature: Signature,
+	mut data: Vec<u8>,
+) -> Vec<u8> {
+	let mut serialized: Vec<u8> = Vec::with_capacity(64 + 32 + data.len());
+	serialized.extend_from_slice(&signer.to_bytes());
+	serialized.extend_from_slice(&signature.to_bytes());
+	serialized.append(&mut data);
+	serialized
+}
```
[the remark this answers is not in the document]
musitdev: We need the signature. @l-monninger explain it in another comment.

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -0,0 +1,183 @@
...
+	let data: Vec<u8> = vec_deserialized.to_vec();
+	Ok((public_key, signature, data))
+}
+
+pub fn validate_batch(
+	new_batch: DaBatch<RawData>,
+) -> Result<DaBatch<FullNodeTxs>, DaSequencerError> {
+	verify_batch_signature(&new_batch.data.data, &new_batch.signature, &new_batch.signer)?;
+
+	let txs: FullNodeTxs = bcs::from_bytes(&new_batch.data.data)
```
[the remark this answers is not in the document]
musitdev: done.

protocol-units/da-sequencer/node/src/server.rs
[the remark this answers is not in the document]
musitdev: The algorithm is spread around several crates: Storage to extract block and produce block, the supervisor loop to send produced block and manage start height and the server stream to return the blocks. I found that it was easier and faster to test the correctness by using the client as an integration test.
Each sub part can be unit test (it takes more time) but it doesn't guaranty that the whole is correct. So we've done the integration first.

protocol-units/da-sequencer/node/src/server.rs
musitdev: In the DB block height start as 1 and the genesis block is not present. I thought it was not accessible. If there's a way to get it for all network, we can add it to the db and remove this limitation.

protocol-units/da-sequencer/node/src/server.rs
musitdev: done

protocol-units/da-sequencer/node/src/server.rs
musitdev: I've changed the naming.
I put `first_produced_block_height_rx` for `current_height_rx`
and `current_produced_block_rx` for `produced_rx`

protocol-units/da-sequencer/node/src/lib.rs
musitdev: Yes, we don't try to guaranty the consistency with Celestia mostly because if an inconsistency is detected there's no simple way to solve it. Block are executed before their writing and confirmation in Celestia. The Db is the truth, and we try our best to guaranty that Celestia block order is correct.

protocol-units/da-sequencer/node/src/lib.rs
musitdev: They are addressed in the PR #1231

protocol-units/da-sequencer/node/src/server.rs
l-monninger: Comment it in the code is what I’m saying.

protocol-units/da-sequencer/node/src/server.rs
l-monninger: I think that’s approach fine. Just leave a comment explaining why.

protocol-units/da-sequencer/node/src/server.rs
l-monninger: I didn’t mean to change the naming. I meant to use the `current_produced_block_rx` to determine the height up to which you back fetch. 

This eliminates the need for one shot, creates a single source of truth, and is easier to reason about.

musitdev: Ok, the names are clearer now, so I’ll keep the change.
The supervisor loop is our only source of truth—channels are just a communication mechanism. Both channels rely on the same source of truth.

We *could* collapse them into a single channel, but that would add complexity. They serve distinct purposes:

- A **oneshot** channel returns immediately so the fetch loop can start at the correct height.  
- An **unbounded** channel delivers blocks as they become available.

Our requirements are:

1. Determine the height of the first produced block so we know up to which block to fetch from the database.  
2. Receive newly produced blocks once we’ve caught up to that initial height.

Because these needs are different, I prefer to keep two separate channels to satisfy them.

l-monninger: I would argue two channels are two sources of truth. Mainly, this because a contributor could come along not realizing that these are suppose to report on objects in the same stream and misuse one or the other. 

I don't think there's too much complexity if you were to change over. But, I'll take on adding a suggestive PR to show you what I mean.

l-monninger: This is what I'm referring to: https://github.com/movementlabsxyz/movement/pull/1239

musitdev:
I've updated the PR description. You need to remove all Celestia and eth entries. Remove the .movement repo before too.

0xmovses (approved):
This is currently in production, additional improvements, optimisations and features should be made to main post merge. This is so that we can start testing metrics on the latest node with the da-sequencer, this week. 

The current open PR that depend on this one, can be redirected to main, because this will be merged as a normal merge commit (not a squash commit), thus retaining the commit by commit history on main.
Author @l-monninger https://github.com/movementlabsxyz/movement/pull/1253

0xmovses (approved):
Ran all the test commands. All looking good from here.

andygolay (approved):
I ran the soak test, LGTM.

0xmovses:
Note, _do not_ delete this branch.

PR #1150, opened 2025-03-24
feat: da-seq persist rocksdb store

0xmovses:
## Summary
Read the full issue #1136 

_NB._ I left out the Celestia related queries, will put that in a seperate PR. It makes this one easier to review and the diff smaller. 

## Changelog
- Implements methods relating to RocksDB storage only
- Adds unit tests for all the methods with some unhappy paths
- Adds block size unit test
- Adds a limit on tx size within a block as well as `SequencerBlock` limit in bytes

## Testing
` cargo test -p movement-da-sequencer-node` 

```
running 8 tests
test storage::tests::test_try_new_invalid_path_should_fail ... ok
test storage::tests::test_get_block_with_digest_returns_none_for_unknown_digest ... ok
test storage::tests::test_get_block_at_height_returns_correct_block ... ok
test storage::tests::test_write_batch_persists_transaction ... ok
test storage::tests::test_try_new_creates_storage_successfully ... ok
test storage::tests::test_get_block_with_digest_returns_correct_block ... ok
test storage::tests::test_produce_next_block_generates_block_and_clears_pending_tx ... ok
test block::tests::test_sequencer_block_rejects_block_larger_than_max_size ... ok

test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.26s
```

## Outstanding issues
Be sure to check the implementation for `produce_block`


protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +23,71 @@ impl SequencerBlockDigest {
...
+		sb.validate_size()?;
+		Ok(sb)
+	}
+
 	pub fn get_block_digest(&self) -> SequencerBlockDigest {
 		SequencerBlockDigest(*self.block.id().as_bytes())
 	}
+
+	pub fn validate_size(&self) -> Result<(), DaSequencerError> {
+		let size = bincode::serialized_size(self)
```
mzabaluev, line 50: Why bincode here? We use bcs for serialization in other places.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +23,71 @@ impl SequencerBlockDigest {
...
+)]
 pub struct BlockHeight(pub u64);
 
-#[derive(Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct SequencerBlock {
-	height: BlockHeight,
-	block: Block,
+	pub height: BlockHeight,
+	pub block: Block,
```
mzabaluev, line 34: Since `SequencerBlock` has a size restriction enforced by `SequencerBlock::try_new`, should these not be made public and therefore freely modifiable?

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +23,71 @@ impl SequencerBlockDigest {
...
 	pub fn get_block_digest(&self) -> SequencerBlockDigest {
 		SequencerBlockDigest(*self.block.id().as_bytes())
 	}
+
+	pub fn validate_size(&self) -> Result<(), DaSequencerError> {
+		let size = bincode::serialized_size(self)
+			.map_err(|e| DaSequencerError::Generic(format!("Serialization failed: {}", e)))?;
+
+		if size > MAX_SEQUENCER_BLOCK_SIZE {
+			Err(DaSequencerError::Generic(format!(
```
mzabaluev, line 54: Should the variant be specific to this case, e.g. `SizeExceedsMax(usize)`?

protocol-units/da-sequencer/node/src/error.rs
```diff
@@ -3,4 +3,6 @@
 pub enum DaSequencerError {
 	#[error("Error during storage access: {0}")]
 	StorageAccess(String),
+	#[error("Generic error: {0}")]
+	Generic(String),
```
mzabaluev, line 7: I think each variant to be specific to some error case.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
+		let key = tx.id.0;
+		let value = bincode::serialize(&tx)
+			.map_err(|e| DaSequencerError::Generic(format!("Serialization error: {}", e)))?;
+
+		let mut write_batch = WriteBatch::default();
+		write_batch.put_cf(&cf, key, value);
+
+		self.db
+			.write(write_batch)
+			.map_err(|e| DaSequencerError::Generic(format!("DB write error: {}", e)))?;
```
mzabaluev, line 65: You're writing a batch in your user-defined format, but there is no need to use a RocksDB batch to write a single KV pair.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
 	/// Return, if exists, the Block at the given height.
 	pub fn get_block_at_height(
 		&self,
 		height: BlockHeight,
 	) -> std::result::Result<Option<SequencerBlock>, DaSequencerError> {
-		todo!();
+		let cf = self
+			.db
+			.cf_handle(cf::BLOCKS)
+			.ok_or_else(|| DaSequencerError::Generic("Missing column family: blocks".into()))?;
```
mzabaluev, line 78: Could fit under the storage error case.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
+
+		let key: [u8; 8] = height.0.to_be_bytes();
+
+		match self.db.get_cf(&cf, key).map_err(|e| DaSequencerError::Generic(e.to_string()))? {
+			Some(bytes) => {
+				let block: SequencerBlock = bincode::deserialize(&bytes).map_err(|e| {
+					DaSequencerError::Generic(format!("Deserialization error: {}", e))
```
mzabaluev, line 85: Could be a specific variant for deserialization errors.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
+		let key = id.0;
+
+		let height_bytes =
+			match self.db.get_cf(&cf, key).map_err(|e| DaSequencerError::Generic(e.to_string()))? {
+				Some(bytes) => bytes,
+				None => return Ok(None),
+			};
+
+		if height_bytes.len() != 8 {
+			return Err(DaSequencerError::Generic(
```
mzabaluev, line 112: Could be a storage error or more specifically, storage format error (where also deserialization failures would fit, since it's the same problem of the database entry format being seemingly wrong)

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
+				"Invalid height length in digest mapping".into(),
+			));
+		}
+
+		let mut arr = [0u8; 8];
+		arr.copy_from_slice(&height_bytes);
+		let height = BlockHeight(u64::from_be_bytes(arr));
```
mzabaluev, line 119:
```suggestion
		let height = BlockHeight(u64::from_be_bytes(height_bytes.try_into().unwrap()));
```

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,36 +1,124 @@
...
+
+		let key = id.0;
+
+		let height_bytes =
+			match self.db.get_cf(&cf, key).map_err(|e| DaSequencerError::Generic(e.to_string()))? {
+				Some(bytes) => bytes,
+				None => return Ok(None),
+			};
+
+		if height_bytes.len() != 8 {
```
mzabaluev, line 111: This and explicity copying into a sized array below could benefit from an available `TryFrom` impl:

```rust
let height_bytes_sized = height_bytes
    .try_into()
    .map_err(|_| DaSequencerError::StorageFormat("invalid length of a height value in digest mapping".into()))?;
let height = BlockHeight(u64::from_be_bytes(height_bytes_sized));
```

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -43,7 +131,79 @@ impl Storage {
...
+			return Ok(None);
+		}
+
+		let tx_set: BTreeSet<Transaction> = txs.into_iter().collect();
+
+		// Construct and validate the block
+		// Check that Id:default() generates a unique id every time.
+		let block = Block::new(BlockMetadata::default(), Id::default(), tx_set);
+
+		let sequencer_block = SequencerBlock::try_new(next_height, block)?; // validates size internally
```
mzabaluev, line 168: This will fail if we have more data than the size limit. However, the limit is not enforced when pending transactions are written, is it?

protocol-units/da-sequencer/node/src/error.rs
0xmovses: Yeah, am aware. I thought it could be a nice follow up PR. As this PR needs to go in for others. It can be a refactor followup.

protocol-units/da-sequencer/node/src/block.rs
0xmovses: I'll switch to bcs.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -1,4 +1,9 @@
 use movement_types::block::Block;
+use serde::{Deserialize, Serialize};
+
+use crate::error::DaSequencerError;
+
+const MAX_SEQUENCER_BLOCK_SIZE: u64 = 1_000_000; // 1 MB
```
sebtomba, line 6: Where does this number come from? Is it related to the Celestia blob size?

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: I could use the try_into, then handle the error rather than unwrap.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +23,71 @@ impl SequencerBlockDigest {
...
+		let sb = SequencerBlock { height, block };
+		sb.validate_size()?;
+		Ok(sb)
+	}
+
 	pub fn get_block_digest(&self) -> SequencerBlockDigest {
 		SequencerBlockDigest(*self.block.id().as_bytes())
 	}
+
+	pub fn validate_size(&self) -> Result<(), DaSequencerError> {
```
musitdev, line 49: Size validation should be done Tx after Tx. You get the Tx size, and you define if it can be added.

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -1,24 +1,49 @@
...
 }
 
+#[cfg(test)]
+impl<T> DaBatch<T>
+where
+	T: Serialize + CryptoHash,
+{
+	/// Creates a test-only `DaBatch` with a real signature over the given data.
+	/// Only usable in tests.
+	pub fn test_only_new(data: T) -> Self {
```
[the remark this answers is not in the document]
0xmovses: This is a style point and an approach I took from aptos-core, I think it's nice to have methods feature flagged and implemented only for tests.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,74 +1,404 @@
...
+		Ok(Storage { db: Arc::new(db) })
+	}
+
+	pub fn write_batch(&self, batch: DaBatch<FullNodeTx>) -> Result<(), DaSequencerError> {
+		let cf = self.db.cf_handle(cf::PENDING_TRANSACTIONS).ok_or_else(|| {
+			DaSequencerError::Generic("Missing column family: pending_transactions".into())
+		})?;
+		let tx = batch.data();
+		let key = tx.id.0;
+		let value = bcs::to_bytes(tx)
```
musitdev, line 53: We use the Tx batch to build a block. In the worst case, not all tx's batch can fit in the block, so you've to fill the block and update the batch in this case. Perhaps it's easier to save each Tx individually with the batch timestamp, in this case when the Tx are added to a saved block you need to delete the Tx.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -1,74 +1,404 @@
...
-	/// Aggregate all pending Tx until the block is filled.
-	/// A block is filled if no more Tx are pending or it's size is more than the max size.
-	/// All pending Tx added to the block are removed from the pending Tx table.
-	/// Save the block for this height
-	/// Return the block.
-	pub fn produce_next_block(
-		&self,
-	) -> std::result::Result<Option<SequencerBlock>, DaSequencerError> {
-		todo!();
+	pub fn produce_next_block(&self) -> Result<Option<SequencerBlock>, DaSequencerError> {
```
musitdev, line 117: This part should be rethought, we speak about it tomorrow. In the first review, I didn't see this file (it was hidden) so I miss it. I thought the block part hasn't been done.

protocol-units/da-sequencer/node/src/block.rs
sebtomba: We should add a `// TODO: use a sensible value for the max sequencer block size`.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +23,71 @@ impl SequencerBlockDigest {
...
-	block: Block,
+	pub height: BlockHeight,
+	pub block: Block,
 }
 
 impl SequencerBlock {
+	/// Try to construct a SequencerBlock, but fail if it exceeds the max encoded size.
+	pub fn try_new(height: BlockHeight, block: Block) -> Result<Self, DaSequencerError> {
+		let sb = SequencerBlock { height, block };
+		sb.validate_size()?;
```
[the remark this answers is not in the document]
0xmovses: As agreed will deal with block production in a follow up dedicated PR

protocol-units/da-sequencer/node/src/block.rs
0xmovses: Will be dealing with this in the `feat: da-seq produce_block` pr

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: To be addressed in followup `produce_block` PR.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: To be addressed in followup `produce_block` PR.

0xmovses:
As agreed, this PR is now concerned only with establishing the local RocksDB store and its operations. The method `produce_block` has some intricies and will be handled in #1159. This is so we can get these changes into the feature branchas other PRs require id.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: Will handle this logic in #1159

protocol-units/da-sequencer/node/src/block.rs
0xmovses: That was done #1156 and merged in here.

protocol-units/da-sequencer/node/Cargo.toml
```diff
@@ -10,20 +10,29 @@ publish.workspace = true
...
 
 [dependencies]
+aptos-crypto-derive = { workspace = true }
+anyhow = { workspace = true }
+bcs = { workspace = true }
+aptos-crypto = { workspace = true }
+thiserror = { workspace = true }
+url = { workspace = true }
+rocksdb = { workspace = true }
+serde = { workspace = true }
```
mzabaluev, line 20: You moved this entries around, why not sort the whole list while you are at it?

protocol-units/da-sequencer/node/src/block.rs
mzabaluev: Never late to correct a wrong? It's a minor thing though.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -45,69 +61,354 @@ pub trait DaSequencerStorage {
...
+			if key.len() != 8 {
+				return Err(DaSequencerError::Generic("Invalid block height key length".into()));
+			}
+
+			let mut arr = [0u8; 8];
+			arr.copy_from_slice(&key);
+			let last_height = u64::from_be_bytes(arr);
+			Ok(BlockHeight(last_height + 1))
+		} else {
+			Ok(BlockHeight(0))
```
musitdev, line 103: I think first block height is 1 and it's compatible with the other return path.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -18,17 +24,72 @@ impl SequencerBlockDigest {
...
-	height: BlockHeight,
-	block: Block,
+	pub height: BlockHeight,
+	pub block: Block,
 }
 
 impl SequencerBlock {
+	/// Try to construct a SequencerBlock, but fail if it exceeds the max encoded size.
+	pub fn try_new(height: BlockHeight, block: Block) -> Result<Self, DaSequencerError> {
+		todo!()
```
mzabaluev, line 41: Does this need to be completed?

util/movement-types/src/block.rs
```diff
@@ -103,8 +112,12 @@ impl Block {
 		)
 	}
 
-	pub fn add_transaction(&mut self, transaction: Transaction) {
+	pub fn add_transaction(&mut self, transaction: Transaction) -> Result<(), BlockError> {
+		if self.transactions.len() >= MAX_TRANSACTIONS {
+			return Err(BlockError::BlockFull);
+		}
```
mzabaluev, line 118: Fallibility, good.

util/movement-types/src/block.rs
```diff
@@ -103,8 +112,12 @@ impl Block {
 		)
 	}
 
-	pub fn add_transaction(&mut self, transaction: Transaction) {
+	pub fn add_transaction(&mut self, transaction: Transaction) -> Result<(), BlockError> {
+		if self.transactions.len() >= MAX_TRANSACTIONS {
```
musitdev, line 116: There's no max Tx per block but more a max size of the Vec<TX> array. Tx size can be very different, and we don't want that a block is rejected by the Celestia Da because of the size.

util/movement-types/src/transaction.rs
```diff
@@ -1,11 +1,12 @@
+use aptos_crypto_derive::{BCSCryptoHash, CryptoHasher};
 use core::fmt;
 use serde::{Deserialize, Serialize};
 use std::cmp::Ordering;
 
 #[derive(
 	Serialize, Deserialize, Clone, Copy, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord,
 )]
-pub struct Id([u8; 32]);
+pub struct Id(pub [u8; 32]);
```
mzabaluev, line 9: Eh, it was a cleanly encapsulated type, what do you need to expose this for?

util/movement-types/src/transaction.rs
```diff
@@ -40,13 +41,15 @@ impl fmt::Display for Id {
...
+#[derive(
+	Serialize, Deserialize, Clone, Default, Debug, PartialEq, Eq, Hash, CryptoHasher, BCSCryptoHash,
+)]
 pub struct Transaction {
 	data: Vec<u8>,
 	// Application priority is stored low to high, i.e., 0 is the highest priority.
 	application_priority: u64,
 	sequence_number: u64,
-	id: Id,
+	pub id: Id,
```
mzabaluev, line 52: The pub member now sticks out like a sore thumb. At least in movement-types I'd push for better API style.

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: Why you resolve the conversation with BlockHeight(0)? BlockHeight start at 1?

musitdev (approved):

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: I thought I had committed this, not sure what happened, maybe in the merge conflict something went wrong.

PR #1128, opened 2025-03-20
fix: correct indexer test

musitdev:
# Summary
- **RFCs**: [Link to RFC](#./link/to/rfc), [Link to RFC](#./link/to/rfc), or $\emptyset$.
- **Categories**: any of `protocol-units`, `networks`, `scripts`, `util`, `cicd`, or `misc`.

Correct the indexer e2e Test
# Changelog

# Testing

```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.celestia-local.indexer.hasura.indexer-test --keep-tui"
```
# Outstanding issues

PR #1151, opened 2025-03-25
feat: Add grpc call to Da sequencer node (server and client)

musitdev:
# Summary
See issue #1135 
First part create the grpc calls.

Add grpc call to Da sequencer node (server and client)

# Changelog
 * create grpc client
 * create grpc server
 * update config.

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/da-sequencer/node/src/server.rs
+++ b/protocol-units/da-sequencer/node/src/server.rs
@@ -1,34 +1,67 @@
+use movement_da_sequencer_proto::da_sequencer_node_service_server::{
+	DaSequencerNodeService, DaSequencerNodeServiceServer,
+};
+use movement_da_sequencer_proto::BatchWriteRequest;
+use movement_da_sequencer_proto::BatchWriteResponse;
+use movement_da_sequencer_proto::ReadAtHeightRequest;
+use movement_da_sequencer_proto::ReadAtHeightResponse;
+use movement_da_sequencer_proto::StreamReadFromHeightRequest;
+use movement_da_sequencer_proto::StreamReadFromHeightResponse;
+use std::net::SocketAddr;
+use tokio_stream::Stream;
+use tonic::transport::Server;
+
 /// Runs the server
-async fn run_server(&self) -> Result<(), anyhow::Error> {
+pub async fn run_server(address: SocketAddr) -> Result<(), anyhow::Error> {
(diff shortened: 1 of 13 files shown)


andygolay:
@musitdev do you recommend a specific way to test this?

process-compose/movement-full-node/process-compose.hasura.yml
```diff
@@ -12,6 +12,6 @@ processes:
       indexer:
         condition: process_healthy
     readiness_probe:
-      initial_delay_seconds: 30
+      initial_delay_seconds: 120
```
andygolay, line 15: What is the reason for the increase in `initial_delay_seconds` here?

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -1,34 +1,67 @@
...
+use movement_da_sequencer_proto::ReadAtHeightResponse;
+use movement_da_sequencer_proto::StreamReadFromHeightRequest;
+use movement_da_sequencer_proto::StreamReadFromHeightResponse;
+use std::net::SocketAddr;
+use tokio_stream::Stream;
+use tonic::transport::Server;
+
 /// Runs the server
-async fn run_server(&self) -> Result<(), anyhow::Error> {
+pub async fn run_server(address: SocketAddr) -> Result<(), anyhow::Error> {
```
andygolay, line 15: How do you actually run the server? `process-compose`?

0xmovses: `cargo run -p movement-da-sequencer-node` should do it. There is no dependency in any test here that requires the full-node or Celestia at this stage.

musitdev: Yes it's too early to test anything. The main is empty. It will be tested in a next PR when we've enough function implemented to start the server.

process-compose/movement-full-node/process-compose.hasura.yml
musitdev: Because the docker par take too much time to start so I increase the time to let process compose wait.
What I don't understand is why this code leak from another PR. Perhaps I start from it. So this change is not related to this PR but this one: #1128 that is merged.

0xmovses:
I had a read of this code. It looked good to me, I didn't run it locally, but fine.

PR #1154, opened 2025-03-26
feat: send batch to aptos-core mempool in dedicated task

andygolay:
# Summary

# Changelog

# Testing

# Outstanding issues


protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -199,47 +213,122 @@ impl TransactionPipe {
...
+						.celestia_da_light_node_config
+						.movement_da_light_node_connection_port()
 				);
-				self.core_mempool.commit_transaction(&transaction_sender, sequence_number);
-			}
 
-			// update the last send
-			self.last_mempool_send = Instant::now();
+				tokio::spawn(async move {
+					let Ok(mut da_client) = MovementDaLightNodeClient::try_http1(&connection_string) else {
```
musitdev, line 298: This code is not the right client. You should use the new da-sequencer client. The client is done bu not the data structure to send. Keep it like that for now and when the other PR will be merged, we come back to update.it.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -199,47 +213,122 @@ impl TransactionPipe {
...
+						.celestia_da_light_node
+						.celestia_da_light_node_config
+						.movement_da_light_node_connection_port()
 				);
-				self.core_mempool.commit_transaction(&transaction_sender, sequence_number);
-			}
 
-			// update the last send
-			self.last_mempool_send = Instant::now();
+				tokio::spawn(async move {
```
musitdev, line 297: The spawn return a join_handle that we can use to get the execution result. I'll will do it later if you don't see how.

musitdev (commented):
You can remove all code related to transaction ingress.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
andygolay: So, I should keep this PR in draft until then?

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
andygolay: I added a join_handle but I'm not sure how to got the execution result. Looking into it now.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -199,47 +223,122 @@ impl TransactionPipe {
...
+					.collect::<Result<Vec<_>, anyhow::Error>>()?;
+
+				let batch_write = BatchWriteRequest { blobs };
+				let mut buf = Vec::new();
+				batch_write.encode_raw(&mut buf);
+				info!("built_batch_write batch_id={} size={}", batch_id, buf.len());
+
+				let connection_string = format!(
+					"{}://{}:{}",
+					MOVEMENT_CONFIG
```
mzabaluev, line 293: Hmm, why should we have this globally intialized config that's separate from the normal config, which is part of the executor state?

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -199,47 +223,122 @@ impl TransactionPipe {
...
+								if attempts >= 3 {
+									warn!("giving up on batch_id={}", batch_id);
+									return Err(e.into());
+								}
+								tokio::time::sleep(std::time::Duration::from_secs(2)).await;
+							}
+						}
+					}
+				});
+				self.da_join_handles.push(handle);
```
mzabaluev, line 333: Why does this need to be a member of `TransactionPipe` state? I don't see where these join handles are joined, are they kept for the duration of the process?

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
andygolay: Oh, I was following a pattern I saw in the faucet, I didn't realize the config was already part of the executor state. I'll fix it.

andygolay: @mzabaluev any hints at how to do get this info

```
	.celestia_da_light_node
						.celestia_da_light_node_config
						.movement_da_light_node_connection_protocol(),
					MOVEMENT_CONFIG
						.celestia_da_light_node
						.celestia_da_light_node_config
						.movement_da_light_node_connection_hostname(),
					MOVEMENT_CONFIG
						.celestia_da_light_node
						.celestia_da_light_node_config
						.movement_da_light_node_connection_port()
```

from the Executor state? I see Executor has a config field:

```
pub struct Executor {
	/// Send commited Tx
	pub mempool_tx_exec_result_sender: futures::channel::mpsc::Sender<Vec<TxExecutionResult>>, // Sender, seq number)
	/// The executing type.
	pub block_executor: Arc<BlockExecutor<AptosVM>>,
	/// The signer of the executor's transactions.
	pub signer: ValidatorSigner,
	// Shared reference on the counter of transactions in flight.
	transactions_in_flight: Arc<RwLock<GcCounter>>,
	// The config for the executor.
	pub(crate) config: Config,
	/// The node config derived from the maptos config.
	pub(crate) node_config: NodeConfig,
}
```					

but in that field or in subfields of the `Config` struct, I don't see how to how to get DA light node details.

cc @musitdev @0xmovses as well if you could point me in the right direction for that, thanks.

protocol-units/execution/maptos/dof/src/lib.rs
```diff
@@ -27,7 +27,7 @@ pub trait DynOptFinExecutor {
 	/// Initialize the background task responsible for transaction processing.
 	fn background(
 		&self,
-		transaction_sender: Sender<(u64, SignedTransaction)>,
+		transaction_sender: Sender<Vec<(u64, SignedTransaction)>>,
```
musitdev, line 30: I think you can remove all these transaction_sender  / receiver because they shouldn't be used.

protocol-units/execution/maptos/dof/src/v1.rs
```diff
@@ -57,7 +57,7 @@ impl DynOptFinExecutor for Executor {
 
 	fn background(
 		&self,
-		transaction_sender: Sender<(u64, SignedTransaction)>,
+		transaction_sender: Sender<Vec<(u64, SignedTransaction)>>,
```
musitdev, line 60: The same

protocol-units/execution/maptos/opt-executor/src/background/task.rs
```diff
@@ -29,7 +29,7 @@ impl BackgroundTask {
 	pub(crate) fn transaction_pipe(
 		mempool_commit_tx_receiver: futures_mpsc::Receiver<Vec<TxExecutionResult>>, // Sender, seq number)
 		mempool_client_receiver: futures_mpsc::Receiver<MempoolClientRequest>,
-		transaction_sender: mpsc::Sender<(u64, SignedTransaction)>,
+		transaction_sender: mpsc::Sender<Vec<(u64, SignedTransaction)>>,
```
musitdev, line 32: The same

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -1,41 +1,62 @@
...
+});
+
 pub struct TransactionPipe {
 	// The receiver for Tx execution to commit in the mempool.
 	mempool_commit_tx_receiver: futures_mpsc::Receiver<Vec<TxExecutionResult>>,
 	// The receiver for the mempool client.
 	mempool_client_receiver: futures_mpsc::Receiver<MempoolClientRequest>,
 	// Sender for the channel with accepted transactions.
-	transaction_sender: mpsc::Sender<(u64, SignedTransaction)>,
+	transaction_sender: mpsc::Sender<Vec<(u64, SignedTransaction)>>,
```
musitdev, line 59: The same you can remove it.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -52,6 +73,8 @@ pub struct TransactionPipe {
 	used_sequence_number_pool: UsedSequenceNumberPool,
 	/// The accounts whitelisted for ingress
 	whitelisted_accounts: Option<HashSet<AccountAddress>>,
+	/// Join handles for DA write tasks to track and await each batch write result
+	da_join_handles: Vec<tokio::task::JoinHandle<anyhow::Result<()>>>,
```
musitdev, line 77: No don't store the join handle. I'll do the change after merge.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -89,6 +112,7 @@ impl TransactionPipe {
 				mempool_config.gc_slot_duration_ms,
 			),
 			whitelisted_accounts,
+			da_join_handles: vec![],
```
musitdev, line 115: No don't use it like that.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
musitdev: No, we can merge it if it builds then update it when the da-sequencer client pat is merged.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
musitdev: They don't need.

andygolay: I've removed it, and also removed the `transaction_sender` and `transaction_receiver` logic.

The latest commit builds but `cargo check --all-targets` throws errors due to issues with `tx_receiver` being expected in tests.

andygolay: @musitdev @musitdev I've adjusted some tests so `cargo check --all-targets` now finishes successfully. We'll need to fix the tests later.

protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
andygolay: Oh I think I understand. For this call it should use `movement_da_sequencer_listen_address` from `DaSequencerConfig`, right?

PR #1159, opened 2025-03-27
feat: da-seq produce block

0xmovses:
## Summary
See the issue: https://github.com/movementlabsxyz/movement/issues/1164

# Changelog
- Implements `produce_next_block`
- Adds new unit tests
- Improves code maintainability with general refactors
- Updates doc comments

## Testing
`cargo test -p movement-da-sequencer-node`

Output: 
```
running 14 tests
test batch::tests::test_sign_and_validate_batch ... ok
test storage::tests::test_try_new_invalid_path_should_fail ... ok
test storage::tests::test_produce_next_block_generates_block_and_clears_pending_tx ... ok
test storage::tests::save_block_should_persist_block_and_remove_pending_transactions ... ok
test storage::tests::test_produce_block_fits_all_tx_and_clears_pending ... ok
test storage::tests::test_try_new_creates_storage_successfully ... ok
test storage::tests::test_get_block_at_height_returns_correct_block ... ok
test storage::tests::test_write_batch_persists_transaction ... ok
test storage::tests::test_get_block_with_digest_returns_none_for_unknown_digest ... ok
test storage::tests::test_get_block_with_digest_returns_correct_block ... ok
test storage::tests::test_write_batch_naturally_orders_lexicographically_with_composite_key ... ok
test block::tests::test_sequencer_block_rejects_block_larger_than_max_size ... ok
test tests::client::test_batch_write_should_fail_with_wrong_sig ... ok
test tests::client::test_client_should_successfully_write_batch ... ok

test result: ok. 14 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.82s
```

## Outstanding issues
I did not implement the Scenario 2 test case described in the Issue because the `MAX_SEQUENCER_BLOCK_SIZE` is static const and not configurable by design.


protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+			DaSequencerError::StorageAccess("Missing colomn family: pending transactions".into())
+		})?;
+
+		// Collect and deserialize all pending transactions
+		let mut transactions: Vec<Transaction> = Vec::new();
+
+		let iter = self.db.iterator_cf(&cf_pending, rocksdb::IteratorMode::Start);
+		for item in iter {
+			let (_key, value) = item.map_err(|e| DaSequencerError::RocksDbError(e.to_string()))?;
+			let tx: Transaction = bcs::from_bytes(&value)
```
musitdev, line 204: You should keep the size of the Tx to help to evaluate the size of the block.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+				.map_err(|e| DaSequencerError::Deserialization(e.to_string()))?;
+			transactions.push(tx);
+		}
+
+		if transactions.is_empty() {
+			return Ok(None);
+		}
+
+		// Sort transactions using their Ord implementation
+		transactions.sort();
```
musitdev, line 214: You can't use the natural order of the movement_type::Transaction because it's not the order we need. Natural order is used on the fullnode size to build the batch. Here we need to order by timestamp first, then the batch order.
I think you need to add the batch timestamp to the Tx storage.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+
+		let mut selected_txs = Vec::new();
+		let mut block_size: u64 = 0;
+
+		for tx in &transactions {
+			let tx_size = bcs::to_bytes(tx)
```
musitdev, line 220: You don't have to serialize the Tx because you get them in byte from the DB; Get the size at that time.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+		let height: u64 = self.determine_next_block_height()?.into();
+
+		let parent_height = BlockHeight(height - 1);
+		let parent_digest = self
+			.get_block_at_height(parent_height)?
+			.ok_or_else(|| DaSequencerError::StorageFormat("Missing parent block".into()))?
+			.get_block_digest();
+
+		// Build the block
+		let tx_set: BTreeSet<_> = selected_txs.clone().into_iter().collect();
```
musitdev, line 242: No need to reiter, build the collection when you aggregate the tx.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+		let block = Block::new(BlockMetadata::default(), Id::new(parent_digest.0), tx_set);
+		let sequencer_block = SequencerBlock::try_new(self.determine_next_block_height()?, block)?;
```
musitdev, line 244: Use the calculated block height we use before. Don't recalculate.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+		})?;
+		let cf_digests = self.db.cf_handle(cf::BLOCKS_BY_DIGEST).ok_or_else(|| {
+			DaSequencerError::StorageAccess("Missing column family: blocks_by_digest".into())
+		})?;
+
+		let mut write_batch = WriteBatch::default();
+		let height_key = height.to_be_bytes();
+
+		write_batch.put_cf(&cf_blocks, height_key, &block_bytes);
+		write_batch.put_cf(&cf_digests, sequencer_block.get_block_digest().0, &height_key);
```
musitdev, line 260: Do public function to save block, it will be helpful for the test or when a block is updated (ex:sent to Celestia).

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: What do you mean by the batch order? The order at which the tx's were received?

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +191,84 @@ impl DaSequencerStorage for Storage {
...
+			}
+
+			block_size += tx_size;
+			selected_txs.push(tx.clone());
+		}
+
+		// Notes that this method will never return less that 1
+		let height: u64 = self.determine_next_block_height()?.into();
+
+		let parent_height = BlockHeight(height - 1);
```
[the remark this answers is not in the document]
0xmovses: Ok so handle the genesis case.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: yah it's cause i'm calling into() there, to turn it into a u64, here it needs `Blockheight` I'll improve it so we only do the call once.

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: In the #1164 issue, line 2 the order of Tx are `order the Tx by batch timestamp / tx order in batch` .
The Tx are ordered correctly by the fullnode in the batch, but not between batch. So first you need to order by batch (ie: batch timestamp, the oldest first) then the order in the batch.

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: BlockHeight is copy. Copy it before the into(). Something like ` (*&block_height).into() `
We should only call `determine_next_block_height() ` more than necessary because it queries the db.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: But, I have to deserialize them in order to sort them, which means I have to serialize again after for building the block.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: I've used Composite key to store in Rocksdb
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
struct TxCompositeKey {
	timestamp: u64,
	index_in_batch: u32,
	tx_id: [u8; 32],
}
```
This means that rocksdb natural ordering will be lexical (when using big endian encoding) and should be sufficient, also I added the test 
`test_write_batch_naturally_orders_lexicographically_with_composite_key`

Which shows that rocksdb correctly orders on timestamp even if written at different times. Let me know if think it's sufficient for our needs.

0xmovses: I've used Composite key to store in Rocksdb
```rust
#[derive(Debug, Clone, PartialEq, Eq)]
struct TxCompositeKey {
	timestamp: u64,
	index_in_batch: u32,
	tx_id: [u8; 32],
}
```
This means that rocksdb natural ordering will be lexical (when using big endian encoding) and should be sufficient, also I added the test 
`test_write_batch_naturally_orders_lexicographically_with_composite_key`

Which shows that rocksdb correctly orders on timestamp even if written at different times. Let me know if think it's sufficient for our needs.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: Now that sorting happens lexicographically by using the composite key, I do not need to order unserialized keys, so, it now just deals with serialized values, bytes directly from the DB.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: Done, and added a unit test for it.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: It's now only called once.

protocol-units/da-sequencer/node/src/storage/mod.rs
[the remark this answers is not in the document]
0xmovses: Handled with `Id::genesis_block()`

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: I get the size directly here now, no deserialization occurs 
```rust
for item in iter {
	let (key, value) = item.map_err(|e| DaSequencerError::RocksDbError(e.to_string()))?;
	let tx_size = value.len() as u64;
```

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +230,56 @@ impl DaSequencerStorage for Storage {
...
+			0 | 1 => Id::genesis_block(),
+			_ => {
+				let parent_block = self.get_block_at_height(height.parent())?.ok_or_else(|| {
+					DaSequencerError::StorageFormat("Missing parent block".into())
+				})?;
+				Id::new(parent_block.get_block_digest().0)
+			}
+		};
+
+		let tx_set: BTreeSet<_> = selected_txs.into_iter().collect();
```
musitdev, line 275: If you use a BTreeSet you'll reorder the Tx using the natural order and not the one we want.  I think we need to change the Block struct to put a Vec and order the Tx before adding.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -188,7 +230,56 @@ impl DaSequencerStorage for Storage {
 	}
 
 	fn produce_next_block(&self) -> Result<Option<SequencerBlock>, DaSequencerError> {
-		todo!()
+		let cf_pending = self.db.cf_handle(cf::PENDING_TRANSACTIONS).ok_or_else(|| {
+			DaSequencerError::StorageAccess("Missing column family: pending_transactions".into())
+		})?;
+
+		let iter = self.db.iterator_cf(&cf_pending, rocksdb::IteratorMode::Start);
```
musitdev, line 237: The ordering of RockDb should be tested. In fact, I don't understand how RocksDb can order with `TxCompositeKey` that doesn't implement Ord. Or you suppose the save order is the right order. I can be ok because timestamp is set when the batch is received, so after the process is nearly sequential. If 2 batch switch, it means they have been received nearly at the same time. It works because there's only one Sequencer, so the order can be non-determinist. What do you think?
Last point, if it's the case, you can avoid saving the Tx with the timestamp and order in batch.

musitdev (commented):
Some point about the ordering.

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: Ok yes I didn't see your remarks. Yes it should be ok. I think we must add some test about it. We'll open a new PR for the test.

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: Don't take into account I just see your other remarks.

protocol-units/da-sequencer/node/src/storage/mod.rs
0xmovses: Ah yes, collect and will re order. I will fix.

0xmovses: We resolved this over slack, we are staying with this for now.

protocol-units/da-sequencer/node/src/block.rs
```diff
@@ -38,7 +57,9 @@ pub struct SequencerBlock {
 impl SequencerBlock {
 	/// Try to construct a SequencerBlock, but fail if it exceeds the max encoded size.
 	pub fn try_new(height: BlockHeight, block: Block) -> Result<Self, DaSequencerError> {
-		todo!()
+		let sb = SequencerBlock { height, block };
+		sb.validate_size()?;
```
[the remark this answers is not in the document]
0xmovses: Done

musitdev (approved):

Issue #1157, opened 2025-03-27
da-sequencer: implement batch signature and verification


musitdev:
Part of issue #1135, this issue describes the process to sign and verify batches.

On the full node, batches are signed before being sent. The signing process uses the movement-sign API to store the private key in the config file. Signature verification is performed by the da-sequencer once the batch has been received. The verification uses the standard Aptos public key and signature. See [batch signature](https://github.com/movementlabsxyz/movement/blob/feature/da-sequencer-node/protocol-units/da-sequencer/node/src/batch.rs). Several pull requests are currently under review that modify this, so you will need to update it depending on the PR that is merged.

For serialisation of the Vec<tx> use bsc instead of serde_json. See the Block-producing / Storage PR that uses it.

Main dev to do:
* add the new key in the da-sequencer config file. Init it by default with a local generated key. Log::info the public key during this process so that it can be declared to the da-sequencer later.
* implement the function `sign_batch(batch_data: &[u8], private_key: Private key) -> Signature;` This function will be added to the [da-sequencer client file](https://github.com/movementlabsxyz/movement/blob/feature/da-sequencer-node/protocol-units/da-sequencer/client/src/lib.rs) as an independent function like [validate_batch](https://github.com/movementlabsxyz/movement/blob/3e540cbb514cf4515af2da5155db9c9b13fbe251/protocol-units/da-sequencer/node/src/batch.rs#L20C4-L20C18). It take the batch tx array serialized (bsc) and the private key. 
* Do a test that signed and verifies a simple batch of Tx.
* Implements the public keys whitelist file management (see issue #1135). Implements only the read function that will update the list of public key loaded in memory. The periodic process will be implemented by the main loop.
* implement the [validate_batch](https://github.com/movementlabsxyz/movement/blob/3e540cbb514cf4515af2da5155db9c9b13fbe251/protocol-units/da-sequencer/node/src/batch.rs#L20C4-L20C18) function that verify the batch signature. It takes the Batch<RawData> that contains the Tx bytes array, the signature and the key. 
   * verify that the Tx bytes are correctly signed with the provided public key.
   * Verify the public key is part of the whitelisted key in the file.
   * Unserialize (bsc) the Tx byte array to get an array of Tx.
   * Build the Batch<FullnodeTx> and return it. See PR #1150 for the FullnodeTx struct content.

Other batch verification is not done in this PR. This PR concentrates on batch sign and verification.

PR #1161, opened 2025-03-27
feat: da-sequencer: implement batch signature and verification

andygolay:
# Summary
- #1157 

# Changelog

- In `da-sequencer` client lib, add `sign_batch` function.
- Add `signing_key` to `DaSequencerConfig` and provide default randomly-generated key
- Implement `validate_batch` function and add helper function `verify_batch_signature`, to validate BSC-serialized transaction batch. 
- Add `test_sign_and_validate_batch` test.
- Use `FullNodeTxns` for transaction batch
- Merge feature branch and update `test_sign_and_validate_batch` to use `serialize_full_node_batch` and `deserialize_full_node_batch`.
- Update to use `ed25519_dalek` rather than `aptos_crypto` deps as this is only used for validating in the DA sequencer.

# Testing

- In `protocol-units/da-sequencer/node` run `cargo test -- --nocapture`. The validation (public) key is logged and the message signed with the signing (private) key from the config is validated.

# Outstanding issues
- Allowlist is not yet implemented
- Signing key defaults to random bytes; this needs to be secured and we need to decide how to add the actual key we'll use, into the Config.
- many warnings to resolve before we complete the feature branch


protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -1,24 +1,95 @@
...
 pub struct RawData {
 	pub bytes: Vec<u8>,
 }
 
-#[derive(Debug)]
-pub struct FullnodeTx {}
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct FullnodeTx {
+	pub id: u64,
+	pub data: String,
```
musitdev, line 14: data is a Vec<Transaction> see PR #1150 and get the definition.

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -1,24 +1,95 @@
...
-/// Batch write blobs.
-fn validate_batch(
+pub fn validate_batch(
 	new_batch: DaBatch<RawData>,
-) -> std::result::Result<DaBatch<FullnodeTx>, DaSequencerError> {
-	todo!()
+) -> Result<DaBatch<Vec<FullnodeTx>>, DaSequencerError> {
+	verify_batch_signature(&new_batch.data.bytes, &new_batch.signature, &new_batch.signer)?;
+
+	let txs: Vec<FullnodeTx> = bcs::from_bytes(&new_batch.data.bytes)
```
musitdev, line 29: See PR: #1155 and how grpc data are [deserialized](https://github.com/movementlabsxyz/movement/blob/6443612af3ba6f53faf66c87db1b5888351076c3/protocol-units/da-sequencer/node/src/server.rs#L76).

`data_bytes` is a Vec of Transactions.

musitdev (commented):
Change the FullNodeTx struct data, for the rest it's fine.
The file public key verification will be done in another PR.

musitdev (approved):

Issue #1158, opened 2025-03-27
da-sequencer: Implements bootstrap algo.


musitdev:
See PR #1131 for initial conception.

The role of the bootstrap algorithm is to start the Celestia client to ensure that it doesn't miss any block digests and doesn't send the same digest twice. The Celestia client buffers block digests in memory before sending them in a blob. If a crash occurs, these digests are lost. The role of the bootstrap is to recover these digest and the new one created during the crash. The celestial client can fail an the main node process still producing block.

Since all digests sent to the Celestia client are stored by the storage, they can be requested. The goal of this algorithm is to determine which digests should be requested using the last sent digest to Celestia and the last Celestia DA (Data Availability) blob processed.

The algorithm can use two types of trust sources:
   * The Celestia network: Be cautious when the node starts, as some blobs may still be produced by the Celestia network. The function needs to wait to ensure that no blob is pending in the Celestia network, then fetch the last finalized blob to examine the last processed blob digest and obtain the corresponding block to determine the last push block height.
  *  The local database: The blocks sent and the finalized block heights (both DA and block) are saved in the database and can be used for determining the bootstrap height. However, if some digests in the database are marked as sent but not finalized, they must be verified against the Celestia network to see if they have been finalized during the node crash, or the algorithm must wait for their processing to complete.

When the next block digest height is determined, the algorithm fetches all blocks up to the current last block height and sends them to Celestia in one or several blobs (if the maximum blob size is reached), containing all pending digests.

Once this is done, the new block digests added during the bootstrap process are integrated into the normal Celestia process.

The normal Celestia process aggregates all received block digests and sends them in blobs when the Celestia network is available and no blob is being processed.

This PR concentrates on the main bootstrap algo and not all Celestia stuff. If some Celestia data are needed they will be simulated via a specific mock function.

[...]

PR #1162, opened 2025-03-28
feat(da-sequencer): Implement the external DA bootstrapping

sebtomba:
# Summary
- #1158 

# Changelog

- Draft the initial idea to bootstrate the Celestia external DA.

# Testing

[WIP]

# Outstanding issues
- Missing functionality to retrieve a block for a digest (block id).
- Implementation of the other functions in `CelestiaExternalDa`.
- Handle errors gracefully.


protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -1,101 +1,150 @@
...
+			let blobs = blobs.expect("get_blobs_at_height returned None");
+			let last = blobs.last().expect("get_blobs_at_height returned no Blobs");
+			let _digest = last.0.last().expect("get_blobs_at_height returned an empty Blob");
+
+			// Step 2: Get the Block for digest
+			// TODO: Request the block for digest
+			let block: SequencerBlock = SequencerBlock::default();
+
+			// Step 3: Request and send all missing blocks
+			for height in block.get_height().0..=current_block_height.0 {
```
musitdev, line 136: You should skip the `block.get_height()` because it has already been sent.

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -1,101 +1,150 @@
...
-	/// The block is not immediatly sent but aggergated in a blob
+	/// The block is not immediately sent but aggregated in a blob
 	/// Until the client can send it to celestia.
-	async fn send_block(
+	fn send_block(
 		&self,
-		block: &SequencerBlockDigest,
-	) -> Pin<Box<dyn Future<Output = std::result::Result<(), DaSequencerError>> + Send>>;
+		block: SequencerBlockDigest,
+	) -> Pin<Box<dyn Future<Output = Result<(), DaSequencerError>> + Send + '_>>;
```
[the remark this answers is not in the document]
sebtomba: I see. Now changed it to `imply Future`.

protocol-units/da-sequencer/node/src/celestia/mod.rs
sebtomba: Right.

.gitignore
```diff
@@ -24,5 +24,7 @@ venv
 *.jot.*
 *.jot
 *.env
+*.envrc.local
```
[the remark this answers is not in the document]
sebtomba: I'm using `direnv` to run `nix flake` immediately when I `cd` into the directory. This requires adding `use flake` into the `.envrc` file. On the other hand, in some projects, the `.envrc` file is used for different things. Instead of adding this file to `.gitignore`, a `.envrc.local` gets included if it exists.

I can create a separate PR for that if required.

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -1,101 +1,150 @@
...
 }
 
 #[derive(Clone, Default, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
 pub struct CelestiaHeight(u64);
 
+impl<T: Into<u64>> Add<T> for CelestiaHeight {
+	type Output = Self;
+
+	fn add(self, rhs: T) -> Self::Output {
+		CelestiaHeight(self.0.saturating_add(rhs.into()))
```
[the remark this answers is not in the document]
sebtomba: Right.

andygolay:
Looks like a couple conflicts to resolve @sebtomba

andygolay (changes requested):
When I run `cargo check --all-targets` I get a couple errors:

```
error[E0515]: cannot return value referencing function parameter `b`
   --> protocol-units/da-sequencer/node/src/celestia/mod.rs:141:18
    |
141 |             .and_then(|b| b.last())
    |                           -^^^^^^^
    |                           |
    |                           returns a value referencing data owned by the current function
    |                           `b` is borrowed here

error[E0382]: borrow of moved value: `last_finalized_celestia_height`
   --> protocol-units/da-sequencer/node/src/celestia/mod.rs:145:5
    |
132 |         last_finalized_celestia_height: CelestiaHeight,
    |         ------------------------------ move occurs because `last_finalized_celestia_height` has type `CelestiaHeight`, which does not implement the `Copy` trait
...
139 |             .get_blobs_at_height(last_finalized_celestia_height)
    |                                  ------------------------------ value moved here
...
145 |                 last_finalized_celestia_height.0
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ value borrowed here after move
    |
```

andygolay (approved):
LGTM pending integration of @musitdev's suggestions.

andygolay (changes requested):
Just noticed there are still conflicts to resolve. Otherwise LGTM.

andygolay (approved):

0xmovses (approved):
This looks good!

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -1,93 +1,160 @@
...
+				last_finalized_celestia_height.0
+			)))?;
+
+		// Step 2: Get the Block for digest
+		let mut block = self.request_block(BlockAt::Digest(digest.clone())).await?;
+
+		// Step 3: Request and send all missing blocks
+		for height in (block.height.0 + 1)..=current_block_height.0 {
+			block = self.request_block(BlockAt::Height(BlockHeight(height))).await?;
+			self.send_block(block.get_block_digest()).await?;
```
musitdev, line 155: During the bootstrap process, new blocks digests can be sent by the main loop using `CelestiaClient.send_block()`.
You should add a way to differentiate these block that must be sent after yours to keep the order of all blocks.

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -1,93 +1,160 @@
...
+	async fn bootstrap(
 		&self,
 		current_block_height: BlockHeight,
-		last_sent_block_height: BlockHeight,
-		last_notified_celestia_height: CelestiaHeight,
+		last_finalized_celestia_height: CelestiaHeight,
 	) -> Result<(), DaSequencerError> {
-		todo!()
+		// wait to ensure that no blob is pending in the Celestia network
+		tokio::time::sleep(DELAY_SECONDS_BEFORE_BOOTSTRAPPING).await;
```
musitdev, line 136: During this wait, new blobs can arrive, so you need to update the `last_finalized_celestia_height` if any.
To get this info you can add a call to the Celestia client, a sort of notification for new blob for example. See with @mzabaluev the best way to be updated.

sebtomba: This makes sense. I will remove the `last_finalized_celestia_height` parameter from this function and require the `CelestiaClient` to provide a method for retrieving the current celestia height.

musitdev: You need this parameter in case there's no pending blob, so keep it. I just say that you need to get new committed blob during that wait. If there's any they will update this parameter.

protocol-units/da-sequencer/node/src/celestia/mod.rs
musitdev: Yes you can merge, and you'll do new PRs for the 2 issues (`last_finalized_celestia_height` and `send_block()`)

PR #1163, opened 2025-03-28
feat(da): accumulate blobs while submit is pending

### LINKED ISSUE #1131: [DA-node] construct aggregate blob of blocks
## Summary

Contains all the Celestia development. All celestia functionalities are integrated in the Celestia client.

All block must be push in Celestia like today. Only the digest is written in Celestia.
It's the main process that decide when a block should be push.
A bootstrap process is necessary to guaranty that all block are pushed in case of crash.

Provide a way to get any blocks or blob at any Celestia height.

Implement a bootstrap mechanism to avoid loosing block in case of crash.

## Implementation Details 

As Celestia blob propagation is a lot slower than block production, Celestia blob will contain 1 or several blocks.
* The main Da Sequencer process send new block digest at any moment
*  All block digest must be sent to Celestia network using the Celestia client like today.
* While the Celestia client is sending, new block digest received are keep until the client can send them. When the client returns, If there's some block digest pending, a new blob is constructed by creating an array of block digest in the order they were produced, first produced at first. A new block is added to the array only if by adding it, the blob size won't exceed the max Celesta block size. 
* Block height are put with the block digest so we can get the height of each block digest contains by a blob.
* When a block digest has been sent, the Celestia client notifies the DA sequencer process about it.
* The celestia client allows getting any blob and its content at any blob Celestia DA height.
* When a block digest is commited on celestia the Celestia client notify the new height to the DA sequencer process.

To avoid missing a block digest in case of node crash, a bootstrap mechanism allows getting the missing block digest from last sent block height to the current block height to Celestia. During this process, the Celestia client aggregate as many block it can in one Celestia blob.

The Celestia client should recover from celestia full node connection issue. If it can't recreate the connection after some time, the Celestia client should notify the issue to the Da Sequecner main loop.

This development will use current Celestia [code](https://github.com/movementlabsxyz/movement/blob/main/protocol-units/da/movement/providers/celestia/src/da/mod.rs).

## Testing 

- Test normal block send to the celestia client and Block height when sent notification and Celetia da height notification when the blob is committed.
- Test bootstrap when there are old blocks and new block arrives during the old block processing. Verify the block ordering.
- Test Celestia connection close / open, stay close and notify to the Da Sequencer main loop.
Write some blob, get them in an unordered way.

mzabaluev:
Resolves #1131

Implement the blob buffering task for Celestia
blob submission.

Remove the boxfest in the DaOperations trait.


protocol-units/da/movement/providers/celestia/src/da/blob_submitter.rs
```diff
@@ -0,0 +1,81 @@
...
+				Some(pending_request) => {
+					// While a submit request is pending, accumulate blobs.
+					// Provide back-pressure if the data size is pushing
+					// against the Celestia sanity limit.
+					select! {
+						res = pending_request => {
+							res?;
+							submit_request = None;
+						}
+						next = self.blob_receiver.recv(), if total_data_size < CELESTIA_BLOB_SIZE_THRESHOLD => {
```
musitdev, line 57: If the size is > to CELESTIA_BLOB_SIZE_THRESHOLD the submitted blob is lost. It should be kept for the next submission.

0xmovses:
Yeah, Vecparty over a Boxfest anyday.

protocol-units/da/movement/providers/celestia/src/da/blob_submitter.rs
mzabaluev: No, the `if` clause in select blocks polling. So the channel will keep the blobs and eventually there will be back-pressure on the sending side.

protocol-units/da-sequencer/node/src/celestia/submit.rs
```diff
@@ -0,0 +1,116 @@
...
+					// While a submit request is pending, accumulate blobs.
+					// Provide back-pressure if the data size is pushing
+					// against the Celestia sanity limit.
+					select! {
+						res = pending_request => {
+							let (block_height, celestia_height) = res?;
+							if self
+								.notifier
+								.send(
+									ExternalDaNotification::BlockCommitted(block_height, celestia_height)
```
musitdev, line 73: The block has been submitted not committed. We need to wait for the Celestia to produce the blob.

mzabaluev: Isn't this what [celestia RPC method](https://node-rpc-docs.celestia.org/?version=v0.21.9#blob.Submit) does?

protocol-units/da-sequencer/node/src/celestia/submit.rs
```diff
@@ -0,0 +1,116 @@
...
+							digests,
+						)));
+					}
+				}
+				Some(pending_request) => {
+					// While a submit request is pending, accumulate blobs.
+					// Provide back-pressure if the data size is pushing
+					// against the Celestia sanity limit.
+					select! {
+						res = pending_request => {
```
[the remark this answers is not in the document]
musitdev: You can see what I 've done for the[ block producing](https://github.com/movementlabsxyz/movement/blob/f9810247156d27dae291afb47bebdcc93cf910cc/protocol-units/da-sequencer/node/src/lib.rs#L96) that is similar in terms of needs. I force only one producing at a time.

protocol-units/da-sequencer/node/src/celestia/submit.rs
musitdev: Yes, I didn't know. I thought you need to submit and then pull until you get the height like it's done in the current client. Ok, so it's fine.
Can you look in the current client why we've changed, the height pulling to pull certificate then pull height? Perhaps we need to pull a proof with `GetProof` or something like that?

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -41,44 +45,53 @@ pub trait DaSequencerExternalDa: Clone {
...
 	/// Use during bootstrap to request block that are missing on Celestia network.
 	RequestBlock { at: BlockAt, callback: oneshot::Sender<Option<SequencerBlock>> },
 }
 
-#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub enum BlockAt {
 	Height(BlockHeight),
-	Digest(SequencerBlockDigest),
+	Digest([u8; SequencerBlockDigest::DIGEST_SIZE]),
```
sebtomba, line 67: Shall we also change the `BlockHeight` to `u64`?

mzabaluev: I think it's better to keep the domain type as the parameter. I would create two separate methods, unless you have reasons to commingle them with the `BlockAt` enum.

sebtomba: Since @mzabaluev added a `BlockHeight` to the `SequencerBlockDigest`, which we keep in the Celestia `Blob`, the bootstrapping algo doesn't need to request blocks for a digest to find out the block height. The `enum BlockAt` can be removed.

protocol-units/da-sequencer/node/src/celestia/submit.rs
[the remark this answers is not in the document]
mzabaluev: I have checked the code of `jsonrpsee` (which is what celestia-rpc uses under the hood), and it looks like request futures are cancellation-safe, e.g. https://docs.rs/jsonrpsee-core/0.24.9/src/jsonrpsee_core/client/async_client/mod.rs.html#508

The RPC client uses an actor model, the request is sent to the background task over the MPSC channel (cancellable atomically), and then the response is polled from a oneshot (cancellable, the sending code on the background [ignores the dropped receiver](https://github.com/paritytech/jsonrpsee/blob/042d8fd7f870aa67db7afbcea1834947264e8716/core/src/client/async_client/helpers.rs#L190).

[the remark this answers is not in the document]
musitdev: Ok if you think it's safe. You should add a comment with this remark.

musitdev (approved):

protocol-units/da-sequencer/node/src/celestia/submit.rs
mzabaluev: In the current architecture, submitting blobs and getting blobs for the current height is done in separate tasks. There may be more than one node submitting blobs, and Celestia serves as the sequencer.

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -99,19 +112,19 @@ impl<C: CelestiaClient + Sync + Clone> CelestiaExternalDa<C> {
...
 
-impl<C: CelestiaClient + Sync + Clone> DaSequencerExternalDa for CelestiaExternalDa<C> {
+impl<C: CelestiaClientOps + Sync + Clone> DaSequencerExternalDa for CelestiaExternalDa<C> {
 	/// Send the given block to Celestia.
 	/// The block is not immediately sent but aggregated in a blob
 	/// until the client can send it to celestia.
-	async fn send_block(&self, _block: SequencerBlockDigest) -> Result<(), DaSequencerError> {
-		todo!()
+	async fn send_block(&self, block: SequencerBlockDigest) -> Result<(), DaSequencerError> {
+		self.celestia_client.send_block(block).await
```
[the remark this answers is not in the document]
mzabaluev: Added this behavior in [93da852](https://github.com/movementlabsxyz/movement/pull/1163/commits/93da8526cc8cdcd02a088a9e52ca382c9016090c)

andygolay (changes requested):
I get a few errors with `cargo check --all-targets`:

```
error[E0433]: failed to resolve: use of undeclared type `BlobSubmitter`
  --> protocol-units/da/movement/providers/celestia/src/da/mod.rs:38:24
   |
38 |         let blob_submitter = BlobSubmitter::new(Arc::clone(&default_client), blob_receiver);
   |                              ^^^^^^^^^^^^^ use of undeclared type `BlobSubmitter`

For more information about this error, try `rustc --explain E0433`.
error[E0432]: unresolved import `crate::celestia::blob::Blob`
 --> protocol-units/da-sequencer/node/src/tests/mock.rs:6:5
  |
6 | use crate::celestia::blob::Blob;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `Blob` in `celestia::blob`
  |
help: consider importing one of these items instead
  |
6 | use aptos_sdk::types::executable::FetchedModule::Blob;
  |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
6 | use celestia_types::Blob;
```
and

```
error[E0599]: no method named `into_vec` found for struct `SequencerBlockDigest` in the current scope
   --> protocol-units/da-sequencer/node/src/server.rs:254:39
    |
254 |             blobckid: block.get_block_digest().into_vec(),
    |                                                ^^^^^^^^
    |
   ::: protocol-units/da-sequencer/node/src/block.rs:10:1
    |
10  | pub struct SequencerBlockDigest {
    | ------------------------------- method `into_vec` not found for this struct
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `into_vec`, perhaps you need to implement one of them:
            candidate #1: `OsStringExt`
            candidate #2: `bstr::ext_vec::ByteVec`
help: there is a method `into` with a similar name
    |
254 |             blobckid: block.get_block_digest().into(),
    |                                     
```

mzabaluev:
@andygolay Sorry, forgot to restore the legacy crate I started off modifying.

andygolay (approved):

protocol-units/da-sequencer/node/src/celestia/mod.rs
musitdev: So you remove it in this PR or a next one?

musitdev (approved):

protocol-units/da-sequencer/node/src/celestia/mod.rs
mzabaluev: Heads up: I have converted this explicit array into the `block::Id` domain type.

mzabaluev:
Unit tests fail, but seemingly the same way as in the feature branch.

Issue #1165, opened 2025-03-29
da-sequencer: allowlist for batch signer public keys


andygolay:
Issue #1157 mentions:

- Implements the public keys whitelist file management (see issue https://github.com/movementlabsxyz/movement/issues/1135).

- and in the `validate_batch` function:

> implement the validate_batch function that verify the batch signature. It takes the Batch that contains the Tx bytes array, the signature and the key.
> verify that the Tx bytes are correctly signed with the provided public key.
> Verify the public key is part of the whitelisted key in the file.

Parent issue #1117 also asks to

> Validate if the current whitelist validation is still required.

As far as I understand, it is required because we only want allowed parties to validate batches. But if that's not the case we should clarify that point.

musitdev:
Yes, it's required.

PR #1168, opened 2025-04-01
feat: signer whitelist file management

andygolay:
# Summary
- [**RFCs**: [Link to RFC](#./link/to/rfc), [Link to RFC](#./link/to/rfc), or $\emptyset$.](https://github.com/movementlabsxyz/movement/issues/1165)
- **Categories**: any of `protocol-units`, `da sequencer node`.

# Changelog

- Add `whitelist.rs` with `Whitelist` struct and methods to initialize global WL (including reloading once per minute), set keys in list, clear list, etc.
- Update `validate_batch` fn to check that the WL contains the signer's pubkey
- Add two tests, `test_sign_and_validate_batch_passes_with_whitelisted_signer` and `test_sign_and_validate_batch_fails_with_non_whitelisted_signer` 
- Check that all other crate tests pass and `cargo check --all-targets` passes.

# Testing

- `cargo test -p movement-da-sequencer-node`

# Outstanding issues


protocol-units/da-sequencer/node/Cargo.toml
```diff
@@ -9,11 +9,16 @@ homepage.workspace = true
 publish.workspace = true
 rust-version.workspace = true
 
+[lib]
+name = "movement_da_sequencer_node"
+path = "src/lib.rs"
+
```
mzabaluev, line 15: Why would you need to specify this explicitly? This is the default name and top-level source file.

protocol-units/da-sequencer/node/Cargo.toml
```diff
@@ -33,6 +38,7 @@ url = { workspace = true }
 futures = { workspace = true }
 dot-movement = { workspace = true }
 godfig = { workspace = true }
+serial_test = "2"
```
mzabaluev, line 41: This is a non-workspace dependency, our normal practice is to manage versions in the workspace `Cargo.toml`.

Is it needed in production code? If this is for tests, move it to the dev-dependencies section.

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -1,182 +1,136 @@
...
-use ed25519_dalek::{Signature, Verifier, VerifyingKey};
+use ed25519_dalek::{Signature, Signer, Verifier, VerifyingKey};
 use movement_types::transaction::Transaction;
 use serde::{Deserialize, Serialize};
 use std::ops::Deref;
 
 #[derive(Debug, Clone)]
 pub struct RawData {
-	pub data: Vec<u8>,
+        pub data: Vec<u8>,
```
mzabaluev, line 13: Please don't reformat files like that. This brings a lot of unneeded diff noise. Check that your editor can pick up settings from `rustfmt.toml`.

NB: it would be good, in my opinion, to switch to space indentation, but use the standard 4 spaces.

mzabaluev (changes requested):

protocol-units/da-sequencer/node/Cargo.toml
andygolay: I've moved to to dev-dependencies.

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,107 @@
...
+                        match Self::load(&path) {
+                                Ok(updated) => {
+                                        if let Ok(mut guard) = inner.write() {
+                                                *guard = updated;
+                                        } else {
+                                                eprintln!("[whitelist] Failed to acquire write lock");
+                                        }
+                                }
+                                Err(err) => {
+                                        eprintln!("[whitelist] Reload failed: {}", err);
```
musitdev, line 62: The same

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -77,6 +78,12 @@ pub fn deserialize_full_node_batch(
 pub fn validate_batch(
 	new_batch: DaBatch<RawData>,
 ) -> Result<DaBatch<FullNodeTxs>, DaSequencerError> {
+	let whitelist = Whitelist::get();
+
+	if !whitelist.contains(&new_batch.signer) {
+		return Err(DaSequencerError::InvalidSigner);
+	}
+
 	verify_batch_signature(&new_batch.data.data, &new_batch.signature, &new_batch.signer)?;
```
musitdev, line 87: Verify the signature first.

musitdev (commented):
Mainly to remove the singleton that is not needed.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -15,17 +18,17 @@ use tonic::transport::Server;
 pub async fn run_server(
 	address: SocketAddr,
 	request_tx: mpsc::Sender<GrpcRequests>,
+	whitelist: Arc<RwLock<Whitelist>>,
```
musitdev, line 21: Don't need to pass the Arc Mutex, just pass the white list then create the mutex inside the function.
Mutex is specific to the function implementation.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -62,29 +66,44 @@ impl DaSequencerNodeService for DaSequencerNode {
...
+			match crate::batch::deserialize_full_node_batch(batch_data) {
+				Ok(res) => res,
+				Err(err) => {
+					tracing::warn!("Invalid batch send: deserialization failed: {err}");
+					return Ok(tonic::Response::new(BatchWriteResponse { answer: false }));
+				}
+			};
+
+		// Read whitelist asynchronously
+		let whitelist = self.whitelist.read().await;
```
musitdev, line 87: Put the whole mutex lock in a section ({..}), because with this code the `self.request_tx.send` code is call with the lock still taken, and this call can take time and block other task unnecessarily.
Always force the lifetime of mutex lock to the shortest code.

andygolay: Okay I've fixed this and the above issue of passing the Arc Mutex, ready for re-review.

protocol-units/da-sequencer/node/src/batch.rs
```diff
@@ -76,14 +77,22 @@ pub fn deserialize_full_node_batch(
...
+	if !whitelist.contains(&new_batch.signer) {
+		return Err(DaSequencerError::InvalidSigner);
+	}
 
-	let txs: FullNodeTxs = bcs::from_bytes(&new_batch.data.data)
+	let data = bcs::from_bytes::<FullNodeTxs>(&new_batch.data.data)
 		.map_err(|_| DaSequencerError::DeserializationFailure)?;
 
+	// Disable this to see if signature is the problem
+	// if !new_batch.signer.verify(&new_batch.data.data, &new_batch.signature).is_ok() {
```
musitdev, line 90: Reenable it and put it in first before the whitelist verification.

protocol-units/da-sequencer/node/src/main.rs
```diff
@@ -27,12 +32,16 @@ async fn main() -> Result<(), Box<dyn Error>> {
 		Godfig::new(ConfigFile::new(config_file), vec![]);
 	let da_sequencer_config: DaSequencerConfig = godfig.try_wait_for_ready().await?;
 
+	// Initialize whitelist
+	let mut whitelist_path = dot_movement.get_path().to_path_buf();
+	whitelist_path.push("default_signer_address_whitelist");
```
musitdev, line 37: Put the path in the config, the same in relative to dot_movement. Use the `da-sequence`r config base path to have all `da-sequencer` file in the same directory.

musitdev (commented):
I didn't see the file path. It's better to have it in the configuration.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -177,27 +181,47 @@ impl DaSequencerNodeService for DaSequencerNode {
...
```
musitdev, line 202: Remove this lock access, you move to the validate section.

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,99 @@
...
+use ed25519_dalek::VerifyingKey;
+use hex::FromHex;
+
+#[derive(Clone)]
+pub struct Whitelist {
+	inner: Arc<RwLock<HashSet<VerifyingKey>>>,
+}
+
+impl Whitelist {
+	pub fn load(path: &PathBuf) -> std::io::Result<HashSet<VerifyingKey>> {
```
musitdev, line 19: Put the `load` and `start_reload_thread` private they aren't call outside the file.

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,99 @@
...
+	pub fn insert(&mut self, key: VerifyingKey) {
+		self.inner.write().unwrap().insert(key);
+	}
+
+	pub fn from_file_and_spawn_reload_thread(path: PathBuf) -> std::io::Result<Self> {
+		let set = Self::load(&path)?;
+		let inner = Arc::new(RwLock::new(set));
+		Self::start_reload_thread(inner.clone(), path);
+		Ok(Self { inner })
+	}
```
musitdev, line 98: All these test parts can be put in the test section. You just have to redefine an `impl Whitelist{}` in the test section and move this code there.

musitdev (commented):
Some minor correction, I think after it should be ok.

protocol-units/da-sequencer/node/src/error.rs
```diff
@@ -23,6 +23,8 @@ pub enum DaSequencerError {
 	DeserializationFailure,
 	#[error("Signature was invalid")]
 	InvalidSignature,
+	#[error("Non-whitelisted signer")]
+	InvalidSigner,
```
0xmovses, line 27: Invalid suggests format or other errors, whereas this case explicitly deals with authorization. 
```suggestion
	UnauthorizedSigner,
```

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,77 @@
...
+use hex::FromHex;
+
+#[derive(Clone)]
+pub struct Whitelist {
+	pub(crate) inner: Arc<RwLock<HashSet<VerifyingKey>>>,
+}
+
+impl Whitelist {
+	fn load(path: &PathBuf) -> std::io::Result<HashSet<VerifyingKey>> {
+		let content = fs::read_to_string(path)?;
```
0xmovses, line 20: Import the result above, then just `Result<HashSet<VerifyingKey>>`, reducing the chars in the fn sig.

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,77 @@
...
+
+		for line in content.lines() {
+			let trimmed = line.trim().trim_start_matches("0x");
+			if trimmed.is_empty() {
+				continue;
+			}
+
+			let key_bytes = <[u8; 32]>::from_hex(trimmed)
+				.map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid hex"))?;
+
```
0xmovses, line 31: I would to same with Error, seems that this whole file deails with std::io:: types so it's ok to import at top.

protocol-units/da-sequencer/node/src/whitelist.rs
```diff
@@ -0,0 +1,77 @@
...
+			let verifying_key = VerifyingKey::from_bytes(&key_bytes)
+				.map_err(|_| std::io::Error::new(std::io::ErrorKind::InvalidData, "Invalid key"))?;
+
+			set.insert(verifying_key);
+		}
+
+		Ok(set)
+	}
+
+	fn start_reload_thread(inner: Arc<RwLock<HashSet<VerifyingKey>>>, path: PathBuf) {
```
0xmovses, line 41: The `inner` type has become rather long. I would give it an alias
`pub type WhitelistFile = Arc<RwLock<HashSet<VerifyingKey>>>;`

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -177,27 +181,44 @@ impl DaSequencerNodeService for DaSequencerNode {
...
-				validate_batch(DaBatch::<RawData>::now(public_key, signature, bytes))
-			},
-		) {
-			Ok(batch) => batch,
-			Err(err) => {
-				tracing::warn!(error = %err, "Invalid batch send, verification / validation failed.");
-				return Ok(tonic::Response::new(BatchWriteResponse { answer: false }));
+
+		// Try to deserialize the batch
+		let (public_key, signature, bytes) =
```
0xmovses, line 192: Instead of matching on the error here, use `map_or_else` for a more concise expression.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -177,27 +181,44 @@ impl DaSequencerNodeService for DaSequencerNode {
...
+					tracing::warn!("Invalid batch send: deserialization failed: {err}");
+					return Ok(tonic::Response::new(BatchWriteResponse { answer: false }));
+				}
+			};
+
+		// Validate the batch
+		let validated = {
+			let whitelist = self.whitelist.read().await;
+			let raw_batch = DaBatch::<RawData>::now(public_key, signature, bytes);
+			match validate_batch(raw_batch, &whitelist) {
```
0xmovses, line 205: Same again, avoid using match statements for only handling errors, use `map_err` or `map_or_else` to return the error from the fn your calling and then map it to the calling fns allowed error type.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -177,27 +181,44 @@ impl DaSequencerNodeService for DaSequencerNode {
...
 			}
 		};
-		if let Err(err) = self.request_tx.send(GrpcRequests::WriteBatch(batch)).await {
-			tracing::error!(error = %err, "Internal grpc request channel closed, no more batch will be processed.");
+
+		// Send it to the internal channel (lock has been dropped by now)
+		if let Err(err) = self.request_tx.send(GrpcRequests::WriteBatch(validated)).await {
+			tracing::error!(
+				"Internal grpc request channel closed, no more batch will be processed: {err}"
+			);
```
0xmovses, line 218:
```suggestion
			tracing::error!(
				"Internal grpc request channel closed, no more batches will be processed: {err}"
			);
```

protocol-units/da-sequencer/node/src/tests/mod.rs
```diff
@@ -1,2 +1,105 @@
...
+use movement_da_sequencer_config::DaSequencerConfig;
+use movement_da_sequencer_proto::BatchWriteRequest;
+use movement_types::transaction::Transaction;
+use serial_test::serial;
+use std::net::TcpListener;
+use std::sync::Arc;
+use tokio::sync::mpsc;
+use tokio::sync::RwLock;
+use tokio::time::Duration;
+use tracing_subscriber;
```
0xmovses, line 18:
```suggestion
use std::{
	net::TcpListener,
	sync::Arc,
};

use tokio::{
	sync::{mpsc, RwLock},
	time::Duration,
};

use ed25519_dalek::Signature;
use serial_test::serial;
use tracing_subscriber;

use movement_da_sequencer_client::{sign_batch, DaSequencerClient};
use movement_da_sequencer_config::DaSequencerConfig;
use movement_da_sequencer_proto::BatchWriteRequest;
use movement_types::transaction::Transaction;

use crate::{
	batch::*,
	run,
	server::{run_server, GrpcRequests},
	tests::{
		mock::{CelestiaMock, StorageMock},
		whitelist::make_test_whitelist,
	},
	whitelist::Whitelist,
};
```

protocol-units/da-sequencer/node/src/tests/whitelist.rs
```diff
@@ -0,0 +1,111 @@
...
+		timestamp: chrono::Utc::now().timestamp_micros() as u64,
+	};
+
+	let validated = validate_batch(raw_batch, &whitelist).expect("Batch should validate");
+	assert_eq!(validated.data.0, txs.0);
+}
+
+#[serial]
+#[tokio::test]
+async fn test_sign_and_validate_batch_fails_with_non_whitelisted_signer() {
```
0xmovses, line 77: 👍

protocol-units/da-sequencer/node/src/tests/whitelist.rs
```diff
@@ -0,0 +1,111 @@
...
+	let _ = tracing_subscriber::fmt()
+		.with_max_level(tracing::Level::INFO)
+		.with_test_writer()
+		.try_init();
+
+	let whitelist = make_test_whitelist(vec![]); // empty whitelist
+
```
0xmovses, line 84: This is just an empty whitelist, can we do a more truer test with some address attempting to sign that is not in the WL file?

0xmovses (changes requested):
Implementation looks good. Some code quality improvements and one suggestion for your test.

0xmovses:
Just had another thought, if your default is relative path `"da-sequencer/whitelist"` I'm assuming that is a raw text file. You might want to use json, to get the benefits of serde_json serialization and fmt catches. What do you think?

protocol-units/da-sequencer/node/src/server.rs
0xmovses: hmm I mean't something like this, I think it's possible?
```rust
let (public_key, signature, bytes) = crate::batch::deserialize_full_node_batch(batch_data)
    .map_or_else(
        |err| {
            tracing::warn!("Invalid batch send: deserialization failed: {err}");
            return Ok(tonic::Response::new(BatchWriteResponse { answer: false }));
        },
        |res| res,
    );
    ```
    no `match` required.

andygolay:
> Just had another thought, if your default is relative path `"da-sequencer/whitelist"` I'm assuming that is a raw text file. You might want to use json, to get the benefits of serde_json serialization and fmt catches. What do you think?

Ah, I meant to ask about this. These unit tests are not actually loading the whitelist file. I was only checking the logic of whether someone can sign if they're on a mock whitelist or not.

I do see a text file `.movement/default_signer_address_whitelist`: 
image

Is that supposed to be used as the whitelist?

Maybe what we should do is merge this, and I'll create an issue for testing loading a whitelist from the setup? I think maybe that relates to the issue @musitdev regarding the setup?

PR #1176, opened 2025-04-02
chore: remove unneeded test

0xmovses:
## Summary
We don't want to repeat this computation, the validation happens internally now in `produce_next_block`. We don't need this method or its unit test. 

If we want to check that `produce_next_block` correctly enforces the max block size constant, another unit test can be added there in storage/mod.rs later.

## Testing
`cargo test -p movement-da-sequencer-node` all tests pass


0xmovses:
Cut branch from wrong one.

PR #1193, opened 2025-04-10
feat[da-sequencer]: remove block height from the Celestia blob

sebtomba:
# Summary
- #1158 

# Changelog

- Remove the block height from the Celestia blob.
- Store an array of block `Id`s in the Celestia blob.
- Remove `SequencerBlockDigest` altogether.
- Resolve all compiler warnings.

# Testing
To run the test, run the cmd:
```
cargo test -p movement-da-sequencer-node 
```
# Outstanding issues

Since we no longer store the block height in the Celestia blob, we need to ensure that all block IDs in the blobs are strictly sequential concerning the block height. The batching algorithm must uphold this invariant.

### DIFF
--- a/protocol-units/da-sequencer/node/src/celestia/blob.rs
+++ b/protocol-units/da-sequencer/node/src/celestia/blob.rs
@@ -4,9 +4,9 @@ use std::slice::Iter;
 
 /// The blob format that is stored in Celestia DA.
 #[derive(Clone, Default, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct CelestiaBlobData(Vec<block::Id>);
+pub struct CelestiaBlob(Vec<block::Id>);
 
-impl CelestiaBlobData {
+impl CelestiaBlob {
 	pub fn iter(&self) -> Iter<'_, block::Id> {
 		self.0.iter()
 	}
@@ -20,7 +20,7 @@ impl CelestiaBlobData {
 	}
 }
 
-impl IntoIterator for CelestiaBlobData {
+impl IntoIterator for CelestiaBlob {
 	type Item = block::Id;
 	type IntoIter = <Vec<block::Id> as IntoIterator>::IntoIter;
 
@@ -29,7 +29,7 @@ impl IntoIterator for CelestiaBlobData {
 	}
 }
 
-impl From<Vec<block::Id>> for CelestiaBlobData {
+impl From<Vec<block::Id>> for CelestiaBlob {
 	fn from(value: Vec<block::Id>) -> Self {
 		Self(value)
 	}
--- a/protocol-units/da-sequencer/node/src/celestia/blob.rs
+++ b/protocol-units/da-sequencer/node/src/celestia/blob.rs
@@ -1,29 +1,36 @@
-use crate::block::{BlockHeight, SequencerBlockDigest};
-use std::slice::Iter;
-
+use movement_types::block;
 use serde::{Deserialize, Serialize};
+use std::slice::Iter;
 
 /// The blob format that is stored in Celestia DA.
 #[derive(Clone, Default, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct CelestiaBlobData {
-	pub digests: Vec<SequencerBlockDigest>,
-}
+pub struct CelestiaBlobData(Vec<block::Id>);
 
 impl CelestiaBlobData {
-	pub fn iter(&self) -> Iter<'_, SequencerBlockDigest> {
-		self.digests.iter()
+	pub fn iter(&self) -> Iter<'_, block::Id> {
+		self.0.iter()
 	}
 
-	pub fn last_block_height(&self) -> Option<BlockHeight> {
-		self.digests.last().map(|b| b.height)
+	pub fn last_block_id(&self) -> Option<block::Id> {
+		self.0.last().copied()
+	}
+
+	pub fn to_vec(self) -> Vec<block::Id> {
+		self.0
 	}
 }
 
 impl IntoIterator for CelestiaBlobData {
-	type Item = SequencerBlockDigest;
-	type IntoIter = <Vec<SequencerBlockDigest> as IntoIterator>::IntoIter;
+	type Item = block::Id;
+	type IntoIter = <Vec<block::Id> as IntoIterator>::IntoIter;
 
 	fn into_iter(self) -> Self::IntoIter {
-		self.digests.into_iter()
+		self.0.into_iter()
+	}
+}
+
+impl From<Vec<block::Id>> for CelestiaBlobData {
+	fn from(value: Vec<block::Id>) -> Self {
+		Self(value)
 	}
 }
--- a/protocol-units/da-sequencer/node/src/celestia/client.rs
+++ b/protocol-units/da-sequencer/node/src/celestia/client.rs
@@ -2,15 +2,13 @@ use super::submit::BlobSubmitter;
 use super::{
 	BlockSource, CelestiaBlobData, CelestiaClientOps, CelestiaHeight, ExternalDaNotification,
 };
-use crate::block::SequencerBlockDigest;
 use crate::error::DaSequencerError;
-
 use celestia_rpc::Client as RpcClient;
 use celestia_types::nmt::Namespace;
+use movement_types::block;
+use std::sync::Arc;
 use tokio::sync::mpsc;
 use url::Url;
-
-use std::sync::Arc;
 
 #[derive(Debug, thiserror::Error)]
 enum Error {
@@ -23,7 +21,7 @@ pub struct CelestiaClient {
 	rpc_client: Arc<RpcClient>,
 	notifier: mpsc::Sender<ExternalDaNotification>,
 	// The sender end of the channel for the background sender task.
-	digest_sender: mpsc::Sender<(SequencerBlockDigest, BlockSource)>,
+	id_sender: mpsc::Sender<(block::Id, BlockSource)>,
 }
 
 impl CelestiaClient {
@@ -44,7 +42,7 @@ impl CelestiaClient {
 			notifier.clone(),
 		);
 		tokio::spawn(blob_submitter.run());
-		Ok(CelestiaClient { rpc_client, notifier, digest_sender })
+		Ok(CelestiaClient { rpc_client, notifier, id_sender: digest_sender })
 	}
 }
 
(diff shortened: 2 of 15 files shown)


0xmovses:
This doesn't fully implement https://github.com/movementlabsxyz/movement/issues/1158 but it's just a step toward it, correct?

protocol-units/da-sequencer/node/src/celestia/blob.rs
```diff
@@ -1,29 +1,36 @@
...
+	type Item = block::Id;
+	type IntoIter = <Vec<block::Id> as IntoIterator>::IntoIter;
 
 	fn into_iter(self) -> Self::IntoIter {
-		self.digests.into_iter()
+		self.0.into_iter()
+	}
+}
+
+impl From<Vec<block::Id>> for CelestiaBlob {
```
mzabaluev, line 32: When `CelestiaBlob` is added with a signature (I'm working on it), this will be more than a thin wrapper over a vector of hashes. But we can think of idiomatic ways to reflect this later.

protocol-units/da-sequencer/node/src/celestia/client.rs
```diff
@@ -1,29 +1,25 @@
...
-	BlockSource, CelestiaBlobData, CelestiaClientOps, CelestiaHeight, ExternalDaNotification,
-};
-use crate::block::SequencerBlockDigest;
+use super::{BlockSource, CelestiaBlob, CelestiaClientOps, CelestiaHeight, ExternalDaNotification};
 use crate::error::DaSequencerError;
-
 use celestia_rpc::Client as RpcClient;
 use celestia_types::nmt::Namespace;
+use movement_types::block;
+use std::sync::Arc;
```
mzabaluev, line 7: Nit: I usually put `std` imports into their own group so they stand out better. But if this is how rustfmt automatically rearranges it, OK.

sebtomba:
@ 0xmovses There isn't much left to do regarding #1158. This is it.
Do you agree @musitdev?

protocol-units/da-sequencer/node/src/celestia/client.rs
```diff
@@ -1,29 +1,25 @@
...
 	#[error("Celestia RPC error: {}", .0)]
 	Rpc(#[from] celestia_rpc::Error),
 }
 
 #[derive(Clone)]
 pub struct CelestiaClient {
-	rpc_client: Arc<RpcClient>,
-	notifier: mpsc::Sender<ExternalDaNotification>,
+	_rpc_client: Arc<RpcClient>,
+	_notifier: mpsc::Sender<ExternalDaNotification>,
```
mzabaluev, line 20: These will come useful when `get_blob_at_height` is implemented (in a separate upcoming PR). But I can deal with it.

0xmovses:
Ok, we will also need some e2e tests to test the algo I assume.

protocol-units/da-sequencer/node/src/celestia/client.rs
```diff
@@ -1,29 +1,25 @@
 use super::submit::BlobSubmitter;
-use super::{
-	BlockSource, CelestiaBlobData, CelestiaClientOps, CelestiaHeight, ExternalDaNotification,
-};
-use crate::block::SequencerBlockDigest;
+use super::{BlockSource, CelestiaBlob, CelestiaClientOps, CelestiaHeight, ExternalDaNotification};
```
0xmovses, line 2: Seems like formatting isn't being applied as this is all on one line.

protocol-units/da-sequencer/node/src/celestia/mod.rs
```diff
@@ -241,58 +261,74 @@ impl<B: BlockOps, C: CelestiaClientOps> DaSequencerExternalDa for CelestiaExtern
...
+	impl Id for Block {
+		fn id(&self) -> block::Id {
+			self.id()
+		}
+	}
+
+	impl Id for CelestiaBlob {
+		fn id(&self) -> block::Id {
+			self.iter().next().copied().unwrap()
+		}
```
mzabaluev, line 305: The generalization seems fishy, but this is test code.

util/movement-types/src/block.rs
```diff
@@ -14,30 +14,32 @@ pub enum BlockError {
 	BlockFull,
 }
 
+pub const ID_SIZE: usize = 32;
```
mzabaluev, line 17: Stylistically I'd prefer this to be an associated const in `Id`.

protocol-units/da-sequencer/node/src/celestia/submit.rs
```diff
@@ -85,16 +85,16 @@ impl BlobSubmitter {
...
-						next = self.digest_receiver.recv(), if total_data_size + SequencerBlockDigest::DIGEST_SIZE <= MAX_CELESTIA_BLOB_SIZE => {
+						next = self.id_receiver.recv(), if total_data_size + block::ID_SIZE <= MAX_CELESTIA_BLOB_SIZE => {
 							match next {
 								None => break,
-								Some((digest, BlockSource::Input)) => {
-									total_data_size += digest.id.as_bytes().len();
-									buffered_digests.push(digest);
+								Some((id, BlockSource::Input)) => {
+									total_data_size += block::ID_SIZE;
+									buffered_ids.push(id);
```
0xmovses, line 93: I suppose this is makes this part easier too, `ID_SIZE` will be constant.

protocol-units/da-sequencer/node/src/server.rs
```diff
@@ -180,10 +180,18 @@ impl DaSequencerNodeService for DaSequencerNode {
 		Ok(tonic::Response::new(Box::pin(output) as Self::StreamReadFromHeightStream))
 	}
 
+	/// Read one block at a specified height.
+	async fn read_at_height(
+		&self,
+		_request: tonic::Request<ReadAtHeightRequest>,
+	) -> std::result::Result<tonic::Response<ReadAtHeightResponse>, tonic::Status> {
+		Err(tonic::Status::unimplemented(""))
```
0xmovses, line 188: Just do 
```rust
Result<tonic::Response<ReadAtHeightResponse>, tonic::Status>
```
Import the result type up top.

util/signing/providers/hashicorp-vault/src/hsm/mod.rs
```diff
@@ -86,7 +87,7 @@ where
 			&self.client,
 			self.mount_name.as_str(),
 			self.key_name.replace("/", "_").as_str(),
-			base64::encode(message).as_str(),
+			base64::engine::general_purpose::STANDARD.encode(message).as_str(),
```
mzabaluev, line 90: Could use the re-exports in `base64::prelude`.
This is to remove deprecation warnings, right? The top-level `base64::encode` uses this engine anyway.

protocol-units/da-sequencer/node/src/storage/mod.rs
```diff
@@ -45,13 +44,13 @@ impl TxCompositeKey {
...
-	fn decode(bytes: &[u8]) -> Result<Self, DaSequencerError> {
+	fn _decode(bytes: &[u8]) -> Result<Self, DaSequencerError> {
 		if bytes.len() != 44 {
 			return Err(DaSequencerError::StorageFormat("Invalid composite key length".into()));
 		}
 
-		let timestamp = u64::from_be_bytes(bytes[0..8].try_into().unwrap());
-		let index_in_batch = u32::from_be_bytes(bytes[8..12].try_into().unwrap());
+		let timestamp = u64::from_be_bytes(bytes[0..8].try_into()?);
+		let index_in_batch = u32::from_be_bytes(bytes[8..12].try_into()?);
```
0xmovses, line 53: Ah yes, not show how these unwrap calls got merged, thanks.

0xmovses (changes requested):
Yeah, I think this a good idea.

protocol-units/da-sequencer/node/src/celestia/blob.rs
sebtomba: Either way, once we change the `struct`, we'll have to update all the code in the crate.

protocol-units/da-sequencer/node/src/celestia/client.rs
sebtomba: At some point, we need to create a document with coding conventions.

protocol-units/da-sequencer/node/src/celestia/client.rs
sebtomba: It's just temporary to get rid of the compiler warnings.

protocol-units/da-sequencer/node/src/celestia/mod.rs
sebtomba: Haha, right, this is nonsense. But it's not used in any tests. I needed `Id` for `Blocks` to index them by height and id. There is no feature to get a blob by ID.

util/signing/providers/hashicorp-vault/src/hsm/mod.rs
sebtomba: Yes, this is done to remove the deprecation warning. We should provide a util function somewhere to hide the ugliness.

protocol-units/da-sequencer/node/src/celestia/client.rs
0xmovses: In general for feature branches we are relaxed on them, only when merged feature branch into main do we want to reduce the warnings. In some ways we want to to see the compiler warnings when working on the feature branch, as they will become todos before we merge into main. Otherwise, we can miss them. But this is fine for now.

protocol-units/da-sequencer/node/src/storage/mod.rs
sebtomba: It has been delivered courtesy of RustRover. :)

protocol-units/da-sequencer/node/src/celestia/client.rs
0xmovses: A style guide such as 
https://github.com/paritytech/substrate/blob/master/docs/STYLE_GUIDE.md

protocol-units/da-sequencer/node/src/celestia/client.rs
sebtomba: The noise generated by warnings in this feature branch was massive. I had difficulty locating compile errors among them and had to scroll back and forth to identify them.

protocol-units/da-sequencer/node/src/celestia/client.rs
sebtomba: Yes, and further restrictions. Blank lines between `use` declarations prevent `rustfmt` from sorting. We should keep them as tidy as possible (sorting + merging).

protocol-units/da-sequencer/node/src/celestia/client.rs
sebtomba: Surprise! It's 98 characters long and adequately formatted.

musitdev (approved):

protocol-units/da-sequencer/node/src/storage/mod.rs
musitdev: They have been merged because they are tested just before, so panic won't ever happen.

Issue #1201, opened 2025-04-13
All ransaction are expired for some accounts


musitdev:
In the Bardock and main net logs, some Tx are discarded after execution because of the error TRANSACTION_EXPIRED. 
It's related to the account because, after some tests from partners, all Tx expires for the same account.
The Tx execution is normal around 6s so it's not because the Tx takes too much time to reach the executor.

Some account that has Tx Expired:

PR #1202, opened 2025-04-13
Bug: Investigate Tx Expired

musitdev:
# Summary

Related to this issue: #1201 

Investigate the issue.

# Changelog

* Add some logs to test on the follower node.

# Testing

# Outstanding issues

### DIFF
--- a/networks/movement/movement-client/src/bin/test_expirity.rs
+++ b/networks/movement/movement-client/src/bin/test_expirity.rs
@@ -1,0 +1,151 @@
+use anyhow::Context;
+use bcs::to_bytes;
+use movement_client::{
+	coin_client::CoinClient,
+	move_types::{
+		identifier::Identifier,
+		language_storage::{ModuleId, TypeTag},
+	},
+	rest_client::{Client, FaucetClient},
+	transaction_builder::TransactionBuilder,
+	types::transaction::{EntryFunction, SignedTransaction, TransactionPayload},
+	types::{account_address::AccountAddress, chain_id::ChainId, LocalAccount},
+};
+use once_cell::sync::Lazy;
+use std::str::FromStr;
+use std::time::SystemTime;
+use std::time::UNIX_EPOCH;
+use url::Url;
+
+static SUZUKA_CONFIG: Lazy<movement_config::Config> = Lazy::new(|| {
+	let dot_movement = dot_movement::DotMovement::try_from_env().unwrap();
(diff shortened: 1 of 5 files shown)


andygolay:
Should this be in draft if it's an investigation?

networks/movement/movement-client/src/bin/test_expirity.rs
```diff
@@ -0,0 +1,151 @@
+use anyhow::Context;
```
l-monninger, line 1: `test_expirity` $\rightarrow$ `test_expiry` or `test_expiration`

networks/movement/movement-client/src/bin/test_expirity.rs
```diff
@@ -0,0 +1,151 @@
...
+#[tokio::main]
+async fn main() -> Result<(), anyhow::Error> {
+	use tracing_subscriber::EnvFilter;
+	tracing_subscriber::fmt()
+		.with_env_filter(
+			EnvFilter::try_from_default_env().unwrap_or_else(|_| EnvFilter::new("info")),
+		)
+		.init();
+	let client_url = NODE_URL.clone();
+	let client_url = Url::from_str("https://testnet.bardock.movementnetwork.xyz/v1").unwrap();
```
l-monninger, line 74: Prefer not to unwrap in case people later copy code.

networks/movement/movement-client/src/bin/test_expirity.rs
```diff
@@ -0,0 +1,151 @@
...
+		"Start Addresses ===\nAlice: {}\nBob: {}",
+		alice.address().to_hex_literal(),
+		bob.address().to_hex_literal()
+	);
+
+	// Create the accounts on chain, but only fund Alice.
+	faucet_client.fund(alice.address(), 100_000_000_000).await?;
+	faucet_client.create_account(bob.address()).await?;
+
+	let transaction =
```
l-monninger, line 103: What is the plan here?

l-monninger (changes requested):

networks/movement/movement-client/src/bin/test_expirity.rs
musitdev: I was a test to investigate the issue we have on mainnet with Tx expiration.
I issue come from the fact that wallet use a very short time for Tx duration (20s) and we can't execute Tx in less than 6/10 s currently. So if there's some delay in the Tx sending, the Tx is expired when executing.

PR #1219, opened 2025-05-04
[DO NOT MERGE] feat: configurable `rev` in framework release compiler

andygolay:
# Summary

# Changelog

# Testing

# Outstanding issues


andygolay:
Closing as this approach is not optimal; it was useful for quickly testing revs over the past couple days, but there are cleaner ways eg by simply modifying `Compiler::movement`, or perhaps using a `config.json` instead of an env var.

PR #1228, opened 2025-05-13
feat[pre-l1-merge]: activate Aptos feature flags

sebtomba:
# Summary
- **RFCs**: $\emptyset$.

Activate previously inactive Aptos feature flags as part of the `pre-l1-merge` upgrade.
See https://github.com/movementlabsxyz/movement-migration/issues/30#issuecomment-2862738427

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/execution/maptos/framework/releases/pre-l1-merge/src/cached.rs
+++ b/protocol-units/execution/maptos/framework/releases/pre-l1-merge/src/cached.rs
@@ -57,6 +57,18 @@ pub mod full {
 		// Note: when testing into the future, you may have to use a different revision of [aptos_types] in this crate's Cargo.toml
 		// Or, I suppose you can keep and GOVERNED_GAS_POOL feature flag and a GOVERNED_GAS_POOL_V2 feature flag and just make sure you're disabling the former and enabling the latter. Thereafter, it won't matter what happens to the GOVERNED_GAS_POOL feature flag, i.e., it can be replaced.
 		aptos_feature_flags.push(AptosFeatureFlag::GOVERNED_GAS_POOL);
+		aptos_feature_flags.push(AptosFeatureFlag::PERIODICAL_REWARD_RATE_DECREASE);
+		aptos_feature_flags.push(AptosFeatureFlag::PARTIAL_GOVERNANCE_VOTING);
+		aptos_feature_flags.push(AptosFeatureFlag::DELEGATION_POOL_PARTIAL_GOVERNANCE_VOTING);
+		aptos_feature_flags.push(AptosFeatureFlag::VM_BINARY_FORMAT_V7);
+
+		aptos_feature_flags.push(AptosFeatureFlag::ALLOW_SERIALIZED_SCRIPT_ARGS);
+		aptos_feature_flags.push(AptosFeatureFlag::ENABLE_ENUM_TYPES);
+		aptos_feature_flags.push(AptosFeatureFlag::FEDERATED_KEYLESS);
+		aptos_feature_flags.push(AptosFeatureFlag::TRANSACTION_SIMULATION_ENHANCEMENT);
+		aptos_feature_flags.push(AptosFeatureFlag::COLLECTION_OWNER);
+		aptos_feature_flags.push(AptosFeatureFlag::NATIVE_MEMORY_OPERATIONS);
+		aptos_feature_flags.push(AptosFeatureFlag::ACCOUNT_ABSTRACTION);
 
 		Features {
 			enabled: aptos_feature_flags.into_iter().map(FeatureFlag::from).collect(),
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11,7 +11,7 @@ checksum = "fe438c63458706e03479442743baae6c88256498e6431708f6dfc520a26515d3"
 [[package]]
 name = "abstract-domain-derive"
 version = "0.1.0"
-source = "git+https://github.com/movementlabsxyz/aptos-core?rev=1d1cdbbd7fabb80dcb95ba5e23213faa072fab67#1d1cdbbd7fabb80dcb95ba5e23213faa072fab67"
+source = "git+https://github.com/movementlabsxyz/aptos-core?rev=4883c47d74bc386e07ba271120c48aa2e3a95c12#4883c47d74bc386e07ba271120c48aa2e3a95c12"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -785,7 +785,7 @@ checksum = "4c95c10ba0b00a02636238b814946408b1322d5ac4760326e6fb8ec956d85775"
 [[package]]
 name = "aptos-abstract-gas-usage"
 version = "0.1.0"
-source = "git+https://github.com/movementlabsxyz/aptos-core?rev=1d1cdbbd7fabb80dcb95ba5e23213faa072fab67#1d1cdbbd7fabb80dcb95ba5e23213faa072fab67"
+source = "git+https://github.com/movementlabsxyz/aptos-core?rev=4883c47d74bc386e07ba271120c48aa2e3a95c12#4883c47d74bc386e07ba271120c48aa2e3a95c12"
 dependencies = [
  "anyhow",
  "aptos-gas-algebra",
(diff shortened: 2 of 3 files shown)


protocol-units/execution/maptos/framework/releases/pre-l1-merge/src/cached.rs
```diff
@@ -57,6 +57,18 @@ pub mod full {
 		// Note: when testing into the future, you may have to use a different revision of [aptos_types] in this crate's Cargo.toml
 		// Or, I suppose you can keep and GOVERNED_GAS_POOL feature flag and a GOVERNED_GAS_POOL_V2 feature flag and just make sure you're disabling the former and enabling the latter. Thereafter, it won't matter what happens to the GOVERNED_GAS_POOL feature flag, i.e., it can be replaced.
 		aptos_feature_flags.push(AptosFeatureFlag::GOVERNED_GAS_POOL);
```
l-monninger, line 59: @andygolay Does `main` and thus this branch already do the removal of the `GOVERNED_GAS_POOL` and then re-enable under a different feature number?

l-monninger, line 59: @sebtomba Can you provide a comment in the code linking to an explanation for these feature flags?

l-monninger (changes requested):

protocol-units/execution/maptos/framework/releases/pre-l1-merge/src/cached.rs
sebtomba: I assume it's fine to leave the flag `73` enabled because
```
    /// Enabled on mainnet, cannot be disabled.
    _USE_COMPATIBILITY_CHECKER_V2 = 73,
```
But we must enable the feature `GOVERNED_GAS_POOL` under the new number.

l-monninger: I'll need to double check before I approve here.

andygolay: > @andygolay Does `main` and thus this branch already do the removal of the `GOVERNED_GAS_POOL` and then re-enable under a different feature number?

No it doesn't do it already. I think @sebtomba is adding it to the pre-l1-merge.

l-monninger: Okay, so I believe what needs to happen here in the script is this: https://github.com/movementlabsxyz/movement-migration/issues/31#issuecomment-2811563483

sebtomba: Please review the comment I added (yesterday).

sebtomba: What exactly is the "removal of the `GOVERNED_GAS_POOL`"?
Is it more than just toggling the flag?

sebtomba:
The `test-migrate-biarritz-rc1-to-pre-l1-merge` is passing for most of the new flags. However, activation of the feature flag `PERIODICAL_REWARD_RATE_DECREASE` causes failure
```
2025-05-15T13:36:41.708287Z INFO aptos_framework_set_feature_flags_release: failed to submit feature flag proposal: Unknown(Transaction committed on chain, but failed execution: ExecutionFailure { location:
0000000000000000000000000000000000000000000000000000000000000001:: staking_config, function: 6, code_offset: 5 })
Error: proposing release failed with error: failed to submit feature flag proposal: Unknown (Transaction committed on chain, but failed execution: ExecutionFailure { location:
0000000000000000000000000000000000000000000000000000000000000001:: staking_config, function: 6, code_offset: 5 })

Caused by:
failed to submit feature flag proposal: Unknown (Transaction committed on chain, but failed execution: ExecutionFailure { location: 0000000000000000000000000000000000000000000000000000000000000001:: staking_config, function: 6, code_offset: 5 })
``` 
I'm looking into this problem now.

sebtomba:
In this upgrade, we can't activate the feature flag `PERIODICAL_REWARD_RATE_DECREASE` because it requires the resource `StakingRewardsConfig` (staking reward configuration), which can only be initialized as part of the governance proposal process. See https://github.com/movementlabsxyz/aptos-core/blob/movement/aptos-move/framework/aptos-framework/sources/configs/staking_config.move#L145

andygolay:
@sebtomba I'm getting the `stack too deep` Solc error. 
image

when I run
```
CELESTIA_LOG_LEVEL=FATAL nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.eth-local.celestia-local.test-migrate-biarritz-rc1-to-pre-l1-merge --keep-tui"
```
locally. Are you getting it on your machine?

sebtomba:
@andygolay I can reproduce this problem when I delete the `.movement` folder.
However, when I run `target/debug/movement-full-node-setup` manually, this error doesn't occur.

0xmovses:
Testing now.

sebtomba:
Since this PR doesn't introduce the problem of building MCR with solc, can we review, approve, and merge it?

0xmovses (approved):

andygolay (approved):

0xmovses:
Awaiting approval from @l-monninger to merge this.

l-monninger:
@sebtomba @andygolay @0xmovses 

It's still not clear to me why we don't have to do something like this?: https://github.com/movementlabsxyz/movement-migration/issues/31#issuecomment-2811563483

I don't think this will have the expected behavior when you are able to test e2e. 

Also, again, just try to cut out some modules to get the `solc` to work, example: https://github.com/movementlabsxyz/ffs/issues/31

andygolay:
> @sebtomba @andygolay @0xmovses
> 
> It's still not clear to me why we don't have to do something like this?: [movementlabsxyz/movement-migration#31 (comment)](https://github.com/movementlabsxyz/movement-migration/issues/31#issuecomment-2811563483)
> 
> I don't think this will have the expected behavior when you are able to test e2e.
> 
> Also, again, just try to cut out some modules to get the `solc` to work, example: [movementlabsxyz/ffs#31](https://github.com/movementlabsxyz/ffs/issues/31)


- I was thinking that by setting the new feature flag set in the upgrade, it already does all of this:
>Upgrade to an unmerged version of the framework containing GGP using a different feature flag number.
> Make a features upgrade to the new GGP feature flag number, turn off the old one.
> Upgrade to a merged version of the framework.
 
- However, I think you're right, we don't know unless we test it, and perhaps the memory for feature flags does need to be reset in a multi-step process like this.
 
- I'll try again on the modules tonight.

l-monninger:
@andygolay 

> I was thinking that by setting the new feature flag set in the upgrade, it already does all of this:

The problem is you're trying to enable both. Yes, if you take the simple additive approach (just turning on the new GGP feature flag), there's a chance it works.

However, there a couple potential pitfalls:
1. If you turn Compatibility Checker v2 and upgrade the framework, perhaps some branch gets tripped that in the upgrade process generally that causes GGP collection to start failing. When this happens, you're in a very tricky spot because you may have screwed up your epilogue and not have transaction-based means to 
2. More specific to the additive approach, if you issue the transaction to upgrade the framework first, you will have feature flags that indicate `COMPATABILITY_CHECKER_V2 = true` and `GGP = false`--which would certainly cause the above if the changes take effect. It won't be until a later transaction wherein you add the GGP feature flag. Until you call `reconfigure` this shouldn't take effect. But, there's a good chance of messing this up.

andygolay:
@sebtomba want to merge `main` into this branch so we can test it?

sebtomba:
@andygolay Merged `main` into this branch. Both tests are passing on my machine.

0xmovses:
This seems good to go now @l-monninger re. your concerns, until we see it happen in the migration repo when we do the e2e with the runners, we can remain optimistic I suppose?

andygolay:
> @andygolay Merged `main` into this branch. Both tests are passing on my machine.

Yes, passing for me as well 👍

l-monninger:
> This seems good to go now @l-monninger re. your concerns, until we see it happen in the migration repo when we do the e2e with the runners, we can remain optimistic I suppose?

My concerns are just based on the upgrade running by itself. So, if it's running now, it should be good to go.

l-monninger (approved):

PR #1229, opened 2025-05-14
vanity move address

Primata:
# Summary
- **RFCs**: *scripts*.

Adds util to mine move vanity addresses.
Requires commenting out flags in .cargo/config.toml
```
  "-C",
  "link-arg=-fuse-ld=lld",
```
# Changelog

util/vanity/*

# Testing

# Outstanding issues


util/vanity/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "vanity"
+version = "0.2.1"
+edition = "2021"
+
+[dependencies]
+aptos-sdk = { workspace = true }
+rand = { workspace = true }
+rayon = "1.7"
```
l-monninger, line 9: Use workspace deps for all.

util/vanity/README.md
```diff
@@ -0,0 +1,20 @@
...
+cd movement
+# Run it directly
+cargo run -p vanity --release -- move --starts-pattern <STARTS_PATTERN> --ends-pattern <ENDS_PATTERN>
+
+cd util/vanity
+# Add it to your path
+cargo install --path .
+```
+
+*Currently requires ignoring all instances of "-C", "link-arg=-fuse-ld=lld" flags in .cargo/config.toml*
```
l-monninger, line 20: Make it work with `lld`.

l-monninger (changes requested):

util/vanity/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+                None => vec![],
+            };
+
+            let account = mine_move_address(
+                &starts_pattern_bytes,
+                &ends_pattern_bytes,
+                num_cpus::get(),
+            );
+
+            println!("Found Move address: {}", account.address());
```
0xmovses, line 28: remove the println use tracing::info

util/vanity/src/main.rs
```diff
@@ -0,0 +1,33 @@
...
+            println!("Private key (hex): {}", hex::encode(account.private_key().to_bytes()));
```
0xmovses, line 29: Ditto

0xmovses (commented):
This looks cool, can you explain a bit the intention of this tool. Why would I want to use it? I think it's worth adding a few unit tests to demonstrate its usage.

0xmovses:
It's so that I can add prefix/suffix patterns to an address right?

util/vanity/src/main.rs
l-monninger: When you’re writing output for CLIs that’s not diagnostic logging you’re supposed to use `println`, i.e., to directly control `stdout` without additional formatting. This facilitates things like piping.

Leave as is.

0xmovses (commented):
Can you write some unit tests for this? Then if all is good happy to approve, I support freedom of expression.

Primata:
Glad you do 🏴‍☠️ 🌈

0xmovses:
Thanks for the test, will run after programme review.

0xmovses (approved):
LGayTM

andygolay (changes requested):
Looks like a Cargo Format check is failing and I got a few failing tests in the `vanity` package:

```
   Compiling vanity v0.2.1 (/Users/andygmove/Downloads/repos/movement/util/vanity)
warning: unused import: `tracing`
  --> util/vanity/src/miner.rs:10:5
   |
10 | use tracing;
   |     ^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `predicates::prelude`
  --> util/vanity/src/cli.rs:70:6
   |
70 |     use predicates::prelude::*;
   |         ^^^^^^^^^^^^^^^^^^^

warning: `vanity` (bin "vanity" test) generated 2 warnings (run `cargo fix --bin "vanity" --tests` to apply 1 suggestion)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 1m 22s
     Running unittests src/main.rs (target/debug/deps/vanity-c8535288652eb800)

running 5 tests
test cli::tests::test_valid_pattern ... ok
test cli::tests::test_cli_runs_with_ends ... FAILED
test cli::tests::test_cli_runs_with_starts ... FAILED
test cli::tests::test_cli_runs_with_starts_and_ends ... FAILED
test cli::tests::test_cli_parsing_starts_pattern ... ok
```

A couple Rust warnings in there to clean up.

Otherwise LGodlyTM. I tested it locally and it produced a valid keypair with my desired vanity address.

andygolay (approved):
LGTM, I hadn't ran `cargo build -p vanity` first before running the tests. I did it just now and all passed.

I ran a `cargo fmt` to fix the Cargo Format CI failure and committed; I think this is good to merge.

util/vanity/src/cli.rs
```diff
@@ -1,186 +1,184 @@
...
-        &self.0
-    }
+	/// Return the raw hex string without trying to decode it.
+	pub fn as_str(&self) -> &str {
+		&self.0
+	}
 }
 
 /// Pattern parsing errors.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, thiserror::Error)]
```
l-monninger, line 24: Do you need `Copy, PartialEq, Eq`? It's pretty rare that you would with error enums.

util/vanity/src/cli.rs
```diff
@@ -0,0 +1,184 @@
...
+#[command(name = "vanity", about = "Vanity is a fast vanity address miner.")]
+pub struct Cli {
+	#[command(subcommand)]
+	pub command: Vanity,
+}
+
+/// Supported CLI subcommands.
+#[derive(Debug, Subcommand)]
+pub enum Vanity {
+	Move {
```
l-monninger, line 57: You should add comments on the enum variants so that they are reported in `clap --help`. 

Also, generally, I would suggest a file-and-dir-based organization, e.g., [mcr-protocol-client](https://github.com/movementlabsxyz/ffs/tree/main/protocol/mcr/cli/client), for maintainability.

util/vanity/src/cli.rs
```diff
@@ -0,0 +1,184 @@
...
+		#[clap(long)]
+		starts: Option<String>,
```
l-monninger, line 59: You probably want to invalidate the case where both `--starts` and `--ends` are `None`. You can either do this in the implementation or use [`ArgGroup`](https://stackoverflow.com/questions/76315540/how-do-i-require-one-of-the-two-clap-options).

util/vanity/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "vanity"
```
l-monninger, line 2: I would call this something like `movement-vanity` to be a bit more specific.

util/vanity/src/main.rs
```diff
@@ -0,0 +1,64 @@
+mod cli;
```
l-monninger, line 1: Since it costs you almost nothing and provides something reusable, I would pull `cli` and `miner` into a `lib.rs`, then `use vanity::{cli, miner}`.

l-monninger (changes requested):

util/vanity/Cargo.toml
Primata: kept as vanity for simplicity

l-monninger: Let's say we want to make this installable with `cargo install` s.t. people have an easy time using. We would then need to register with crates.io, in which case `vanity` is likely either already taken or not specific enough to be useful.

util/vanity/src/lib.rs
```diff
@@ -0,0 +1,286 @@
...
+        /// Return the raw hex string without trying to decode it.
+        pub fn as_str(&self) -> &str {
+            &self.0
+        }
+    }
+
+    /// Pattern parsing errors.
+    #[derive(Clone, Debug, PartialEq, Eq, thiserror::Error)]
+    pub enum PatternError {
+        #[error("the pattern's length exceeds 39 characters or the pattern is empty")]
```
l-monninger, line 124: If this is the case, why not make it `pub struct Pattern(Box<[u8;39]>)`. That's a fixed-size, heap allocated byte-array with buffer that is always 39 characters and never any more.

util/vanity/src/cli.rs
l-monninger: You're still not invalidating this case. Is there a reason you don't want to do this?

PR #1236, opened 2025-05-20
feat: `format-module-bytes` utility

ganymedio:
# Summary

- Add the `format-module-bytes` utility to provide the serialized args for calling [`code::publish_package_txn`](https://explorer.movementlabs.xyz/account/0x0000000000000000000000000000000000000000000000000000000000000001/modules/run/code/publish_package_txn?network=mainnet) on Movement explorer.
- Useful where publishing via explorer is needed for some reason (eg an org using a multisig provider that doesn't support CLI operations).

# Changelog

- Adds the package `format-module-bytes` and tests.

# Usage

See the [README](https://github.com/movementlabsxyz/movement/blob/andygolay/format-module-bytes/util/format-module-bytes/README.md) for instructions with the included example `hello.move` package.

The main steps are:

1. From `util/format-module-bytes`, `movement init` or copy in a `.movement/config.yaml` into the `util/format-module-bytes` dir to set up the account you intend to use for publishing. The `default` profile is the one used in this util.
2. Add your modules under `sources`. 
3. `cargo run -p format-module-bytes`
4. Publish via explorer with the same address you used in your `.movement/config.yaml`. (Ensure the account is funded.)

# Testing

Unit tests: `cargo test -p format-module-bytes`

E2E testing is best done manually; you should be able to publish the package with the serialized inputs and see a success result as follows:
image


Cargo.toml
```diff
@@ -30,6 +30,7 @@ members = [
   "util/tracing",
   "util/syncador",
   "util/collections",
+  "util/publish-via-explorer",
```
l-monninger, line 33: So, this is really just a way to log the byte payloads that you need for filling out that form to the shell, right?

l-monninger: If this is the case, I would say something like `format-module-bytes` is better and then you could have a subcommand `explorer`, i.e., `format-module-bytes explorer ...`, or abbreviated `fmt-mb explorer ...` .

andygolay: @l-monninger yes, correct.

Hmm yeah that would be better, since the payloads could be used for the explorer or with Movement CLI, so `format-module-bytes` does more accurately capture the essence of this utility. I don't see it really needing to have a subcommand or be a CLI though. If you just run `cargo run -p format-module-bytes` it logs the bytes, which is all that is needed, as far as I can tell. Do you see other functionality needed that would make subcommands or a CLI necessary?

l-monninger: To me, that would just make sense if you had multiple formats we were targeting.

andygolay: I'm just targeting `vector(u8)` since that's what one partner needed for publishing via the explorer, and @shayansanjideh mentioned other teams wanting help with publishing via multisig. That was the main motivation for creating this utility.

l-monninger (approved):
Seems to be working correctly for me. As this isn't on the critical path, I think it's well enough tested.

Primata:
Closing as stale — no activity since 2025-05-28, and the codebase has moved
substantially since (the explorer link in the description points at the old
`movementlabsxyz` org).

Not a judgement on the utility itself: if `format-module-bytes` is still wanted
for the `code::publish_package_txn` flow, it'll be quicker to reopen or re-raise
against current `main` than to rebase fifteen months of drift. Happy to review
it then.

PR #1242, opened 2025-05-27
fix: duplicate submodule path

andygolay:
# Summary

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,6 @@
 [submodule "protocol-units/dispute/lib/forge-std"]
 	path = protocol-units/settlement/mcr/contracts/lib/forge-std
     url = https://github.com/foundry-rs/forge-std
-[submodule "protocol-units/dispute/lib/openzeppelin-contracts"]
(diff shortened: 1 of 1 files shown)


l-monninger:
> I don't know whether deleting the paths labeled "dispute" might have some unintended side effects.

This should be fine. It was from back when we were implementing "Monza," if you can remember.

l-monninger (approved):
Assuming CI passes, LGTM.

andygolay:
> Assuming CI passes, LGTM.



> > I don't know whether deleting the paths labeled "dispute" might have some unintended side effects.
> 
> This should be fine. It was from back when we were implementing "Monza," if you can remember.

Yes, I recall the possibility of someone disputing a settlement outcome using eg a fraud proof.

andygolay:
> Assuming CI passes, LGTM.

The only failure is the unit tests: 
image
but also I see that in recent commits to `main`, so I think this should be okay as it doesn't introduce those test failures.

PR #1243, opened 2025-05-27
fix: stack too deep solc error

andygolay:
# Summary
- Addresses https://github.com/movementlabsxyz/movement-migration/issues/76

# Changelog

- Remove several unneeded scripts and contracts from `protocol-units/settlement/mcr/contracts/`

# Testing

- Stack too deep error should not appear in `setup` when running
```
nix develop --extra-experimental-features nix-command --extra-experimental-features flakes --command bash  -c "just movement-full-node native build.setup.eth-local.celestia-local --keep-tui"
```

# Outstanding issues
- After removing the Safe contract and some scripts, I get past the `stack too deep` error and `setup` succeeds. 

image

But now in `celestia-bridge` I get
```
Error: node: failed to start: failed to get node info: rpc error: code = Unavailable desc = connection error: desc = "error reading server preface: http2: frame too large"
```

image

If anyone can run the test locally to see whether they can repro the `frame too large` error, it would be helpful, to check that it's not just a local issue for me.


musitdev (approved):
I've tested, it solves the solc issue to it's fine.

0xmovses (approved):
`just movement-full-node native build.setup.eth-local.celestia-local` ran successfully :shipit:

PR #1249, opened 2025-06-12
Bug, warn on batch rejected by the da sequencer.

musitdev:
# Summary

If the da-sequencer rejects the full node batch, the full node crashes.
Remove the panic to avoid the node to  crash and only log a warning.

# Changelog

# Testing

# Outstanding issues

### DIFF
--- a/protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
+++ b/protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
@@ -177,8 +177,7 @@ impl TransactionPipe {
 								Ok(Ok(response)) => {
 									debug!("After sent batch.");
 									if !response.answer {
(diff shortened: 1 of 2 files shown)


protocol-units/execution/maptos/opt-executor/src/background/transaction_pipe.rs
```diff
@@ -177,8 +177,7 @@ impl TransactionPipe {
 								Ok(Ok(response)) => {
 									debug!("After sent batch.");
 									if !response.answer {
-										tracing::error!("DA Sequencer reject batch, can't send batch, exit process");
-										return Err(Error::InternalError(format!("DA Sequencer reject batch, can't send batch, exit process")));
+										tracing::warn!("DA Sequencer reject batch, can't send the batch.");
```
l-monninger, line 180: Should we maybe add some logging that gives information about the batch and rejection reason?

andygolay (approved):

PR #1275, opened 2025-07-12
expeirmetn: migrate processors from v1 to v2.

larryl3u:
# Summary

# Changelog

# Testing

# Outstanding issues


musitdev:
I don't think the indexer V1 of the current mainnet will be upgraded to v2 until the migration. There's some change in the indexer code to manage the data of movement Tx that are different from the Aptos one. I don't think we'll update indexer v2 to manage these changes because we want to avoid code divergence from root repo and they won't be need after the migration.
So it's better if you create a new  crate to be able to have both indexers in main branch after the merge.

0xmovses:
Should this experimental PR be closed?

Issue #1258, opened 2025-06-23
Migrate The Indexer


0xmovses:
## Description

Our current indexer Is some commits behind the up stream https://github.com/movementlabsxyz/aptos-indexer-processors

We need the latest indexer to be up and running post migration, so that we can continue providing a reliable indexer service that can both Index blocks post migrate _and_ pre migration DBs. 

## Ideal UX 
Would be to have a dapp point their application to one indexer endpoint, do their queries there and simple retrieve the data they need. 

### Problem 
- It's possible that Indexer GraphQL schemas have changed, and therefor, their associated queries. But, this may be a non-problem, if schemas _have not_ changes. Need to verify the state of schema diff (if any)
   - If no change, then we would just need to _re-index_ all the historical Transaction Data, in the latest indexer, store that and serve that. Not too bad.
   - If change, then we need to write additional logic to either migrate the data OR update the schemas, so that one could specific in the query, if the data being queried is pre or post merge block, then in the backend fetch the data from the relavent DB 

## Acceptance Criteria 
Have the new Indexer Running on a devnet, do some txs on movement-network current db. Index that, store it, query it. Do some txs on an aptos-core latest full node. Index that, store it, query it. 

I am keeping the acceptance criteria high level and trying to keep the goal clear. Likely there are sub isues and several steps involved to do this. Feel free to add comments and/or sub-issues to this.

andygolay:
@musitdev for Aptos in the [quickstart guide](https://aptos.dev/en/build/indexer/indexer-sdk/quickstart) I see the gRPC endpoint is `https://grpc.mainnet.aptoslabs.com:443`.

When I visit that endpoint, I see that it's up, and showing `unauthorized`:

<img width="1124" alt="Image" src="https://github.com/user-attachments/assets/fdc4ff6a-b7c0-44f3-ba8a-205ab34da13f" />

So, essentially the node URL `https://mainnet.aptoslabs.com` with `grpc` prepended.

For Movement, in [the docs](https://docs.movementnetwork.xyz/devs/indexing#third-party-indexers) I see 

> A running a Movement Full Node that serves Aptos gRPC. In this example https://full.mainnet.movementinfra.xyz/v1 and listens to indexer queries on port 30734.

But when I go to `https://grpc.full.mainnet.movementinfra.xyz` it doesn't show `unauthorized`, it only shows that the page isn't found.

Is there a different address for the grpc stream, other than in the Movement docs?

[...]

PR #1278, opened 2025-07-16
[feat] add metrics+health server component for indexer v2 (1/n)

larryl3u:
# Summary
- **RFCs**: https://github.com/movementlabsxyz/movement/issues/1258
- **Categories**:  `util`

* Add simple components for health + metrics server. 
  * We use a different approach that drives the processors and there is no metrics component in v2. To support the metrics, we create a component here.

# Changelog

* Add health check and metrics endpoint component. 

# Testing

Locally tested and looks good! 

* Metrics

image

* Health check

image

# Outstanding issues

### DIFF
--- a/protocol-units/execution/maptos/util/src/config/health_server.rs
+++ b/protocol-units/execution/maptos/util/src/config/health_server.rs
@@ -1,0 +1,39 @@
+use super::common::{default_health_server_hostname, default_health_server_port};
+use anyhow::Error;
+use poem::listener::TcpListener;
+use poem::{get, handler, IntoResponse, Response, Route, Server};
+use serde::{Deserialize, Serialize};
+
+// An additional health server to be used by the indexer(or any other service).
+// Do not use this with node since it exposes various endpoints to verify the health of the node.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct Config {
+	#[serde(default = "default_health_server_hostname")]
+	pub hostname: String,
+	#[serde(default = "default_health_server_port")]
+	pub port: u16,
+}
+
+impl Default for Config {
+	fn default() -> Self {
+		Self { hostname: default_health_server_hostname(), port: default_health_server_port() }
+	}
+}
+
+impl Config {
+	pub async fn run(self) -> Result<(), anyhow::Error> {
+		let url = format!("{}:{}", self.hostname, self.port);
+		run_service(url).await
+	}
+}
+
+pub async fn run_service(url: String) -> Result<(), Error> {
+	let route = Route::new().at("/health", get(health));
+	tracing::info!("Start health check access on :{url} .");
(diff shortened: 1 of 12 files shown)


protocol-units/execution/maptos/util/src/config/health_server.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+impl Config {
+	pub async fn run(self) -> Result<(), anyhow::Error> {
+		let url = format!("{}:{}", self.hostname, self.port);
+		run_service(url).await
+	}
+}
+
+pub async fn run_service(url: String) -> Result<(), Error> {
```
musitdev, line 30: Not sure if we should put the server start in the config. Why didn't you add the health server start in the indexer main ?

protocol-units/execution/maptos/util/src/config/metrics_server.rs
```diff
@@ -0,0 +1,64 @@
...
+				tracing::error!("Metrics server error: {}", e);
+			}
+		});
+
+		Ok(server_handle)
+	}
+}
+
+#[handler]
+async fn metrics_handler() -> impl IntoResponse {
```
musitdev, line 45: The same has the health server.

protocol-units/execution/maptos/util/src/config/metrics_server.rs
```diff
@@ -0,0 +1,64 @@
+use super::common::{default_metrics_server_hostname, default_metrics_server_port};
+use poem::http::StatusCode;
+use poem::{get, handler, listener::TcpListener, IntoResponse, Route, Server};
+use prometheus::{gather, Encoder, TextEncoder};
+use serde::{Deserialize, Serialize};
+use tokio::task::JoinHandle;
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct MetricsConfig {
```
musitdev, line 9: @0xIcarus has added metric to the full node. Did you see with him to see if it's compatible with its integration?

musitdev (commented):
My comment is mostly on the fact that if your metric integration is compatible with the Fullnode one.

protocol-units/execution/maptos/util/src/config/metrics_server.rs
larryl3u: yea but it introduces many metrics that indexer doesn't need. update it to a component under `tracing` which can be used by anything other than fullnode.

protocol-units/execution/maptos/util/src/config/metrics_server.rs
larryl3u: ditto

andygolay (approved):

larryl3u:
move to https://github.com/movementlabsxyz/movement/pull/1283 ty!

PR #1284, opened 2025-07-24
[feat] Enable internal indexer db for grpc stream.

larryl3u:
# Summary
- **RFCs**: $\emptyset$.
- **Categories**: `indexer`

# Changelog

* Enable the internal indexer to make grpc stream work. 
* This fixes the `write_item` unwrap failure. 

# Testing

Locally tested and looks good. 

image

DB looks good

image

# Outstanding issues

The processor tracker looks broken; i'll fix in the next PR


protocol-units/execution/maptos/opt-executor/src/indexer.rs
```diff
@@ -1,31 +1,53 @@
...
+		let (_table_info_runtime, async_indexer) = bootstrap_table_info(
 			&self.node_config,
 			self.maptos_config.chain.maptos_chain_id.clone(),
 			self.db.clone(),
 			self.mempool_client_sender.clone(),
 		)
 		.ok_or(anyhow::anyhow!("Failed to bootstrap table info runtime"))?;
+		// Internal indexer db.
+		let internal_indexer_db = InternalIndexerDBService::get_indexer_db(&self.node_config);
+		let (_internal_indexer_db_runtime, txn_event_reader) = match bootstrap_internal_indexer_db(
```
musitdev, line 32: Why you put an _ before `internal_indexer_db_runtime`, the variable is used.

protocol-units/execution/maptos/opt-executor/src/indexer.rs
```diff
@@ -1,31 +1,53 @@
...
+			&self.node_config,
+			self.db.clone(),
+			internal_indexer_db,
+		) {
+			Some((runtime, reader)) => (Some(runtime), Some(reader)),
+			None => (None, None),
```
musitdev, line 38: I think in case of `None` you can immediately `anyhow::bail!` the error like below and in case of `Some()`, return `(runtime, Some(reader))` .

protocol-units/execution/maptos/opt-executor/src/indexer.rs
```diff
@@ -1,31 +1,53 @@
...
+		};
+		let _internal_indexer_db_runtime = match _internal_indexer_db_runtime {
+			Some(runtime) => runtime,
+			None => {
+				anyhow::bail!("Failed to bootstrap internal indexer db runtime");
+			}
+		};
```
musitdev, line 45: It allows simplifying this code.

protocol-units/execution/maptos/opt-executor/src/indexer.rs
```diff
@@ -1,31 +1,53 @@
...
-		let (_table_info_runtime, _async_indexer) = bootstrap_table_info(
+		let (_table_info_runtime, async_indexer) = bootstrap_table_info(
 			&self.node_config,
 			self.maptos_config.chain.maptos_chain_id.clone(),
 			self.db.clone(),
 			self.mempool_client_sender.clone(),
 		)
 		.ok_or(anyhow::anyhow!("Failed to bootstrap table info runtime"))?;
+		// Internal indexer db.
+		let internal_indexer_db = InternalIndexerDBService::get_indexer_db(&self.node_config);
```
musitdev, line 31: What is the purpose of the IndexerDB on the full node side?

larryl3u: This is a hard dependency. 

For write_table_item, there are 3 elements: table_handle, key in bytes, and value in bytes.

table info servive provides a mapping from table_handle to type. the deserialization is depenendent on move viewer, which depdends on the internal indexer db.

https://github.com/aptos-labs/aptos-core/blob/c7033160162cf9f46ccebe4195666e126972788b/api/types/src/convert.rs#L571-L572

musitdev: Ok It's a sort of local cache to convert some Tx data to a usable value. To avoid adding more load on the current node, we should create a new full node that will handle indexer grpc query and this DB.
Does this db need to sync all db tx from the beginning when it's activated?

protocol-units/execution/maptos/opt-executor/src/indexer.rs
larryl3u: updated; also update the rest of the file with the same pattern.

protocol-units/execution/maptos/opt-executor/src/indexer.rs
larryl3u: removed.

protocol-units/execution/maptos/opt-executor/src/indexer.rs
larryl3u: removed.

protocol-units/execution/maptos/opt-executor/src/indexer.rs
larryl3u: Removed the shard db, i.e., `internal_indexer_db`. Keep the storage indexer, which is a hard dependency to deserialize the data into Move type.

larryl3u:
> Wondering if this is strictly required, as per my one question in the PR. Thanks!

Nope it's not required. Removed now.

PR #1295, opened 2025-08-13
feat: deprecate ggp and add e2e tests to check for correct gas collection using collect_fee

0xIcarus:
# Summary
- **RFCs**:  $\emptyset$.
- **Categories**: `networks`, `misc`.

Deprecating governed gas pool and replacing gas collection logic with `transaction_fee::collect_fee` module using a fw upgrade.  Added e2e tests to confirm deprecation and correct fee collection values and to query for the module on chain.
# Changelog

- Removed all usage of ggp.
- Updated the framework ref to match the latest ref in branch `0xIcarus/ggp-deprecate` on `movementlabsxyz/aptos-core` to allow feature `COLLECT_AND_DISTRIBUTE_GAS_FEES` to be voted through.
- Added e2e tests to make sure gas collection, feature enablement and ggp deprecation works as expected.

# Testing

New e2e tests added to verify if fee collection behavior matches expectations
Run using `nix develop --command bash -c "just movement-full-node native build.setup.test-e2e-verify-collect-fee --keep-project"`
# Outstanding issues

### DIFF
--- a/networks/movement/movement-client/src/bin/e2e/e2e_ggp_deprecation.rs
+++ b/networks/movement/movement-client/src/bin/e2e/e2e_ggp_deprecation.rs
@@ -266,12 +266,13 @@ async fn main() -> Result<(), anyhow::Error> {
 	// Execute test transaction
 	println!("Executing test transaction...");
 	
+	let transfer_amount: u64 = 500;
 	let test_txn = coin_client
-		.transfer(&mut sender, beneficiary.address(), 1_000, None)
+		.transfer(&mut sender, beneficiary.address(), transfer_amount, None)
 		.await
 		.context("Failed to submit test transaction")?;
 	
-	rest_client
+	let confirmed_txn = rest_client
 		.wait_for_transaction(&test_txn)
 		.await
 		.context("Failed when waiting for transfer transaction")?;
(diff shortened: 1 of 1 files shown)


justfile
```diff
@@ -12,3 +12,10 @@ build-push-container IMAGE:
     ./scripts/movement/build-push-image {{ IMAGE }}
 container-tests:
     ./scripts/tests/container-tests
+
+# E2E Tests for GGP Deprecation
+test-e2e-verify-collect-fee:
+    process-compose -f process-compose/movement-full-node/process-compose.test-e2e-verify-collect-fee.yml up --wait --follow
+
+test-e2e-framework-upgrade-collect-gas-fees:
+    process-compose -f process-compose/movement-full-node/process-compose.test-e2e-framework-upgrade-collect-gas-fees.yml up --wait --follow
```
sebtomba, line 21: You can remove this. E2E tests require a full-node setup and faucet. We are probably going to run it with
```
 just movement-full-node native build.setup.test-e2e-verify-collect-fee
 just movement-full-node native build.setup.
test-e2e-framework-upgrade-collect-gas-fees
```

protocol-units/execution/maptos/framework/releases/known-release/src/lib.rs
```diff
@@ -27,23 +29,26 @@ pub enum KnownReleaseError {
...
+				)),
+				"pre-l1-merge" => Ok(KnownRelease::PreL1Merge(
+					aptos_framework_pre_l1_merge_release::cached::full::feature_upgrade::PreL1Merge::new(),
+				)),
+				"head" => Ok(KnownRelease::Head(aptos_framework_head_release::Head::new())),
+				_ => Err(KnownReleaseError::InvalidIdentifier(
+					format!("unknown release string: {}", release).into(),
+				)
+				.into()),
+			}
```
sebtomba, line 50: The code formatting is off.

sebtomba (changes requested):

musitdev:
I've tested to update mainnet db. The `aptos-framework-pre-l1-merge-release-tool` works well. I didn't manage to pass the test. The faucet is blocking the end-to-end (E2E) test execution, and I try to verify the on-chain value with an RPC query, but I couldn't find how to obtain the correct value. The RPC calls always fails
From my test the important point I saw is that the  `aptos-framework-pre-l1-merge-release-tool` compiles all scripts, which takes time. For the mainnet migration, we must use precompiled scripts. It will significantly decrease the migration time.
The other point, precompiled move script guarantee that we use the right compiler, which is not guaranteed on the migration instance.

PR #1299, opened 2025-08-21
fix(docs): update broken Aptos Faucet API link in README

VolodymyrBg:
https://aptos.dev/apis/#faucet-api-only-testnetdevnet - broken
https://aptos.dev/build/apis/faucet-api - correct


musitdev (approved):

VolodymyrBg:
> @VolodymyrBg shouldn't the faucet APIs link to Movement's faucets rather than Aptos? I might be missing some context here.

i may be missing some context, too. But the previous link was leading to aptos faucets

PR #1303, opened 2025-08-26
[DO NOT SUBMIT] enable the logger

larryl3u:
# Summary

# Changelog

# Testing

# Outstanding issues


musitdev:
I tried to see why I was discarded, but I didn't have the logs of the VM execution. Are there any other parameters that need to be changed to enable these logs?

PR #1312, opened 2025-10-28
Primata/add deploy Hyperliquid and Secondary chains

Primata:
# Summary

Adds:

Contracts:
1. MOVETokenHyperliquid
2. MOVETokenL2

Test units:
1. Hyperliquid Mainnet

Scripts:
1. MOVETokenHyperliquid Mainnet
2. MOVETokenHyperliquid Testnet

# Changelog

# Testing

MOVETokenL2 is untested but if we assume MOVETokenHyperliquid is safe, L2 would also be safe. L2 implementation is a version with less features than Hyperliquid.

# Outstanding issues
Add script for MOVETokenL2 deployment
Branch typo Hypernative X Hyperliquid OK


protocol-units/settlement/mcr/contracts/script/DeployMOVETokenHyperliquid.s.sol
```diff
@@ -0,0 +1,329 @@
...
+}
+
+/**
+ * @title DeployMOVETokenHyperliquid
+ * @notice Deployment script for MOVE token on Hyperliquid with full LayerZero configuration
+ * @dev Deploys implementation, proxy via CREATE3, configures LayerZero DVNs, and sets peer
+ */
+contract DeployMOVETokenHyperliquid is Script {
+    // Deployment addresses
+    address constant DEPLOYER_ADDRESS = 0xB2105464215716e1445367BEA5668F581eF7d063;
```
andygolay, line 40: What is this deployer address? Is it intended to be the owner of the contract? Is it a multisig and if not will it transfer ownership to a multisig?

andygolay (changes requested):
@Primata could you add more context to the PR description for example explain "MOVETokenL2 is untested but if we assume MOVETokenHyperliquid is safe, L2 would also be safe."

Eg why MOVETokenHyperliquid should be assumed to be safe and how it can be tested, and why it follows that if MOVETokenHyperliquid is safe, it follows that MOVETokenL2 is safe.

protocol-units/settlement/mcr/contracts/script/DeployMOVETokenHyperliquid.s.sol
Primata: it's the original token address, we must use the same address, this way we can have the same address on Hyperliquid. Transfer to Multisig will be done after confirmation that setup is okay.

Primata:
> @Primata could you add more context to the PR description for example explain "MOVETokenL2 is untested but if we assume MOVETokenHyperliquid is safe, L2 would also be safe."
> 
> Eg why MOVETokenHyperliquid should be assumed to be safe and how it can be tested, and why it follows that if MOVETokenHyperliquid is safe, it follows that MOVETokenL2 is safe.

updated

protocol-units/settlement/mcr/contracts/test/token/MOVETokenV2.t.sol
```diff
@@ -562,10 +631,131 @@ contract MOVETokenV2Test is Test {
...
+        uint256 amountOutV3 = abi.decode(dataV3, (uint256));
+
+        // Adds eth and move balance to uniswapv4 pool
+        vm.deal(0x1B42bb0771690a3A82beDb1BC74933788145CbfD, 100 ether);
+        deal(address(move2), 0x1B42bb0771690a3A82beDb1BC74933788145CbfD, 100 ether);
+        uint24 v4fee = 10000;
+        // quote from uniswap v4
+        (bool successV4, bytes memory dataV4) = uniswapv4Quoter.call(
+            abi.encodeWithSignature(
+                "quoteExactInputSingle((address,address,uint256,uint24,uint160))",
```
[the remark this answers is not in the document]
andygolay: As far as I can tell, this is not a legitimate finding. As the comment admits, uint256 is correct. So it seems like a false positive.

0xmovses:
What cmds did you run for your unit testing?

